PR #2, opened 2019-02-21
Remove spaces in enum names

enmand:
Enum values can be more than one word, so replace whitespace so that enum names are valid

### DIFF
--- a/src/swagger-2.ts
+++ b/src/swagger-2.ts
@@ -31,9 +31,7 @@ function capitalize(str: string) {
 }
 
 function camelCase(name: string) {
-  return name.replace(/(-|_|\.|\s)+[a-z]/g, letter =>
-    letter.toUpperCase().replace(/[^0-9a-z]/gi, '')
-  );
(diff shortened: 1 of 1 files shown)


src/swagger-2.ts
```diff
@@ -27,7 +27,7 @@ const TYPES: { [index: string]: string } = {
 };
 
 function capitalize(str: string) {
-  return `${str[0].toUpperCase()}${str.slice(1)}`;
+  return `${str[0].toUpperCase()}${str.slice(1).replace(' ', '')}`;
```
dangodev: If it’s all the same, maybe this should go at the end of `camelCase` instead? I think camelCase is expecting a non-space anyway.

Something like `.replace(/\s+/g, '')` (I think in your example, only the first space would be replaced, and then it would skip the rest—maybe).

dangodev: Actually, wait—`camelCase` should be replacing spaces with capital letters. 🤔 Why is that failing for `'American Express'`?

PR #4, opened 2019-02-21
Create interfaces for additionalProperties objects

enmand:
This PR adds supports for object definitions with `additionalProperties: boolean | Swagger2Definition`

### DIFF
--- a/src/swagger-2.ts
+++ b/src/swagger-2.ts
@@ -170,7 +170,7 @@ function parse(spec: Swagger2, namespace: string) {
 
     if (additionalProperties) {
       if (<boolean>additionalProperties === true) {
-        output.push(`[name: string]: ${ID}`);
+        output.push(`[name: string]: any`);
       }
 
       if ((<Swagger2Definition>additionalProperties).type) {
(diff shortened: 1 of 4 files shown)


src/swagger-2.ts
```diff
@@ -181,7 +198,7 @@ function parse(spec: Swagger2, namespace: string) {
 
   output.push('}'); // Close namespace
 
-  return prettier.format(output.join('\n'), { parser: 'typescript' });
+  return prettier.format(output.join('\n'), { parser: 'typescript', singleQuote: true });
```
dangodev: Just a note: I personally like singleQuote, but since we’re open-sourcing it I didn’t know the best thing to decide on. Maybe we could add options one day to allow customization of this, but this is fine for now (we can always change it if someone asks!)

dangodev:
@enmand are you able to add a test to `tests/swagger-2.test.ts` to test for this case? (also make sure all the existing tests pass?)

tests/swagger-2.test.ts
```diff
@@ -249,6 +249,44 @@ describe('Swagger 2 spec', () => {
...
+    };
+
+    const ts = format(`
+    export interface Credentials {
+      [name: string]: string;
+    }`);
+
+    expect(swaggerToTS(swagger)).toBe(ts);
+  });
+
```
dangodev: 💯

PR #11, opened 2019-04-11
 Add an option to export the namespace

tpdewolf:
I'm autogenerating types on build and had to add the export keyword every time. I thought it would be nice to add it as an option. I even would suggest to enable it by default but I didnt want to change the current default behaviour


dangodev:
Nice! I really like this.

I’d like to add a test for this option. I’m happy to merge this and add it, or you can take a crack at adding a test if you’d prefer. Which would you like?

dangodev:
@tpdewolf I’m going to merge this, and add a test as a followup. I’ll release a new version to npm this week. Thanks again!

dangodev:
Thanks for your patience. `v1.1.3` is now on npm with your export option. Thanks!

PR #25, opened 2019-07-10
changing type check to use typeof operator instead of object property

thomascclay:
I was having an issue where the output was not being generated.  After debugging the script I found out this type check was the problem.  There's no `type` property on the object, so it returns `undefined` which fails the check and then the object never makes it onto the generation queue.

The proper method of type checking is to use `typeof`

### DIFF
--- a/src/swagger-2.ts
+++ b/src/swagger-2.ts
@@ -191,7 +191,7 @@ function parse(spec: Swagger2, options: Swagger2Options = {}): string {
   Object.entries(definitions).forEach(
     (entry): void => {
       // Ignore top-level array definitions
-      if (entry[1].type === 'object') {
+      if (typeof entry[1] === 'object') {
         queue.push(entry);
       }
     }


dangodev:
@thomascclay thanks so much for submitting a PR! Though I think for the Swagger spec, there _should_ be a `.type` key on the object. It’ll _always_ be `typeof 'object'` in JavaScript, but it may be `type: object` or `type: array` or `type: string`, …

As you can see from the tests in Travis, changing this breaks the output.

Can you provide the Swagger 2 spec that’s not working for you? I want to make sure I handle your usecase without breaking the output for existing code.

thomascclay:
Ah I did miss that part in my definitions section; thought I was good when https://editor.swagger.io/ did not report any warnings on it.  Adding the `type: object` to it worked.

I am having another problem with it missing some enums, but I'll open an issue for this.

dangodev:
Yeah please do! For future reference, providing a) the spec you’re using, and b) the expected TS output would help lots. Enums in particular were weird; we went with a route that worked better for us, but maybe there are some options we can provide if you have different enum needs.

PR #48, opened 2019-08-16
feat: add capability to use url input

trenoncourt:


dangodev:
Wow! 😮 Great Pull Request! I’m going to test this out later tonight; hang tight till then!

Having a test for this would be great, but I know adding tests for the CLI is difficult, currently (right now the `npm run generate` command is the only test). Would love your thoughts on this.

trenoncourt:
Hi @dangodev,

Thanks, i've added a generate:url script in package.json with your repo raw yaml input files to test.

dangodev:
> Thanks, i've added a generate:url script in package.json with your repo raw yaml input files to test.

Right, yeah, but I was saying that Travis actually runs that command to make sure that bit is working before any PR is merged. But since yours talks to a live URL we may need to mock that somehow to make that a merge blocker (unless that URL has crazy uptime or something)

dangodev:
Tested and this works great! 1 thing I’d like to change, though, and 1 thing I’d like to add:

**Change**
Rather than `axios`, can you use Node’s [`https.get()`](https://nodejs.org/api/https.html#https_https_get_url_options_callback) instead for the fetching? I’m a big fan of axios, but for grabbing one url (and because we have users using `npx`), it’s a bit of weight we don’t need.

**Add**
Would you mind adding to the README docs on the feature you added?

Once those 2 things are in I’m happy to merge & cut a new version on npm. But if you can’t tackle either let me know! In that case I can do either.

dangodev:
I’m closing this for now because I’m getting pings on carrying this over. But what I’m going to do is: close it for now, but I’ll re-open and merge when I have time to make the changes requested.

PR #67, opened 2019-09-30
feat(doc): prints out jsdoc

### LINKED ISSUE #57: Descriptions as JSDoc comments
It would be just great to have `description` fields output not as just regular comments (https://github.com/manifoldco/swagger-to-ts/blob/master/src/swagger-2.ts#L172-L175), but rather as JSDoc ones.

This way, IDEs can show that description in the popup hints, which is super handy.

lbenie:
Instead of having regular comments. Let's use jsdoc comments.

closes #57

### DIFF
--- a/example/nowrapper.ts
+++ b/example/nowrapper.ts
@@ -1,7 +1,11 @@
 export interface ValueProp {
-  // Heading of a value proposition.
+  /**
+   * Heading of a value proposition.
+   */
   header: string;
-  // Body of a value proposition.
+  /**
+   * Body of a value proposition.
+   */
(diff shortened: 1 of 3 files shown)


dangodev:
@lbenie Awesome! I’m cool with this change.

Do me a favor and run `npm run generate` to re-generate the example, then commit that file. That’s why thy Travis tests are failing. But once that’s good, happy to merge & ship!

lbenie:
@dangodev done !

dangodev:
Published `1.4.0` 🎉

PR #70, opened 2019-10-09
Add optional warning at top of generated files

tombarton:
This optional feature injects a warning at the top of the generated file. This is pretty standard behaviour followed by Apollo, Prisma, etc. when dealing with auto-generated files to discourage people from editing them.


tombarton:
@dangodev might you need to merge it I'm afraid as I don't have write access :)

dangodev:
Hey @tombarton! I liked your warning so much I turned it on by default 🙂. I changed the flag to `--no-warning` to disable; otherwise it’ll add by default (so it’s backwards compatible in that `--injectWarning` will still generate a warning because it’ll be ignored).

Let me know what you think!

Issue #77, opened 2019-10-17
support for x-nullable (and other x- props)


atlefren:
This library works great, but the swagger endpoints i'm using it against use the x-nullable property to indicate optional parameters (why it's doing that is another issue, don't ask).

I think it's easier to change the code generation than it is changing a lot of swagger defs, so I would really appreciate if this library had an option to support various x-flags.

What I'm thinking is an option to provide a mapper to the generator, something along the lines of 

```

interface SwaggerProps {
 [string]: any //not sure how "unknown" props should be typed
}

interface Parameter {
  name: string;
  type: Type;
  required: boolean;
  //etc...
}


const mapper = (swaggerProps: SwaggerProps, parameter: Parameter): Parameter => {
    return {...parameter, required: !swaggerProps['x-nullable']};
}

swaggerToTS(cleanedSwaggerDefinition, {wrapper: false, mapper: mapper})
```

Does this sound like something you would consider including? I could try to implement it and submit a PR, but I would like a discussion up front in order to know it's something that will be considered for inclusion

dangodev:
> the swagger endpoints i'm using it against use the x-nullable property

Actually the Swagger endpoints we use have that too! It wasn’t included just because we didn’t need that in TypeScript. I’m not sure if it’s true spec, but it seems common enough to include

> Does this sound like something you would consider including? I could try to implement it and submit a PR, but I would like a discussion up front in order to know it's something that will be considered for inclusion

Let’s include it! The good part of this repo is that there are tests to ensure everything is working, so we can add that without disrupting anything else. If you could open a PR that’d be great, and I can help write a test if you’re unfamiliar.

If you don’t have the time for a PR (I know it’s the new year now) I can add this to the roadmap of things to improve.

atlefren:
Great!

I will try to find some time to start working on a PR for this in the next couple of weeks. I'll keep you updated on the progress (or lack thereof)

dangodev:
version `1.6.0` just published to npm 🎉

PR #118, opened 2020-01-15
Property mapper function

atlefren:
As discussed in #77 .

Had to change the logic in properties handling a bit, and introduced a new interface Properties. In addition, the Swagger2Definition interface is now exported from the top-level, as it is used in the mapper. 

New tests added for the new functionality, all existing tests still passes. Readme is also updated. 

Comments and suggestions welcome.


atlefren:
Great, thanks!

PR #137, opened 2020-02-06
fix(utils/sanitize): Take into account the at sign

sorin-davidoi:
This is not a scalable solution, but it's quite common for some fields to be prefixed with `@`.


dangodev:
`v1.7.1` is now up on npm 🎉! Thanks so much!

PR #180, opened 2020-04-05
Add OpenAPI3 primitives

dangodev:
This is a PR to support primitive types (`string` | `number` | `boolean`) for [OpenAPI 3.0](https://swagger.io/specification/).

## Problem

If you look in the `tests/v3/index.test.ts` file, you’ll see a test that takes a given schema, and expects a certain output. For example, given this schema (JSON):

```json
{
  "openapi": "3.0",
  "components": {
    "schemas": {
      "object": { "properties": { "string": { "type": "string" } }, "type": "object" },
      "string": { "type": "string" },
      "string_ref": { "$ref": "#/components/schemas/string" }
    }
  }
}
```

we expect this output:

```ts
export interface components {
  schemas: {
    object: { string: string };
    string: string;
    string_ref: components['schemas']['string'];
  };
}
```

But when we run our tests:

```
npm test v3
```

❌ We get errors!

We’re not getting the right output because we’re not transforming anything.

## Solution

A good solution will make use of `JSON.parse()`’s [reviver function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#Using_the_reviver_parameter). You’ll have to read the article linked for a full explanation, but the short version is: **it’s a really great way to take a deeply-nested object and transform every single part of that object without having to write any loops.**

But while it can do a lot with very little code (and performantly, too, I might add), it can be tough to follow. I suggest reading a bit more about that function, but once you start to get it, it’s the key to everything.

## Tips

- You can look in `src/v2.ts` for some hints, specifically the “1st pass” and “3rd pass” sections (don’t worry about the `propertyMapper` block for now—that’s not a part of this PR).
- You may also notice that v2 adds question marks for optional keys in TypeScript. Don’t worry about that for this PR, either.
- At the end you may notice that you end up with `{ string: "string" }` rather than `{ string: string }` (no quotes at the end). Once you’ve made it this far, consider using `escape()` and/or `unescape()` from `utils.ts` (also look in `src/v2.ts` for hints)

## Setup

With [Node](https://nodejs.org) v12 or v13 installed, clone this repo and run `npm install` from a terminal, from this folder.

Then to see the failing tests, run `npm test v3` (note: you can just run `npm test`, but that runs all the tests; you can add the `v3` at the end to only run tests that have `v3` in the name, saving a little time).

Your job is to get the `string`, `number`, and `boolean` tests working. Good luck!


src/types/OpenAPI3.ts
```diff
@@ -1,5 +1,23 @@
...
+  description?: string;
+  enum?: string[];
+  items?: OpenAPI3SchemaObject | OpenAPI3Reference;
+  nullable?: boolean;
+  oneOf?: (OpenAPI3SchemaObject | OpenAPI3Reference)[];
+  properties?: { [key: string]: OpenAPI3SchemaObject | OpenAPI3Reference };
+  required?: string[];
+  title?: string;
+  type: OpenAPI3Type;
+}
```
dangodev, line 23: For the most part, this file can be ignored.

src/v3.ts
```diff
@@ -0,0 +1,39 @@
...
+export const PRETTIER_OPTIONS: prettier.Options = { parser: 'typescript', singleQuote: true };
+
+export default function generateTypesV3(schema: OpenAPI3, propertyMapper?: PropertyMapper): string {
+  if (!schema.components || !schema.components.schemas) {
+    throw new Error(`🧐 Components Object missing from schema https://swagger.io/specification`);
+  }
+
+  // 1st pass: expand $refs first to reduce lookups & prevent circular refs
+  const expandedRefs = JSON.parse(JSON.stringify(schema.components.schemas), (_, node) => {
+    // TODO: if a node has a $ref, then convert "#/components/schemas/<$ref>" -> components['schemas']['<$ref>']
```
dangodev, line 19: First TODO: test if a node (a “part” of the object) has a `.$ref`. If it does, then change the string `#/components/schemas/some_thing` to `components['schemas']['some_thing']`.

src/v3.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    return node;
+  });
+
+  // 2nd pass: support primitives
+  const primitives = JSON.parse(JSON.stringify(expandedRefs), (_, node) => {
+    // TODO: convert `{ some_key: { type: "string" } }` -> `{ some_key: string }`
```
dangodev, line 26: 2nd todo: convert the following objects to the following strings:

| Original                                        | Transformed |
|:--------------------------------|:--------------|
| `{ some_key: { type: 'string' } }` | `{ some_key: string }` |
| `{ some_key: { type: 'number' } }` |  `{ some_key: number }` |
| `{ some_key: { type: 'integer' } }` |  `{ some_key: number }` (same as `number`) |
| `{ some_key: { type: 'boolean' } }` |  `{ some_key: boolean }` |

tests/v3/index.test.ts
```diff
@@ -0,0 +1,125 @@
...
+    };
+    expect(v3(schema)).toBe(
+      format(`
+      export interface components {
+        schemas: {
+          object: { string: string };
+          string: string;
+          string_ref: components['schemas']['string'];
+        }
+      }`)
```
dangodev, line 31: We want the end result to look like this 👆

tests/v3/index.test.ts
```diff
@@ -0,0 +1,125 @@
...
+import v3, { WARNING_MESSAGE, PRETTIER_OPTIONS } from '../../src/v3';
+
+// test helper: don’t throw the test due to whitespace differences
+function format(types: string): string {
+  return prettier.format([WARNING_MESSAGE, types.trim()].join('\n'), PRETTIER_OPTIONS);
+}
+
+describe('types', () => {
+  it('string', () => {
+    const schema: OpenAPI3 = {
```
dangodev, line 13: Given this schema… 👇

tests/v3/index.test.ts
```diff
@@ -0,0 +1,125 @@
...
+      format(`
+      export interface components {
+        schemas: {
+          object: { integer: number; number: number }
+          number: number;
+          number_ref: components['schemas']['number'];
+        }
+      }`)
+    );
+  });
```
dangodev, line 59: 👆 Same for the `number` and `boolean` tests too 👇  (note that every test kinda has multiple parts to each—it’s OK to focus on one line at a time for each test; you won’t get all in one pass!)

Issue #112, opened 2020-01-02
Support Swagger v3


dangodev:
Let’s support Swagger v3! 

Should support the data types mentioned here: https://github.com/manifoldco/swagger-to-ts/issues/92

milanzivic:
Bump.

Is this enhancement in development? Should we expect it in foreseeable future?

dangodev:
Hello! I haven’t started working on it yet, but plan to within the month. If anyone has realistic v3 Swagger definitions they’d like me to test, please post them here!

I know that Swaggers docs contain examples, but I’ve usually found those to be too simple and not good for testing support.

dangodev:
So this library is testing out support of v3 (see [notes](https://github.com/manifoldco/swagger-to-ts/releases/tag/v1.7.0)), but it’s in beta. Please try this against your existing v3 schema.

If you encounter bugs, please post your full schema here (or the part of your schema that’s giving you trouble if you can’t post the full thing)!

Rikilele:
Hi!
Per company reasons (heavy eye-rolling), I cannot post any of the schemas here, but I am getting a
`Cannot convert undefined or null to object` error for practically all my valid OpenAPI v3.0.0 specs. Can someone kindly explain to me typical causes for these errors? Thanks in advance!

tlenex:
Hi, same issue, here as @Rikilele has. 
@dangodev I've invesgtigated it a little and I think the issue is that theres is no more `definitions` field in v3 spec, but it has `components` instead https://swagger.io/specification/#fixed-fields.

gdspicer58:
Hello,

Same issue as @Rikilele , company-reasoned schema censoring and everything.


However, I can provide an example of something that isn't allowed in V2 but is now supported in V3.

```
{
	"components": {
		"examples": {},
		"headers": {},
		"parameters": {},
		"requestBodies": {},
		"responses": {},
		"schemas": {
			"RoutingTable": {
				"properties": {},
				"type": "object",
				"additionalProperties": {
					"properties": {},
					"additionalProperties": {
						"properties": {
							"routing": {
								"type": "boolean"
							}
						},
						"required": ["routing"],
						"type": "object"
					},
					"type": "object"
				}
			},
			"ContractRequest": {
				"properties": {
					"stateToStateRoutingTable": {
						"$ref": "#/components/schemas/RoutingTable",
						"nullable": true
					}
				},
				"type": "object",
				"additionalProperties": true
			},
[...]

PR #200, opened 2020-04-22
Add basic support for OpenAPI 3.x

dangodev:
Adds tests for OpenAPI v3! 🎉

Getting closer to #112


bin/cli.js
```diff
@@ -45,6 +45,16 @@ const timeStart = process.hrtime();
...
     const outputFile = path.resolve(process.cwd(), cli.flags.output);
+
+    // recursively create parent directories if they don’t exist
+    const parentDirs = cli.flags.output.split(path.sep);
+    for (var i = 1; i < parentDirs.length; i++) {
+      const dir = path.resolve(process.cwd(), ...parentDirs.slice(0, i));
+      if (!fs.existsSync(dir)) {
+        fs.mkdirSync(dir);
+      }
+    }
```
dangodev, line 56: This replaces the previous need for `fs-extra`. Now if we could find a quick replacement for Prettier…

PR #218, opened 2020-04-22
Support null values

antonk52:
Since some properties can be nullable, the default value can be `null`. We need to check for it explicitly. Otherwise the command will fail with a runtime exception.

Below is the schema to reproduce the issue

```yaml
openapi: 3.0.1

info:
  title: sample schema
  version: LATEST

components:
  schemas:
    abstractName:
      type: integer
      format: int64
      nullable: true
      default: null
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -148,3 +148,10 @@ export function tsIntersectionOf(types: string[]): string {
 export function tsUnionOf(types: string[]): string {
   return types.join(" | ");
 }
+
+/**
+ * Conditionally convert value into nullable
+ */
+export function nullableWrap(nullable: boolean, value: string): string {
+  return nullable ? tsUnionOf([`(${value})`, "null"]) : value;
+}
--- a/src/v3.ts
+++ b/src/v3.ts
@@ -9,6 +9,7 @@ import {
   isOneOfNode,
   isSchemaObj,
   makeOptional,
+  nullableWrap,
   tsArrayOf,
   tsIntersectionOf,
   tsUnionOf,
@@ -46,7 +47,8 @@ export default function generateTypesV3(
   const objectsAndArrays = JSON.parse(
     JSON.stringify(propertyMapped),
     (_, node) => {
-      if (node === null) return "null";
+      // avoiding runtime exceptions for `null` values
+      if (node === null) return null;
 
       // $ref
       if (node["$ref"]) {
@@ -57,26 +59,34 @@ export default function generateTypesV3(
         ); // important: use single-quotes here for JSON (you can always change w/ Prettier at the end)
       }
 
+      const { nullable = false } = node;
+
       // oneOf
       if (isOneOfNode(node)) {
-        return escape(tsUnionOf(node.oneOf));
+        return nullableWrap(nullable, escape(tsUnionOf(node.oneOf)));
       }
 
       // anyOf
       if (isAnyOfNode(node)) {
-        return escape(
-          tsIntersectionOf(
-            (node.anyOf as string[]).map((item) => escape(`Partial<${item}>`))
+        return nullableWrap(
+          nullable,
+          escape(
+            tsIntersectionOf(
+              (node.anyOf as string[]).map((item) => escape(`Partial<${item}>`))
+            )
           )
         );
       }
 
       // primitive
       if (node.type && PRIMITIVES[node.type]) {
-        return escape(
-          node.enum
-            ? tsUnionOf((node.enum as string[]).map((item) => `'${item}'`))
-            : PRIMITIVES[node.type]
+        return nullableWrap(
+          nullable,
+          escape(
+            node.enum
+              ? tsUnionOf((node.enum as string[]).map((item) => `'${item}'`))
+              : PRIMITIVES[node.type]
+          )
         );
       }
 
@@ -88,7 +98,7 @@ export default function generateTypesV3(
           (!node.allOf || !node.allOf.length) &&
           !node.additionalProperties
         ) {
-          return escape(`{[key: string]: any}`);
+          return nullableWrap(nullable, escape(`{[key: string]: any}`));
         }
 
         // unescape properties if some have been transformed already for nested objects
@@ -119,25 +129,28 @@ export default function generateTypesV3(
           properties[escape("[key: string]")] = escape(addlType || "any");
         }
 
-        return tsIntersectionOf([
-          // append allOf first
-          ...(node.allOf
-            ? node.allOf.map((item: any) =>
-                isSchemaObj(item)
-                  ? JSON.stringify(makeOptional(item, node.required))
-                  : item
-              )
-            : []),
-          // then properties + additionalProperties
-          ...(Object.keys(properties).length
-            ? [JSON.stringify(properties)]
-            : []),
-        ]);
+        return nullableWrap(
+          nullable,
+          tsIntersectionOf([
+            // append allOf first
+            ...(node.allOf
+              ? node.allOf.map((item: any) =>
+                  isSchemaObj(item)
+                    ? JSON.stringify(makeOptional(item, node.required))
+                    : item
+                )
+              : []),
+            // then properties + additionalProperties
+            ...(Object.keys(properties).length
+              ? [JSON.stringify(properties)]
+              : []),
+          ])
+        );
       }
 
       // arrays
       if (isArrayNode(node) && typeof node.items === "string") {
-        return escape(tsArrayOf(node.items));
+        return nullableWrap(nullable, escape(tsArrayOf(node.items)));
       }
 
       return node; // return by default
--- a/src/v3.ts
+++ b/src/v3.ts
@@ -46,6 +46,8 @@ export default function generateTypesV3(
   const objectsAndArrays = JSON.parse(
     JSON.stringify(propertyMapped),
     (_, node) => {
+      if (node === null) return "null";
+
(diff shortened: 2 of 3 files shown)


dangodev:
@antonk52 are you able to provide a test for this? I don’t think this PR will have its intended effect.

Also, what should the generated types look like for this? Currently I think it would be:

```ts
components: {
  schemas: {
    abstractName?: number;
  }
}
```

But you’re saying it should be `abstractName?: number | null`?

Or something else?
```

**Comment by antonk52** [comment]
```
> I don’t think this PR will have its intended effect.
```

not sure I understood this, can you paraphrase it?

The type you provided is incorrect. I did not get why the `abstractName` is optional in your type example. It is missing the `required` property, which is by default has value `true`. the property is nullable though. Therefore the correct type would be

```ts
components: {
  schemas: {
    abstractName: ?number; // number | null
  }
}
```
please note that question mark moved to the right from `:`.

I am happy to cover this case with a unit test.

dangodev:
> It is missing the required property, which is by default has value `true`

Ah you know what? In the current `master` we make things optional by default in OpenAPI 3.0, a carryover from Swagger 2.0. But you’re right the schema dictates that all properties are required unless specified otherwise. I’ll make that change in another PR! 🙏 

---

> please note that question mark moved to the right from `:`.

I don’t think that `?number` is valid TypeScript; I think it would need to be `number | 'null'`.

---

> not sure I understood this, can you paraphrase it?

Sorry; I didn’t go into detail before because it was a rather long explanation.

This library uses JSON.stringify()’s [replacer parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#The_replacer_parameter), which transforms from the bottom-up.

Your code has `if (node === null)`, which will _only_ transform `default: null` into `default: "null"` (the `node` here is the value).

Then, because it’s bottom-up transformation, it’ll move to the parent node:

```ts
abstractName: {
  type: "integer",
  default: "null"
}
```

Its parent node is of type `object`, not `null`, and the the current code finds `type: "integer"` which will be currently transformed into `number`. So the previous transformation will be undone.

---

Anyway I think **we should ignore `default` altogether** and just focus on the `nullable: true` bit. That will actually be pretty far-reaching, because it could affect all types. Basically we need to append a ` | 'null'` at the end of every possible type this library could generate, which means we need to write a lot of tests for this.

Are you up to the challenge? 😄 I’m happy to let you take on as much as you’d like. I’d really appreciate the help and would love for you to run with this, but also want to be sensitive of your time and contributions, too!

Also, **thank you** for your feedback and comments! You are right this is important, and something we should definitely handle before we release 3.0 support.

antonk52:
You're right regarding the `?number`, I've spent too much time with flowtype recently 😅

Got your point about the `default` property. I believe it is ignored right now any way, but since there was no explicit check for the `null` value the bin was failing with a runtime exception.

I've pushed another commit with a draft decision, I would like to hear your thoughts on it before writing any tests.

src/v3.ts
```diff
@@ -46,6 +47,9 @@ export default function generateTypesV3(
   const objectsAndArrays = JSON.parse(
     JSON.stringify(propertyMapped),
     (_, node) => {
+      // avoiding runtime exceptions for `null` values
+      if (node === null) return null;
```
dangodev, line 51: If we want to remove the `null` values, I think we should return `undefined` instead–that’ll remove the node. Returning the node as-is will just keep as-is, which will likely not solve the problem.

src/utils.ts
```diff
@@ -148,3 +148,10 @@ export function tsIntersectionOf(types: string[]): string {
 export function tsUnionOf(types: string[]): string {
   return types.join(" | ");
 }
+
+/**
+ * Conditionally convert value into nullable
+ */
+export function nullableWrap(nullable: boolean, value: string): string {
+  return nullable ? tsUnionOf([`(${value})`, "null"]) : value;
+}
```
dangodev, line 157: Awesome. Yeah this was exactly what I had in mind. I think this makes sense.

src/v3.ts
```diff
@@ -55,26 +59,34 @@ export default function generateTypesV3(
         ); // important: use single-quotes here for JSON (you can always change w/ Prettier at the end)
       }
 
+      const { nullable = false } = node;
```
dangodev, line 62: 👍 good to set the default.

dangodev:
I think this is the perfect direction! I agree with this. Now it’s just writing tests!

Like we also discussed, it seems like we should probably remove all the `?` from the type generation altogether, since if `nullable: true` that means that it will return ` | 'null'` as a value, rather than optional. Is my thinking correct?

Or in OpenAPI 3.0, are properties **only optional if `required: []` is present, otherwise they’re all required?** That part I’m confused on. Also, to me that behavior _feels_ wrong I’m also not super familiar with how 3.0 handles required properties (we still use 2.0 at my company).

Either way I’m not saying we have to tackle that in this PR, too, but I am wondering how that’ll affect the tests and whether or not to think about it.

antonk52:
If I got your idea right I'll do the suggest change

```diff
- if (node === null) return null;
+ if (node === null) return undefined;
```

Your assumption about `nullable` values is correct. As [the spec](https://swagger.io/specification/#schemaNullable) goes 
```
A true value adds "null" to the allowed type specified by the type keyword, only if type is explicitly defined within the same Schema Object. 
```

Therefore the key is still present and we need to change `?:` to `:` when defining the property type.

To determine the optionality of a property in an object we need to check the `required` field on the Schema Object. The [OAPv3 spec](https://swagger.io/specification/#properties) notes that it reuses `required` property exactly as it is specified in the [json schema](https://json-schema.org/understanding-json-schema/reference/object.html#required-properties). From which we can come to the conclusion that if not set explicitly in `required` field, all properties should be treated as optional. This does go out of scope of this PR though 😅

dangodev:
> To determine the optionality of a property in an object we need to check the required field on the Schema Object. The OAPv3 spec notes that it reuses required property exactly as it is specified in the json schema.

OK if that’s the case then I _think_ the `?` is working as intended? That’s how we treat it now—add `?` unless the field name is found within `required: []`.

dangodev:
Also sorry—my example above was confusing. We _do_ add `?` but only for nested properties, not top-level `schemas`. But other than top-level properties I think it’s working now as the spec intended.

antonk52:
I've updated `manifold.ts` to have nullable types actually nullable. I have an issue with prettier and escaped string quotes. I think it is related to [un]escape utils but not too sure, any idea on why this is happening?

dangodev:
> I have an issue with prettier and escaped string quotes. I think it is related to [un]escape utils but not too sure, any idea on why this is happening?

Ah yeah that’s the most confusing implementation detail of this library—I’m actually wondering if there’s a way to avoid that now, but for this PR, how they work is:

Say you want to transform something to a `string` TS type. That will be turned into `'string'` with JSON.stringify(). So the way to avoid that is by using `escape('string')` which is just a way of saying “this should be TypeScript, not a string.”

If that’s the issue you’re having then that may be a fix; if not I’m happy to take a closer look later tonight.

src/utils.ts
```diff
@@ -148,3 +148,10 @@ export function tsIntersectionOf(types: string[]): string {
 export function tsUnionOf(types: string[]): string {
   return types.join(" | ");
 }
+
+/**
+ * Conditionally convert value into nullable
+ */
+export function nullableWrap(nullable: boolean, value: string): string {
+  return nullable ? tsUnionOf([`(${value})`, "null"]) : value;
```
dangodev, line 156: So you know what? I realized the difficulty you’re experiencing, and it’s actually really challenging to pull off.

To start, the only thing your code missing is this:

```
-  return nullable ? tsUnionOf([`(${value})`, "null"]) : value;
+  return nullable ? tsUnionOf([`(${value})`, escape("null")]) : value;
```

But the problem you’re experiencing with the syntax error is a flaw with how this library transforms things through JSON. Normally, the library works because it transforms into a JSON-friendly syntax. But with the `(${value))` line, the JSON syntax is lost in a string interpolation.

So I’ve been thinking of a way to make this simpler, because the `escape()` / `unescape()` were “cheats” to fix the small things that broke the JSON syntax. But with this nullable edition, I think I’ll have to rethink how this library transforms things (which I’ll have to do with comments, as well).

dangodev: But what you’ve done is perfect, and I apologize that there’s a deeper issue preventing this from being merged.

Let me spend a little time thinking about this.

antonk52: Thank you for your comment and no worries. I've attempted to apply the change that you suggested but the test still fails. The reason being is the type generated like so

```ts
interface schemas {
  links?: ({\"data\":components['schemas']['file_link'][],\"has_more\":boolean,\"object\":'list',\"url\":string}) | null
//         ^
}
```

You can see that it is trying to escape the quotes for the property names which breaks prettier.

dangodev: Right, that’s from the problem I outlined. All those stringified property names get transformed back TS if they’re in JSON format with `JSON.parse()`, but they cant when wrapped with parentheses.

I think I have a solution which involves rewriting part of the library a bit, unfortunately. But I had a feeling that this was necessary because I was struggling with the same thing with JSDoc comments and `description`

dangodev:
As we discussed in the other branch, this PR was on the right track but couldn’t land due to a flaw in the library itself. It’s been resolved in #223

PR #219, opened 2020-04-22
Update .all-contributorsrc

lbenie:
Thanks for adding me in there!

I fixed my personal website URL in my GitHub profile


dangodev:
> Thanks for adding me in there!

It’s something I had been meaning to do for a while. I think it’s an awesome way to recognize contributors like yourself, because you’re what makes open source possible! 🤗 

I think the table below needs to be updated, too. Would you mind updating the link down there? The `.all-contributorsrc` file is what is used to generate the table, but the README itself is the link people click on.

PR #220, opened 2020-04-23
add default default prettier config

antonk52:
This is useful to avoid prettier to pick up global prettier config in case it is different from what is used in the project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
       ]
     ]
   },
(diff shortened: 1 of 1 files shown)


dangodev:
Works for me!

dangodev:
@all-contributors please add @antonk52 for bug, code, ideas, test

PR #223, opened 2020-04-27
Transform recursively, without JSON.parse()

dangodev:
Ditches the `JSON.parse()` approach which was limiting in 2 major ways:

- Couldn’t add `nullable` support (#112)
- Couldn’t add support for comments (now added! 🎉 )

At the end of the day, TypeScript isn’t JSON 🤷. And it turns out the original codebase was closer to what it needed to be. This PR refactors the output generation a little to be _even simpler_ than before.

### DIFF
--- a/src/v3.ts
+++ b/src/v3.ts
@@ -1,18 +1,13 @@
 import propertyMapper from "./property-mapper";
 import { OpenAPI3, OpenAPI3SchemaObject, SwaggerToTSOptions } from "./types";
 import {
-  escape,
-  fromEntries,
-  isAnyOfNode,
-  isArrayNode,
-  isObjNode,
-  isOneOfNode,
-  isSchemaObj,
-  makeOptional,
+  comment,
+  nodeType,
+  transformRef,
   tsArrayOf,
   tsIntersectionOf,
+  tsPartial,
   tsUnionOf,
-  unescape,
 } from "./utils";
 
 export const PRIMITIVES: { [key: string]: "boolean" | "string" | "number" } = {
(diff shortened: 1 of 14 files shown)


src/v3.ts
```diff
@@ -43,112 +38,98 @@ export default function generateTypesV3(
...
-      return node; // return by default
-    }
-  );
+      // 2. name (with “?” if optional property)
+      output += `"${key}"${!required || !required.includes(key) ? "?" : ""}: `;
+
+      // 3. open nullable
+      if (value.nullable) {
+        output += "(";
+      }
```
dangodev, line 112: cc @antonk52 sorry about doing this for you, but your PR pointed a flaw in the design. And I wanted to make sure I was solving for `nullable` (and it seems to be!).

So this _generates_, and it _looks_ correct, but I actually didn’t add tests in this PR making sure that `| null` works in all cases. Are nullable tests (and potentially fixing some edge cases) something you’re interested in writing in a followup PR? 👍 / 👎 ?

tests/bin/expected/petstore.ts
```diff
@@ -5,7 +5,17 @@
...
   schemas: {
-    Order: { id?: number; petId?: number; quantity?: number; shipDate?: string; status?: 'placed' | 'approved' | 'delivered'; complete?: boolean }
+    Order: {
+      id?: number
+      petId?: number
+      quantity?: number
+      shipDate?: string
+      /**
+       * Order Status
+       */
```
dangodev, line 15: This PR adds comment support! 🎉

tests/v3/expected/stripe.ts
```diff
@@ -5,106 +5,325 @@
...
+      /**
+       * The business type.
+       */
       business_type?:
-        | "company"
-        | "government_entity"
-        | "individual"
-        | "non_profit";
+        | ("company" | "government_entity" | "individual" | "non_profit")
+        | null;
```
dangodev, line 28: I _believe_ `nullable` has been added, but we can’t be sure before we write unit tests.

src/v3.ts
antonk52: No worries man, whatever gets the job done =)

IMHO I think the existing cases covered the nullable fairly well. If you can point out a few extra test cases where the explicit test cases are needed I have no problem sending a pr later on

dangodev: Yeah, fair. Thanks for understanding.

Also, a new version is up on npm with the changes, so I’d love to see if the new `@next` version works for your schemas :)

PR #252, opened 2020-05-18
Fix missing port bug

mmiszy:


drwpow:
@all-contributors please add @mmiszy for code

Nnoerregaard:
@drwpow any change we can get a latest tag on npm, which I can point npx at, until this is released?

drwpow:
Unfortunately, I still don’t have access to publish a release again. Hang tight; we’ll get it sorted soon. One way or another.

PR #258, opened 2020-05-20
Support for components.responses

tshelburne:
I'm using Typescript on both the client and server side to strictly type my API requests / responses, and the [Swagger components section](https://swagger.io/docs/specification/components/) includes suggested usage of the `components.responses` key.

This adds the support with as little ceremony as possible (responses isn't required, property mapping is only used for schemas, no changes to schemas usage, etc.). If you'd like it to be generalized a bit more to encourage mapping more pieces of the Swagger spec, happy to do so.

### DIFF
--- a/src/types/OpenAPI3.ts
+++ b/src/types/OpenAPI3.ts
@@ -8,6 +8,7 @@ export interface OpenAPI3 {
   openapi: string;
   components: {
     schemas: { [key: string]: OpenAPI3SchemaObject | OpenAPI3Reference };
+    responses?: { [key: string]: OpenAPI3SchemaObject | OpenAPI3Reference };
   };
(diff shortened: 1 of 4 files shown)


tshelburne:
@drwpow Is this likely to get merged? I'm working on publishing our Swagger docs, and I'd rather use the upstream lib than my own fork.

drwpow:
Hey @tshelburne! Thanks so much for putting this together. I’m on a break this week and next, and would be happy to review this as soon as I’m back (also sorry for not seeing this before I took a break). But at a cursory glance this looks perfect, and I’ll probably merge as-is when I’m back in front of a computer again.

drwpow:
@all-contributors please add @tshelburne for code, test

Issue #262, opened 2020-05-26
swagger-to-ts considers components mandatory


henhal:
When running `swagger-to-ts` on a specification without `components`, it immediately fails.
However, `components` is an [optional part of the api schema](https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.0/schema.yaml#L5) and I can think of at least two scenarios where it would not be present:

1. All schemas are inline in `paths`
2. All schemas are put in separate file(s) such as `schemas.yml`

My specifications are quite large and I tend to split them up into separate files as per 2). This means that operations reference schemas such as `{$ref: './schemas/foo.yml#/Foo'}`.
Neither of these two approaches seem to work with `swagger-to-ts` since it requires `components` to be present (and, I assume, also ignores any schemas outside of `components`)?

1. Would it be possible to invoke swagger-to-ts with a raw schema file, i.e., one that simply contains the schemas instead of requiring a full OpenAPI document?
2. Would it be possible to properly support OpenAPI documents without `components`? It should be possible to fetch all externally referenced schemas using standard json-ref-parser etc?

hasparus:
I encountered the same problem with openapi schema generated from Django :c

drwpow:
Just merged #236, so this will go out with the next release!

PR #263, opened 2020-05-26
Support raw schemas in addition to Swagger/OpenAPI documents

henhal:
This adds support for supplying raw v3 schemas or v2 definitions as an alternative to supplying full Swagger/OpenAPI documents, using an option `--raw-schema` in combination with `--version`.

Previously, swagger-to-ts would require `components` to be present in the given document, even though it's an optional property. When dealing with large documents, it's common practice to
split it over several files, putting e.g. schemas in separate files.

This change enables producing types for such stand-alone schemas by adding an option to parse the given file as a map of schemas  rather than a document. For this to work, the version must also be passed since it's not present in raw schemas/definitions.

For example, if the full document is:

```
openapi: 3.0.2
components:
  schemas:
    Foo:
      type: object
    Bar:
      type: object
      properties:
        foo: {$ref: '#/components/schemas/Foo'}
```

Then it's now possible to parse the schemas separately by passing a file with the following contents, and `--raw-schema --version 3` as options:

```
Foo:
  type: object
Bar:
  type: object
  properties:
    foo: {$ref: '#/Foo'}
```

For V3, the output differs from the output produced for OpenAPI3 documents in that
`schemas: { ... }` is the exported interface instead of
`components: { schemas: { ... } }`.

For V2, `definitions: { ... }` is still exported.

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -15,6 +15,8 @@ Options
   --help                display this
   --output, -o          specify output file
   --prettier-config     (optional) specify path to Prettier config file
+  --raw-schema          (optional) Read from raw schema instead of document
+  --version             (optional) Schema version (must be present for raw schemas)
 `,
   {
     flags: {
@@ -25,6 +27,12 @@ Options
       prettierConfig: {
         type: "string",
       },
+      rawSchema: {
+        type: "boolean"
+      },
+      version: {
+        type: "number"
+      }
     },
   }
 );
@@ -47,6 +55,8 @@ const timeStart = process.hrtime();
 
   const result = swaggerToTS(spec, {
     prettierConfig: cli.flags.prettierConfig,
+    rawSchema: cli.flags.rawSchema,
+    version: cli.flags.version
   });
 
   // Write to file if specifying output
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,13 @@
 import path from "path";
 import prettier from "prettier";
 import { swaggerVersion } from "./utils";
-import { OpenAPI2, OpenAPI3, SwaggerToTSOptions } from "./types";
+import {
+  OpenAPI2,
+  OpenAPI2Schemas,
+  OpenAPI3,
+  OpenAPI3Schemas,
+  SwaggerToTSOptions,
+} from "./types";
 import v2 from "./v2";
 import v3 from "./v3";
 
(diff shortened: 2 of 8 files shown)


henhal:
This relates to #262.

Note that no new test cases are added etc, as this was somewhat whipped together to get early feedback on if this is something you want at all. 
Possible improvements could include automatically detecting the type of passed document etc?

Anyway, all existing tests work and my manual tests of my schema files also produce correct output.
If interesting to proceed with, tests should of course be added.

henhal:
I thought some more about this, and while I await your feedback, I've realized that perhaps this tool should focus on parsing documents and not raw schemas. Also, it may be fine to not have  `components/schemas` in your document if you split it into several files, but the dereferenced/bundled single document will be very verbose with no re-use of schemas at all.
I've come to think that it would suit me better to use some pre-processor such as yaml-import to build my document from separate parts and still have components/schemas. 

However, I still think that swagger-to-ts should not enforce components being present since it is an optional property, and that it should instead also parse schemas from other parts of the document. Creating a separate issue with suggestions.

Howe

drwpow:
Hey thanks for the patience on this. I’ve left Manifold and lost access to this repo, and recently they’ve agreed to add me back as a maintainer.

This PR does introduce different code paths, and I’m a bit hesitant to add a flag that works for one version but not the other. And with the merging of #258 it does make the output logic more complicated. Is there any reason you can’t accomplish the same thing outside the library? e.g.:

```ts
type Schemas = OpenAPIV3['components']['schemas']
type Thing1 = Schemas['Thing1'];
```

Open to pushback on this; right now with limited time I’m trying to keep the codebase as simple as possible.

drwpow:
Sorry—I started with the PRs first and then read both issues you commented in. I understand your use case a bit better, so just ignore my previous comment.

I think I’m in favor of this PR now. I wanted to highlight the central goal of this project which is what I think you’re hinting at, which is **this project should convert Swagger to TypeScript 1:1 and not transform anything.** I don’t want to enable any options where the generated types deviate from your schema, and that was a realization in v1 of this project that introduced some big bugs.

With that said, you do make a pretty good case for adding this, because Swagger allows it. If Swagger allows it, so should this project 🙂. Sorry for the trouble, but if you could fix the merge conflicts from #258 I’d be happy to merge.

henhal:
Thanks @drwpow. I've fixed the merge conflicts, but I'm a bit short of time today so honestly I haven't had time to test this just yet. Ideally some unit tests should be added of course.

SmuliS:
I am also running into https://github.com/manifoldco/swagger-to-ts/pull/263. Looks like this PR is approved, so is there something that is holding back the merge if this?

drwpow:
Hey @SmuliS, thanks for checking in. Originally I was asking the original author to add some tests, but you raise a good point—better to release something without tests than not release something at all. Will merge this (and will add tests if I have time).

drwpow:
@all-contributors please add @henhal for code, docs

PR #264, opened 2020-06-02
fix url parse error when port specified

Hrysa:


drwpow:
Thank you! Unfortunately this was already fixed in #252. Will be released soon.

Issue #265, opened 2020-06-02
Cannot find name 'ref'


Diokuz:
I have an OpenAPI json like that:
```json
{
  "swagger": "2.0",
  "definitions": {
    "Messages": {
      "type": "object",
      "additionalProperties": {
        "$ref": "#/definitions/Message"
      }
    },
    "Message": {
      "type": "object",
      "properties": {
        "code": {
          "type": "integer"
        },
        "text": {
          "type": "string"
        }
      }
    }
  }
}
```

Which results in typescript:

```ts
export interface definitions {
  Messages: { [key: string]: ref }
  Message: { code?: number; text?: string }
}
```

It is not valid, because of `ref`. What is `ref`, and why it is not defined?

Diokuz:
AFAIK should be
```ts
export interface definitions {
  Messages: { [key: string]: definitions['Message'] }
  Message: { code?: number; text?: string }
}
```

Diokuz:
Looks like it was broken here https://github.com/manifoldco/swagger-to-ts/pull/223/files#diff-92e97370f1732d8782fd35b2cfaf277bL59

wmonk:
Also facing this issue. Any ideas on a timeframe for a fix?

pedrotorchio:
Same issue

lukeshiru:
Same problem, here!

jcf120:
Yep, same here

drwpow:
Fixed (hopefully) in #266

PR #266, opened 2020-06-03
additionalProperties interpolation, primitive enums

skh-:
if `additionalProperties` is not of a primitive type, the generated file will contain things like `ref` or `array` instead of interpolating that ref or array.  this PR should fix that and #265.

not aware of any side effects, let me know if you can think of any.

### DIFF
--- a/examples/stripe-openapi2.ts
+++ b/examples/stripe-openapi2.ts
@@ -678,83 +678,83 @@ export interface definitions {
       | "unapplied_from_invoice"
       | "unspent_receiver_credit";
   };
-  deleted_account: { deleted: "true"; id: string; object: "account" };
+  deleted_account: { deleted: true; id: string; object: "account" };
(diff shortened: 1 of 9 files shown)


skh-:
had another problem where:

```
type: number
enum:
  - -1
  - 0
  - 1
```

would produce `"-1" | "0" | "1"` when I needed `-1 | 0 | 1`.  

Added fix to this PR.

FourF-src:
https://github.com/manifoldco/swagger-to-ts/issues/278

drwpow:
@all-contributors please add @skh- for code, test

skh-:
@drwpow if you would like maintenance help with the repo feel free to tap me 😄 .

drwpow:
@skh- awesome! Yes I would… as soon as I’m able to get access to the repo again (still locked out of npm deployments and adding contributors at the moment).

PR #288, opened 2020-07-23
Improve error message when server is unreachable

selbekk:
This change improves the error message for when the remote file you're trying to access is unreachable. This can happen if you misspelled the URL, the server is down for some reason or if you don't have access to the internet.


drwpow:
@all-contributors add @selbekk for code

PR #285, opened 2020-07-16
Add support for http ports

vansosnin:
Add support for ports. I have a case: run swagger on localhost:5000. And now I have an error:

```
❌ "Error: getaddrinfo ENOTFOUND localhost:5000"
Error: � version missing from schema; specify whether this is OpenAPI v3 or v2 https://swagger.io/specification
    at swaggerVersion (C:\Users\isosnin\AppData\Roaming\npm-cache\_npx\38628\node_modules\@manifoldco\swagger-to-ts\dist-node\index.js:126:9)
    at swaggerToTS (C:\Users\isosnin\AppData\Roaming\npm-cache\_npx\38628\node_modules\@manifoldco\swagger-to-ts\dist-node\index.js:352:19)
    at C:\Users\isosnin\AppData\Roaming\npm-cache\_npx\38628\node_modules\@manifoldco\swagger-to-ts\bin\cli.js:48:18
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

This PR fixes issue.

PR #289, opened 2020-07-23
Add support for non-standard ports

### LINKED ISSUE #267: Doesn't work on non-standard ports
At least on macOS platform, running node v14.3.0, using a url such as http://192.168.7.245:8091/doc/swagger/swagger.yaml results in an error "Error: getaddrinfo ENOTFOUND 192.168.7.245:8091" - clearly, some code that is attempting to split the url up is not splitting the port off the host part before doing a lookup.

selbekk:
This makes it possible to connect to i.e. localhost:8080 to generate the types.

Fixes #267


selbekk:
Ah duplicate of #285. Closing.

PR #310, opened 2020-08-18
Handle quotes in enum values

### LINKED ISSUE #256: Enum value with single quote causes syntax error
Hello,

I've encountered a bug where if an enum value contains a single quote, a syntax error is thrown.

```json
            "enum": [
              "",
              "Cash",
              "Insurance",
              "Insurance Out of Network",
              "Auto Accident",
              "Worker's Comp"
            ],
```

https://app.drchrono.com/openapi-schema has a download link for the schema which contains this issue.

```shell
Loading spec from .\openapi-schema.json…
SyntaxError: ')' expected. (1796:123)
  1794 |   * One of `""`, `"Cash"`, `"Insurance"`, `"Insurance Out of Network"`, `"Auto Accident"` or `"Worker's Comp"`.<br>**Note:** Patient must already have either `primary_insurance` or `secondary_insurance` or new `primary_insurance` or `secondary_insurance` is passed in request if `Insurance`, `Auto Accident` or `Worker's Comp` payment profiles are chosen.
  1795 |   */
> 1796 | "patient_payment_profile"?: ('') | ('Cash') | ('Insurance') | ('Insurance Out of Network') | ('Auto Accident') | ('Worker's Comp');
```

scastiel:
Hi! I propose this fix for issue #256, about handling the single quotes in enum values.
Hope it helps!
Sebastien.

Closes #256.

### DIFF
--- a/tests/v2/specs/petstore.yaml
+++ b/tests/v2/specs/petstore.yaml
@@ -596,9 +596,9 @@ definitions:
         type: "string"
         description: "Order Status"
         enum:
-          - "placed"
-          - "approved"
-          - "delivered"
+          - "it's placed"
+          - "it's approved"
+          - "it's delivered"
       complete:
         type: "boolean"
         default: false
--- a/src/v2.ts
+++ b/src/v2.ts
@@ -59,7 +59,7 @@ export default function generateTypesV2(
           (node.enum as string[]).map((item) =>
             typeof item === "number" || typeof item === "boolean"
               ? item
-              : `'${item}'`
+              : `'${item.replace(/'/g, "\\'")}'`
           )
         );
       }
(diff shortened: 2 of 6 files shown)


tests/v2/specs/petstore.yaml
```diff
@@ -596,9 +596,9 @@ definitions:
         type: "string"
         description: "Order Status"
         enum:
-          - "placed"
-          - "approved"
-          - "delivered"
```
drwpow, line 601: Thanks so much for adding this in a test, but would you mind instead adding unit tests to:
- `tests/v2/index.test.ts`
- `tests/v3/index.test.ts`

This `petstore.yaml` is a direct copy-paste of Swagger’s example that’s not meant to be edited, and these snapshots are so easy to blow away and lose this test in the process. This is more of a “pass/fail” test than testing specific behavior like what you added—that’s what those spec tests are for.

Issue #236, opened 2020-05-05
Add paths support


drwpow:
Suggested in #112:

> any plans for `components.parameters`?

Looking into it, `parameters` is part of `components.paths`, which is the same as Swagger 2.0’s `paths`. So adding support for v3 should also include v2. That said… what should the schema look like for this? I propose something like :

```ts
interface components {
  paths: {
    "/v1/checkout/sessions": {
      get: {
        parameters: {
          ending_before?: string;
          expand?: string[];
        }
      }
    }
  }
}
```

Aliasing that would look like;

```
type parameters = components["paths"]["/v1/checkout/sessions"]["get"]["parameters"];
```

This ticket is a proposal for implementation. This looks the most straightforward to me, but are there any other considerations when it comes to `paths`? Is this usage intuitive and does this work for most people?

I’d love to hear stories from using `paths` to confirm if this would help or if any additional info would be helpful.

henhal:
I would really appreciate support for paths, however, putting them under components seems odd since that's not where they originate from in the document?

Also, parameters can be `header`, `query`, `path` or `cookie`. It doesn't make sense to create one `parameters` type with all of them if the purpose is to be able to process headers in a request handler etc?

Let's consider a typical OpenAPI3 document:

```
  components:
    parameters:
      Foo:
        name: x-foo
        in: header
        required: true
        schema:
          type: string
    schemas:
      Person:
        type: object
        additionalProperties: false
        required: [id, name]
        properties:
          id:
            type: string
          name:
            type: string
          age:
            type: number
      Error:
        type: object
        additionalProperties: false
        required: [message]
        properties:
          message:
            type: string
  paths:
    /person:
      post:
        operationId: addPerson
        parameters:
          - $ref: '#/components/parameters/Foo'
          - name: x-bar
            in: header
            schema:
              type: string
          - name: baz
            in: query
            schema:
              type: boolean
        requestBody:
          content:
            application/json:
              schema:
                type: object
                additionalProperties: false
                required: [name]
                properties:
                  name:
                    type: string
                  age:
                    type: number
        responses:
          200:
            headers:
              x-baz:
                schema:
                  type: string
            content:
              application/json:
                schema:
                  $ref: '#/components/schemas/Person'
          400:
            content:
              application/json:
                schema:
                  $ref: '#/components/schemas/Error'
```

[...]

PR #328, opened 2020-10-08
Add paths support

drwpow:
Satisfies #236. Though this doesn’t have 100% of all paths features, it’s at least a start 🙂 

cc @gr2m (wanted to tag you to let you know this is done, and will be ready for testing in the next release!)


src/v3.ts
```diff
@@ -154,27 +153,107 @@ export default function generateTypesV3(
...
+        // handle parameters
+        if (responses.parameters) {
+          output += `parameters: {\n`;
+          const allParameters: Record<
+            string,
+            Record<string, OpenAPI3Parameter>
+          > = {};
+          responses.parameters.forEach((p) => {
+            if (!allParameters[p.in]) allParameters[p.in] = {};
+            // TODO: handle $ref parameters
```
drwpow, line 173: `parameters` can be a `$ref` rather than an object. The first cut doesn’t handle that (yet)

src/v3.ts
```diff
@@ -154,27 +153,107 @@ export default function generateTypesV3(
...
+    finalOutput += `export interface paths {
+  ${transformPaths(paths)}
+}
+
+`;
+  }
+
+  finalOutput += "export interface components {\n";
+
+  // TODO: handle components.parameters
```
drwpow, line 241: We also don’t yet handle `components.parameters`, which would be needed for `$ref` parameters (above)

gr2m:
fyi I'm currently exploring generating types from GitHub's OpenAPI spec and will send pull requests to address the bugs I've found. See https://github.com/gr2m/github-openapi-types/pull/1

PR #334, opened 2020-10-19
fix: set responses without content schemas to `never` or `unknown`, depending on status code

### LINKED ISSUE #333: set empty response type to `never`
If a response has no content key, e.g. 

```json
"responses": {
  "204": {
    "description": "Empty response"
  }
}
```

The resulting type is currently `any`: 

```ts
responses: {
  /**
   * Empty response
   */
  "204": any;
};
```

But I'd argue the better type would be `never`

gr2m:
fixes #333

### DIFF
--- a/tests/v3/index.test.ts
+++ b/tests/v3/index.test.ts
@@ -630,4 +630,78 @@ describe("OpenAPI3 features", () => {
       }`)
     );
   });
+
+  it("empty responses (#333)", () => {
+    const schema: OpenAPI3 = {
+      openapi: "3.0.1",
+      paths: {
+        "/no-content": {
+          get: {
+            responses: {
+              "204": {
+                description: "Empty response",
+              },
+            },
+          },
+        },
+        "/not-modified": {
+          get: {
+            responses: {
+              "304": {
+                description: "Empty response",
+              },
+            },
+          },
+        },
+        "/not-found": {
+          get: {
+            responses: {
+              "404": {
+                description: "Empty response",
+              },
+            },
+          },
+        },
+      },
+    };
+
+    expect(swaggerToTS(schema)).toEqual(
+      format(`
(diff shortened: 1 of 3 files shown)


src/types/OpenAPI3.ts
```diff
@@ -32,7 +32,7 @@ export interface OpenAPI3Parameter {
 
 export interface OpenAPI3ResponseObject {
   description?: string;
-  content: {
+  content?: {
```
drwpow, line 35: 👍🏻

tests/bin/expected/petstore.ts
```diff
@@ -403,11 +403,11 @@ export interface paths {
         /**
          * Invalid username supplied
          */
-        '400': any
+        '400': unknown
         /**
          * User not found
          */
-        '404': any
+        '404': unknown
```
drwpow, line 410: 👍🏻 I’m OK with this change. Really, any undeclared type should be “dangerous” and I do like that this will surface more errors than `any`

tests/v3/index.test.ts
```diff
@@ -630,4 +630,78 @@ describe("OpenAPI3 features", () => {
       }`)
     );
   });
+
+  it("empty responses (#333)", () => {
```
drwpow, line 634: Nice unit test!

Currently the strategy has been to rely on unit tests as the real core of testing, but the full-schema “snapshots” are just ensuring that this actually works on real-world schemas. Glad you were able to figure it out!

The unit tests themselves could probably be better-organized; I’m not that picky. Open to feedback there. As long as they’re there (like the great ones you added), I’m happy!

drwpow (approved):
This looks great! Thank you

drwpow:
@all-contributors please add @gr2m for code, test

PR #336, opened 2020-10-19
style: make eslint happy

gr2m:
both my PRs #334 and #335 failed because of a new eslint rule. I thought you might want to merge this in first in case the PRs get blocked. I added the same commit to the PRs to get them to green

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -8,7 +8,7 @@ export function comment(text: string): string {
 }
 
 /** shim for Object.fromEntries() for Node < 13 */
-export function fromEntries(entries: [string, any][]): object {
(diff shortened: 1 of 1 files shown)


drwpow (approved):
Thank you!

I have been on-the-fence about the strictness of the ESLint rules. I do feel that they can help beginners as they encourage good habits (and beginners are what I want to optimize for). But also, no one has fun debugging ESLint. Either way, thanks for making this change, and happy for feedback on the ESLint rules in general.

gr2m:
I just use prettier for formatting these days. I really like the setup in snowpack, where an action just fixes formatting after merging, it's not even part of the CI because it doesn't have to. And when using TypeScript, many errors that eslint would catch are caught by tsc

PR #343, opened 2020-10-23
Improve CONTRIBUTING.md

drwpow:
Add note on regenerating schemas

cc @gr2m


drwpow:
(moved to #342 after updating the default branch to `main`)

PR #344, opened 2020-10-23
Improve CONTRIBUTING.md

drwpow:
Add note on regenerating schemas cc @gr2m

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -37,7 +37,9 @@ Options
   }
 );
 
-console.info(chalk.bold(`✨ openapi-typescript 2.3`));
+console.info(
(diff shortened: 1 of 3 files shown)


bin/cli.js
```diff
@@ -37,7 +37,7 @@ Options
   }
 );
 
-console.info(chalk.bold(`✨ openapi-typescript 2.2`));
+console.info(chalk.bold(`✨ openapi-typescript 2.3`));
```
gr2m, line 40: unrelated, but why not read the version out of package.json?

gr2m (approved):
Looks good to me 👍🏼

PR #348, opened 2020-11-04
Set index type to all possible values

gr2m:
Here is a minimal test case to reproduce the problem: [TypeScript Playground](
https://www.typescriptlang.org/play?#code/C4TwDgpgBAKhDOwoF4oG8BQVtQJb3gFcEB+ALikQCdcA7AcwG4scBjACwlYGt5zLgNBsxxQAthGABDACZTp-anSYtsrAPa1gELXwpLhqqDIhgANupATdiwcpE4A2twgh9dhgF13QlQF8gA)

This feels like a potential TypeScript bug to me, what do you think? But basically this comes down to:

- If a schema is of type `object`
- and it has `additionalProperties`
- then instead of adding `[key: string]: string;` to the object, we should create an intersection with a separate `{ [key: string]: string; }` object

Result: [TypeScript Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAKhDOwoF4oG8BQVtQJb3gFcEB+ALikQCdcA7AcwG4scBjACwlYGt5zLgNBsxxQAthGABDACZTp-anSYtsrAPa1gELXwpLhqqDIhgANupATdiwcuYBfKADJ0UANrcIIfXYYBdXyEmKAcgA)

### DIFF
--- a/src/v3.ts
+++ b/src/v3.ts
@@ -92,18 +92,21 @@ export default function generateTypesV3(
 
         let properties = createKeys(node.properties || {}, node.required);
 
-        // if additional properties, add to end of properties
-        if (node.additionalProperties) {
-          properties += `[key: string]: ${
-            node.additionalProperties === true
-              ? "any"
-              : transform(node.additionalProperties) || "any"
-          };\n`;
-        }
+        // if additional properties, add an intersection with a generic map type
+        const additionalProperties = node.additionalProperties
+          ? [
+              `{ [key: string]: ${
+                node.additionalProperties === true
+                  ? "any"
+                  : transform(node.additionalProperties) || "any"
+              };}\n`,
+            ]
+          : [];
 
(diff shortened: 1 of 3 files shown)


gr2m:
@drwpow this is ready for review. It's currently blocking me from adapting it for Octokit: https://github.com/octokit/types.ts/pull/197

drwpow:
Interesting! Yes this does seem like a bug; to me there doesn’t seem to be any obvious reason why the intersection would work better. But that said I’m all for it if it improves things. Thanks!

gr2m:
Thanks Drew! Could you please publish a new version to npm?

drwpow:
`2.3.1` published!

drwpow:
@allcontributors please add @gr2m for bug

drwpow:
(i forgot)

PR #365, opened 2020-11-12
fix: escape quotes in enum string values

### LINKED ISSUE #256: Enum value with single quote causes syntax error

gr2m:
I've run into this problem when updating to the latest GitHub schema, this PR reproduces the problem the same way. I'm looking into fixing it

fixes #256

### DIFF
--- a/src/v3.ts
+++ b/src/v3.ts
@@ -73,7 +73,7 @@ export default function generateTypesV3(
           (node.enum as string[]).map((item) =>
             typeof item === "number" || typeof item === "boolean"
               ? item
-              : `'${item}'`
+              : `'${item.replace(/'/g, "\\'")}'`
           )
         );
       }
--- a/tests/v3/expected/github.ts
+++ b/tests/v3/expected/github.ts
@@ -13,6 +13,10 @@ export interface paths {
   "/app-manifests/{code}/conversions": {
     post: operations["apps/create-from-manifest"];
   };
+  "/app/hook/config": {
+    get: operations["apps/get-webhook-config-for-app"];
+    patch: operations["apps/update-webhook-config-for-app"];
+  };
   "/app/installations": {
     get: operations["apps/list-installations"];
   };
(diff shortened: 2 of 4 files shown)


src/v3.ts
```diff
@@ -73,7 +73,7 @@ export default function generateTypesV3(
           (node.enum as string[]).map((item) =>
             typeof item === "number" || typeof item === "boolean"
               ? item
-              : `'${item}'`
+              : `'${item.replace(/'/g, "\\'")}'`
```
drwpow, line 76: This makes sense to me. In my experience, you should never be too aggressive escaping quotes like this, as you may end up with more things escaped than anticipated.

drwpow (approved):

drwpow:
I believe this fixes #256?

gr2m:
> I believe this fixes #256?

ah yes, thanks!

drwpow:
`2.4.1` released!

Issue #360, opened 2020-11-12
TypeError: Cannot convert undefined or null to object


rendall:
**Summary**: TypeError when attempting to convert valid spec

**Detail**: This spec passes validity tests <https://github.com/rendall/simple-comment/blob/master/src/schema/simple-comment-api.json> but attempting to convert it yields the following error (with stack trace):

```
✨ openapi-typescript 2.4.0
🤞 Loading spec from https://raw.githubusercontent.com/rendall/simple-comment/master/src/schema/simple-comment-api.json…
TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at transformOperation (...\npm-cache\_npx\13920\node_modules\openapi-typescript\dist-node\index.js:428:12)
    at ...\npm-cache\_npx\13920\node_modules\openapi-typescript\dist-node\index.js:461:39
    at Array.forEach (<anonymous>)
    at ...\npm-cache\_npx\13920\node_modules\openapi-typescript\dist-node\index.js:452:31
    at Array.forEach (<anonymous>)
    at transformPaths (...\Roaming\npm-cache\_npx\13920\node_modules\openapi-typescript\dist-node\index.js:450:27)
    at generateTypesV3 (...\npm-cache\_npx\13920\node_modules\openapi-typescript\dist-node\index.js:485:5)
    at swaggerToTS (...\npm-cache\_npx\13920\node_modules\openapi-typescript\dist-node\index.js:545:32)
    at ...\npm-cache\_npx\13920\node_modules\openapi-typescript\bin\cli.js:56:18
```

**Reproduce**: 

`npx openapi-typescript https://raw.githubusercontent.com/rendall/simple-comment/master/src/schema/simple-comment-api.json --output schema.ts`

**Expected**:

Either conversion or helpful error message

**Compliment, pep talk**:

It's just great that you all have made this, and I want to congratulate you on making such a useful tool!

gr2m:
Bug seems to be that we do not handle the `"summary"` key on the paths level. We currently only ignore the `parameters` key: https://github.com/drwpow/openapi-typescript/blob/1a74e25b496ffa072c912a3fa57bb66462dbf0ca/src/v3.ts#L269 and don't do any further checks whether the `operation` value from line 528 is an actual operation.

I think we should make a hardcoded list of all http methods and iterate through them.

[...]

PR #366, opened 2020-11-13
fix: check if path item is method operation

rendall:
Fixes issue #360

* Adds a check `isMethod` in `transformPaths` in [./src/v3.ts](./src/v3.ts)
* Adds a new `OpenAPI3PathItemObject` type in [./src/types/OpenAPI3.ts](./src/types/OpenAPI3.ts)
* This added type `OpenAPI3PathItemObject` is per "PathItemObject" spec <https://swagger.io/specification/#path-item-object>
* Replaces the key-value pair in `OpenAPI3Paths`
* `OpenAPI3PathItemObject` *is* missing the `servers` field from the specs
* Variable name bikeshedding - changed `method` to `field` since it *might not* be a method

*Edit*: updated type to `OpenAPI3PathItemObject` from `OpenAPI3PathItem`

### DIFF
--- a/src/types/OpenAPI3.ts
+++ b/src/types/OpenAPI3.ts
@@ -12,20 +12,20 @@ export interface OpenAPI3Paths {
   [path: string]: Partial<OpenAPI3PathItem>;
 }
 
-export type OpenAPI3PathItem= {
-  $ref: string
-  summary: string
-  description: string
-  get: OpenAPI3Operation
-  put: OpenAPI3Operation
-  post: OpenAPI3Operation
-  delete: OpenAPI3Operation
-  options: OpenAPI3Operation
-  head: OpenAPI3Operation
-  patch: OpenAPI3Operation
-  trace: OpenAPI3Operation
-  parameters: Parameter[]
-}
+export type OpenAPI3PathItem = {
+  $ref: string;
+  summary: string;
+  description: string;
+  get: OpenAPI3Operation;
+  put: OpenAPI3Operation;
+  post: OpenAPI3Operation;
+  delete: OpenAPI3Operation;
+  options: OpenAPI3Operation;
+  head: OpenAPI3Operation;
+  patch: OpenAPI3Operation;
+  trace: OpenAPI3Operation;
+  parameters: Parameter[];
+};
 
 export interface OpenAPI3Operation {
   operationId?: string;
--- a/src/types/OpenAPI3.ts
+++ b/src/types/OpenAPI3.ts
@@ -24,7 +24,6 @@ export type OpenAPI3PathItem= {
   head: OpenAPI3Operation
   patch: OpenAPI3Operation
   trace: OpenAPI3Operation
-  servers: any
   parameters: Parameter[]
 }
 
--- a/src/types/OpenAPI3.ts
+++ b/src/types/OpenAPI3.ts
@@ -9,13 +9,28 @@ export interface OpenAPI3Schemas {
 }
 
 export interface OpenAPI3Paths {
-  [path: string]: {
-    [method: string]: OpenAPI3Operation | Parameter[];
-  };
+  [path: string]: Partial<OpenAPI3PathItem>;
+}
+
+export type OpenAPI3PathItem= {
+  $ref: string
+  summary: string
+  description: string
+  get: OpenAPI3Operation
+  put: OpenAPI3Operation
+  post: OpenAPI3Operation
+  delete: OpenAPI3Operation
+  options: OpenAPI3Operation
+  head: OpenAPI3Operation
+  patch: OpenAPI3Operation
+  trace: OpenAPI3Operation
+  servers: any
+  parameters: Parameter[]
 }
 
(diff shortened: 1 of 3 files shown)


src/types/OpenAPI3.ts
```diff
@@ -9,13 +9,27 @@ export interface OpenAPI3Schemas {
...
+  get: OpenAPI3Operation;
+  put: OpenAPI3Operation;
+  post: OpenAPI3Operation;
+  delete: OpenAPI3Operation;
+  options: OpenAPI3Operation;
+  head: OpenAPI3Operation;
+  patch: OpenAPI3Operation;
+  trace: OpenAPI3Operation;
+  parameters: Parameter[];
+};
```
gr2m, line 28: nice. I wonder why you didn't made all the keys optional and instead used the `Partial<>` method above? Is there any benefit in that approach? Because I don't see the `OpenAPI3PathItem` type by itself to ever be valid.

tests/v3/index.test.ts
```diff
@@ -988,4 +988,154 @@ describe("OpenAPI3 features", () => {
...
+        schemas: {
+          ErrorResponse: { error: string; message: string };
+          SearchResponse: { title: string; date: string }
+        }
+        responses: {
+          NotFound: { [key: string]: any }
+        }
+      }`)
+    );
+  });
```
gr2m, line 1140: can you reduce the test case here? Having only the `/` path should be sufficient, you can remove the `/search` bits?

gr2m (commented):
Great PR overall!

src/types/OpenAPI3.ts
rendall: This is a good point and I don't think I have a good enough reason to push back on it. I will update the PR with this change.

rendall: This is updated with the [last push here](https://github.com/drwpow/openapi-typescript/pull/366/files/9b31327b90c39d899b788a974e0f7f80d6f045fa..efed10f2a66e1cf1385de147ed01b784e8995b72)

tests/v3/index.test.ts
rendall: [Done!](https://github.com/drwpow/openapi-typescript/pull/366/files/9b31327b90c39d899b788a974e0f7f80d6f045fa..efed10f2a66e1cf1385de147ed01b784e8995b72#diff-f6cd969653f24a9df9131814e4643dc857e0044c9d415cd41445030dafda1e25)

I added a "description" check to 'get' for completeness

src/types/OpenAPI3.ts
gr2m: cheers 👍🏼

tests/v3/index.test.ts
```diff
@@ -988,4 +988,95 @@ describe("OpenAPI3 features", () => {
...
+        responses: {
+          NotFound: {
+            content: {
+              "application/json": {
+                schema: { $ref: "#/components/schemas/ErrorResponse" },
+              },
+            },
+          },
+        },
+      },
```
gr2m, line 1049: I think you don't use `components` at all and can just leave it out from the schema?

rendall: You are correct! I was not paying attention.  I'll pare it down to the minimum necessary.

rendall: [There we are](https://github.com/drwpow/openapi-typescript/pull/366/files/efed10f2a66e1cf1385de147ed01b784e8995b72..2ad93b18e57f6bb542240bbc5895fbfb70ddb3a9). Much nicer, I think

gr2m (approved):
looks good to me 👍🏼 

ping @drwpow for 2nd review, I'm new here :D

src/types/OpenAPI3.ts
drwpow: This change looks good to me! In general it’s probably best to err on the side of making too many properties optional. In my experience even if the official OpenAPI spec marks something as required, almost all generators/parsers will still work, so this library should too (I think of it as a loose spec).

Unrelated additional context: we only have partial types in this repo, and only the ones this library knows about, for 2 reasons:

1. I’ve found some projects with OpenAPI TypeScript types but the ones I looked at are incomplete or inaccurate in some way (if some exist that work well I’d be open to it)
2. In some instances we may want to deviate from the spec, such as making something optional even if it’s not, so we can allow more generation.

Sorry; a very long way to say “LGTM” 😄

tests/v3/index.test.ts
```diff
@@ -988,4 +988,39 @@ describe("OpenAPI3 features", () => {
...
+            responses: { }
+          }
+        };
+      }
+
+      export interface operations {}
+        
+      export interface components {}`)
+    );
+  });
```
drwpow, line 1025: Nice test!

drwpow (approved):

drwpow:
@all-contributors please add @rendall for code, bug, test

drwpow:
Released in `v2.4.2`! 🚀

PR #369, opened 2020-11-18
Fixing 'content' being expected to be present.

robertmassaioli:
I spotted that this caused failures for our OpenAPI files being generated.

In general, I think that `OpenAPI3.ts` should be replaced with something like https://www.npmjs.com/package/atlassian-openapi so that the types are more reliable and thus the logic is more correct.

### DIFF
--- a/src/types/OpenAPI3.ts
+++ b/src/types/OpenAPI3.ts
@@ -43,7 +43,7 @@ export interface OpenAPI3ResponseObject {
 
 export interface OpenAPI3RequestBody {
   description?: string;
-  content: {
+  content?: {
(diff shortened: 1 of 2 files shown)


drwpow (approved):
Thank you for the fix!

I am open to using some library like the one you suggested for the types. At a quick glance, the one suggested does deviate from ours in one area I noticed: it requires `paths`. While that does match the schema, we want this library to be more lenient and allow for generation of partial schemas (it’s this project’s goal to generate types, not validate your schema). I haven’t found one yet that meets our needs, but I’m always on the lookout for one.

drwpow:
@all-contributors please add @robertmassaioli for code, bug

drwpow:
Released in `2.4.2`! 🚀

PR #378, opened 2020-11-23
fix: default path/query types to string

### LINKED ISSUE #377: SyntaxError: Type Expected
Hello, I'm trying to generate schema from Discourse openapi file https://docs.discourse.org/ but I get following error:

image

Any ideas? Thanks

Please note: Postman was able to load this file

MikeYermolayev:
This PR adds failing test for #377 
I faced this issue with a real-world openapi schema (discourse.org)

I'm not sure anymore if this even should be fixed but having a better error message for bad schemas would be neat.

fixes #377

### DIFF
--- a/tests/v3/index.test.ts
+++ b/tests/v3/index.test.ts
@@ -1014,4 +1014,52 @@ describe("OpenAPI3 features", () => {
       export interface components {}`)
     );
   });
+  it("missing schema in parameteres (#377)", () => {
+    const schema: OpenAPI3 = {
+      openapi: "3.0.1",
+      paths: {
+        "/test/{test_id}": {
+          get: {
+            summary: "some summary",
+            description: "some description",
+            parameters: [
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore
+              {
+                name: "id",
+                in: "path",
+                required: true,
+              },
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore
+              {
+                name: "page",
+                in: "query",
+                required: false,
+              },
+            ],
+            responses: {},
+          },
+        },
+      },
+    };
+
+    expect(swaggerToTS(schema)).toBe(
+      format(`
+      export interface paths {
+        '/test/{test_id}': {
+         /**
+          * some description
+          */
+         get: {
+            responses: { }
+          }
+        };
+      }
+
+      export interface operations {}
+
+      export interface components {}`)
+    );
+  });
 });


tests/v3/index.test.ts
```diff
@@ -1011,6 +1011,54 @@ describe("OpenAPI3 features", () => {
...
+      paths: {
+        "/test/{test_id}": {
+          get: {
+            summary: "some summary",
+            description: "some description",
+            parameters: [
+              // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+              // @ts-ignore
+              {
+                name: "id",
```
gr2m, line 1029: I think you either have to change the path to `"/test/{id}"` or change this name to `"test_id"`?

tests/v3/index.test.ts
```diff
@@ -1011,6 +1011,54 @@ describe("OpenAPI3 features", () => {
...
+    );
+  });
+  it("missing schema in parameteres (#377)", () => {
+    const schema: OpenAPI3 = {
+      openapi: "3.0.1",
+      paths: {
+        "/test/{test_id}": {
+          get: {
+            summary: "some summary",
+            description: "some description",
```
gr2m, line 1024: you can remove these to simplify the test case. We test these properties in other places

tests/v3/index.test.ts
```diff
@@ -1011,6 +1011,54 @@ describe("OpenAPI3 features", () => {
...
+
+    expect(swaggerToTS(schema)).toBe(
+      format(`
+      export interface paths {
+        '/test/{test_id}': {
+         /**
+          * some description
+          */
+         get: {
+            responses: { }
```
gr2m, line 1055: you can add the expected types here:

```suggestion
          parameters: {
            path: {
              test_id: string
            },
            query: {
              page?: string
            },
          },
            responses: { }
```

gr2m:
Tests look good, just left a few comments. I think it's a good idea to default the schema for path/query parameters to string, that will resolve your problem at #377

Next step is to make the test pass :)

drwpow:
@MikeYermolayev thanks so much for adding this PR! As @gr2m said, happy to merge this once we get the tests working again.

Also this is 100% optional, but we could also add the entire Discourse schema to this repo, too, to make sure the whole thing builds.

The tests are 2-tiered:

1. unit tests test the individual parts of the schema, like the one you just added. These are the most important parts
2. less important, but still valid, are whole-schema integration tests. These just ensure that this library is working as intended with real-world schemas.

Again, once everything is passing we can merge.

MikeYermolayev:
Okay, I'll try to come back to this during the week. As I figured out the whole discourse schema is a bit flawed and so it's useless as a source for type-gen since actual responses differ from the schema, so I'm not going to generate types from it anymore.
But still can cover this case

drwpow:
@MikeYermolayev thanks so much for your work on this. Unfortunately the interface / fix has changed a bit since this was added. #440 finishes this work for the updated generation and adds a test. This will be fixed in the next release! 🎉

PR #381, opened 2020-11-23
make status codes numeric

### LINKED ISSUE #380: Numeric status codes
Hello, I really like this library but I had to fork it to be able to actually use it.

My issue is that the `responses` object lists the response status codes as strings. I need to be able to use numeric values in my codebase and I am not able to check them against the generated TS types.

---

My solution was to replace the `{ [statusCode]: responseBodies }` object with a discriminated union:

```ts
responses:
  | { statusCode: 200, bodies: … }
  | { statusCode: 202, bodies: … }
  | { statusCode: 400, bodies: … }
```

I can get a numeric enum from that simply by using `responses['statusCode']`.

---

I've also noticed that TS allows actual numeric keys such as `{ 200: … }` (instead of `{ "200": … }`) so that could be enough.

Would such a change be acceptable? What is the actual reason the keys are strings anyway?

jankuca:
Closes #380

### DIFF
--- a/examples/stripe-openapi3.ts
+++ b/examples/stripe-openapi3.ts
@@ -813,7 +813,7 @@ export interface operations {
       /**
        * Successful response.
        */
-      "200": {
+      200: {
         "application/json": components["schemas"]["three_d_secure"];
       };
       /**
@@ -846,7 +846,7 @@ export interface operations {
       /**
        * Successful response.
        */
-      "200": {
+      200: {
         "application/json": components["schemas"]["three_d_secure"];
       };
       /**
@@ -872,7 +872,7 @@ export interface operations {
       /**
        * Successful response.
        */
-      "200": {
+      200: {
         "application/json": components["schemas"]["deleted_account"];
       };
       /**
@@ -902,7 +902,7 @@ export interface operations {
       /**
        * Successful response.
        */
-      "200": {
+      200: {
         "application/json": components["schemas"]["account"];
       };
       /**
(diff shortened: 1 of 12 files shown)


gr2m (commented):
great PR! @drwpow if you agree with the change, I think we should release this as a breaking change.

I'll test the change against my own work with @octokit to see if I run into any trouble, as I ran into the same problem you did

gr2m (approved):
This changes works well for my usage in @octokit. Accessing the responses works the same using `...["responses"]["200"]`. The only difference is that `keyof paths["..."]["get"]["responses"]` returns `200 | 404` instead of `"200" | "404"`

I say let's go ahead with it. I'm not even sure if I would consider the change a breaking change, but when in doubt I'd always rather go with a breaking change.

drwpow (approved):

drwpow:
> I'm not even sure if I would consider the change a breaking change, but when in doubt I'd always rather go with a breaking change.

Yeah I could go either way here. I think it technically is a breaking change, but it’s a breaking change on a newly-released feature. I agree with erring on the side of breaking change, and cutting `3.0` for this

drwpow:
@all-contributors please add @jankuca for code, test

drwpow:
OK I tagged @ released `3.0.0-rc.0` with these changes. If y’all have time to test the changes that’d be great! Overall I feel pretty confident in it, but just in case wanted to just kick the tires a bit before releasing `3.0.0`.

PR #382, opened 2020-11-23
apply prettier

jankuca:
I've noticed prettier is not being followed. This PR formats all JS/TS files according to the prettier config and removes YAML files from the prettier scope as it was apparently not being used for such files at all.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,6 @@
+examples/**/*.ts
+tests/*/expected/**/*.ts
+tests/*/generated/**/*.ts
+*.md
+*.yaml
+*.yml
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,4 @@
 module.exports = {
-  preset: 'ts-jest',
-  testEnvironment: 'node',
+  preset: "ts-jest",
+  testEnvironment: "node",
 };
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
   "scripts": {
     "build": "pika-pack build",
     "deploy": "npm run build && pika publish",
+    "format": "yarn prettier -w .",
     "lint": "eslint --ignore-path .gitignore --ext .js,.ts src",
     "pregenerate": "npm run build",
     "test": "npm run build && jest",


drwpow:
Thanks! I think for the `expected/*` files, those were actually correct before because they use **the default Prettier setup**. However, the internal code for this project uses a customized Prettier setup, which you did correctly identify was not being applied to `jest.config.js`.

I’d be OK with adding `*.yaml` files being ignored by Prettier, however, notice that the tests are failing—that’s because your changes to the generated files are now incorrectly formatted.

**I’m in favor of this change if you revert the `tests/**.ts` and `examples/*.ts` changes.** If you don’t believe me, run `yarn build && ./pkg/bin/cli.js [path to YAML] -o [output]` locally and you’ll see that they will be generated exactly as they were before, with the default Prettier config.

jankuca:
Since there was no `.prettierignore`, I think it is reasonable to expect running `yarn prettier -w` would lead to the correct result; VSCode even formatted these files automatically on edit/save for me.

I think that all files which should not follow the project-specific prettier config should be listed in the `.prettierignore`. (I will add it, didn't really know which files to list but the YAML files seemed wrong immediately.)

It's weird though that the tests were passing locally but I guess a rebase which removed code for #381 messed it up.

jankuca:
Ok, fixed. I've also added a `yarn format` script which runs prettier.

jankuca:
Now I'm thinking whether the `.prettierignore` affects the prettier runs which should use the default config. The tests are passing but can you check if this is alright?

.prettierignore
```diff
@@ -0,0 +1,6 @@
+examples/**/*.ts
+tests/*/expected/**/*.ts
+tests/*/generated/**/*.ts
+*.md
+*.yaml
+*.yml
```
drwpow, line 6: Cool. Thanks for adding these 👍🏻

drwpow:
> Now I'm thinking whether the `.prettierignore` affects the prettier runs which should use the default config. The tests are passing but can you check if this is alright?

This is great! Thanks for doing this 🙏🏻

> Now I'm thinking whether the `.prettierignore` affects the prettier runs which should use the default config.

To my knowledge, no, they shouldn’t be. We’re using Prettier’s Node API, so that should be theoretically not be paying attention to any config files in the filesystem, unless someone explicitly passes in `--prettier-config`. But if you find this is not the case please let me know!

drwpow (approved):

PR #383, opened 2020-11-24
handle case where additionalRef is a ref (fixes #345)

### LINKED ISSUE #345: Cannot find name 'ref'
> Same as https://github.com/drwpow/openapi-typescript/issues/265

I have an OpenAPI json like that:

```
{
  "swagger": "2.0",
  "definitions": {
    "Messages": {
      "type": "object",
      "additionalProperties": {
        "$ref": "#/definitions/Message"
      }
    },
    "Message": {
      "type": "object",
      "properties": {
        "code": {
          "type": "integer"
        },
        "text": {
          "type": "string"
        }
      }
    }
  }
}
```

Which results in typescript:

```
export interface definitions {
  Messages: { [key: string]: ref }
  Message: { code?: number; text?: string }
}
```

It is not valid, because of ref. What is ref, and why it is not defined?

```
 openapi-typescript --version
✨ openapi-typescript 2.2
```

atlefren:
Fixes #345 

As stated in the comment, the functionality in getAdditionalPropertiesType could perhaps be added in nodeType, but then I had to pass in the rawDefinition there, and risk messing up things.

This seems to work, added a new tests based on the inut in #345, both this and all old tests pass

### DIFF
--- a/src/types/OpenAPI2.ts
+++ b/src/types/OpenAPI2.ts
@@ -32,6 +32,12 @@ export type OpenAPI2Type =
 
 export type OpenAPI2Reference = { $ref: string };
 
+export function isOpenAPI2Reference(
+  additionalProperties: OpenAPI2SchemaObject | OpenAPI2Reference | boolean
+): additionalProperties is OpenAPI2Reference {
+  return (additionalProperties as OpenAPI2Reference).$ref !== undefined;
+}
+
 export interface OpenAPI2SchemaObject {
   additionalProperties?: OpenAPI2SchemaObject | OpenAPI2Reference | boolean;
   allOf?: OpenAPI2SchemaObject[];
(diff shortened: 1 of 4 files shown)


atlefren:
Not completely sure how codecov calculates coverage on my patch and what I can do to improve it, as I cannot find an easy way to run codecov patch locally?

drwpow:
> Not completely sure how codecov calculates coverage on my patch and what I can do to improve it, as I cannot find an easy way to run codecov patch locally?

Yeah I personally have never benefitted much from paying attention to the patch coverage; maybe it’s a project-level setting that I need to disable. Sorry about that! But I would just ignore that.

drwpow (approved):
This looks great, thank you!

atlefren:
Great, any plans for a release soon? This was kind of a deal-breaker for us, we're using a local copy based on this PR for now, but would like to get back to the original library as fast as possible.

drwpow:
> Great, any plans for a release soon? 

Yes actually `3.0.0-rc.0` is released due to #381. It’s probably not a huge breading change for people, but wanted to play it safe. Would love to know if that works well for you and everyone else, and if so we can cut a `3.0.0` release

PR #388, opened 2020-11-26
Update remote HTTP test

drwpow:
Change out the remote schema for the HTTP test. The one we used previously may not be around forever; this uses Swagger’s public example.

While Swagger’s Petstore schema is too simple to even be worth converting, here it’s OK because we’re really focused on the fetching; the conversion is rather moot and is extensively tested elsewhere.


tests/v2/index.test.ts
```diff
@@ -16,22 +16,22 @@ function format(types: string): string {
...
+      const generated = fs.readFileSync(path.join(__dirname, "generated", `${file}.ts`), "utf8");
+      expect(generated).toBe(expected);
     });
   });
 
   it("reads swagger.json spec (v2) from remote resource", () => {
-    execSync("../../pkg/bin/cli.js https://api.catalog.stage.manifold.co/swagger.json -o generated/http.ts", {
-      cwd: path.resolve(__dirname),
+    const expected = fs.readFileSync(path.join(__dirname, "expected/http.ts"), "utf8");
+    execSync("../../pkg/bin/cli.js https://petstore.swagger.io/v2/swagger.json -o generated/http.ts", {
```
gr2m, line 30: I try not to rely on 3rd party URLs for testing, it won't work when I'm offline and it will fail randomly in case their website is down. It's outside of the scope of this PR, but maybe we could serve the swagger.json file with a local server and then download it from localhost instead?

drwpow: Ha as soon as I put this PR up I just realized: _Duh! Test GitHub with GitHub!_ Now we’re using `raw.githubusercontent.com` URLs (they can be slow, but should at least be reliable?)

PR #428, opened 2020-12-31
Exit on error code 1

### LINKED ISSUE #423: Set exit codes for errors
Hello 👋, 
first of all, thanks for maintaining such a great package!

I run openapi-typescript in CI and noticed that even unsuccessful schema conversion returns an exit code of 0.

For example:
```bash
$ npx openapi-typescript swagger.yaml --output schema.ts
✨ openapi-typescript 2.4.2
🤞 Loading spec from swagger.yaml…
❌ "Cannot find spec under the following path: /.../swagger.yaml"
$ echo $?
0
```

I've built a workaround to this but maybe it'd be a good idea to return different exit statuses for errors.

drwpow:
Screen Shot 2020-12-30 at 18 14 35

Errors now exit with status code `1` instead of `0`. Fixes #423

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -56,6 +56,7 @@ async function main() {
   try {
     spec = await loadSpec(pathToSpec, { log: output !== "STDOUT" });
   } catch (err) {
+    process.exitCode = 1; // needed for async functions
     throw new Error(red(`❌ ${err}`));
   }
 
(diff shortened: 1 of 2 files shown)


gr2m (approved):
This is technically a breaking change, so maybe let's release it together with #429?

It'd be neat to add a test for it. But not worth to put too much time into it

PR #429, opened 2020-12-31
Add support for components.responses

### LINKED ISSUE #408: components -> responses types not generated correctly
Hi! 

First off, great work with the tool - we're finding it really useful!

We've found that the responses we define under `components.responses` in our spec don't get generated correctly, and they're not properly referenced in the generated route types, either.

Example Spec:
```yaml
openapi: 3.0.1
info:
  title: Test
  description: Test
  version: 1.0.0
paths:
  /test:
    get:
      tags:
        - Test
      summary: Test
      responses:
        400:
          $ref: "#/components/responses/400BadRequest"

components:
  schemas:
    BadRequestError:
      type: object
      properties:
        message:
          type: string
        code:
          type: string
          enum: [capacity_check_failed, limit_check_failed]
        activityId:
          type: string
      required: [message]
  responses:
    400BadRequest:
      description: There were issues with the request
      content:
        application/json:
          schema:
            type: object
            properties:
              errors:
                type: array
                items:
                  $ref: "#/components/schemas/BadRequestError"
            required:
              - errors
```

Here's the output:
```typescript
/**
 */

export interface paths {
  '/test': {
    get: {
      responses: {
        '400': unknown
      }
    }
  }
}

export interface operations {}

export interface components {
  schemas: {
    BadRequestError: {
      message: string
      code?: 'capacity_check_failed' | 'limit_check_failed'
      activityId?: string
    }
  }
  responses: {
    /**
     * There were issues with the request
     */
    '400BadRequest': { [key: string]: any }
  }
}
```

As you can see, `400BadRequest` is just an object, and the path that uses it has `unknown` rather than referencing the type in the `components` type.

Thanks!

drwpow:
Fixes #408

Before we were skipping detailed responses with `unknown` or `{ [key: string]: any }`. This adds proper support for responses, as requested in #408. Tests have been updated to reflect the new generation.

### DIFF
--- a/tests/v3/expected/github.ts
+++ b/tests/v3/expected/github.ts
@@ -1672,8 +1672,8 @@ export interface operations {
             pem: string;
           } & { [key: string]: any });
       };
-      404: unknown;
-      422: unknown;
+      404: components["responses"]["not_found"];
+      422: components["responses"]["validation_failed_simple"];
     };
   };
   /**
@@ -1755,8 +1755,8 @@ export interface operations {
       200: {
         "application/json": components["schemas"]["installation"];
       };
-      404: unknown;
-      415: unknown;
+      404: components["responses"]["not_found"];
+      415: components["responses"]["preview_header_missing"];
     };
   };
   /**
@@ -1775,7 +1775,7 @@ export interface operations {
        * Empty response
        */
       204: never;
-      404: unknown;
+      404: components["responses"]["not_found"];
     };
   };
   /**
@@ -1815,11 +1815,11 @@ export interface operations {
       201: {
         "application/json": components["schemas"]["installation-token"];
       };
-      401: unknown;
-      403: unknown;
-      404: unknown;
-      415: unknown;
-      422: unknown;
+      401: components["responses"]["requires_authentication"];
+      403: components["responses"]["forbidden"];
+      404: components["responses"]["not_found"];
+      415: components["responses"]["preview_header_missing"];
+      422: components["responses"]["validation_failed"];
     };
   };
   /**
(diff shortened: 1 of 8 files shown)


tests/v3/specs/petstore.yaml
```diff
@@ -1,10 +1,9 @@
 openapi: 3.0.1
 info:
   title: Swagger Petstore
-  description:
-    "This is a sample server Petstore server.  You can find out more about     Swagger
+  description: 'This is a sample server Petstore server.  You can find out more about     Swagger
```
drwpow, line 4: The Swagger Petstore schema was updated because it was technically invalid before. We had modified it to try and test for `schema.responses`, but it was modified into an invalid format. The original Petstore schema was restored, and we now test for `schema.responses` in a better way in our unit tests.

tests/v3/expected/github.ts
```diff
@@ -1672,8 +1672,8 @@ export interface operations {
             pem: string;
           } & { [key: string]: any });
       };
-      404: unknown;
-      422: unknown;
+      404: components["responses"]["not_found"];
+      422: components["responses"]["validation_failed_simple"];
```
drwpow, line 1676: @gr2m how does this sit with you? Is this better?

gr2m: looks great!

gr2m: looks great!

gr2m (approved):

gr2m (approved):
Changes look great! I run it against https://github.com/octokit/openapi-types.ts and it improved the types quite a bit. I'd still ship it as a breaking change, because the status code keys were strings before, and are now numbers (which they should be)

Issue #436, opened 2021-01-06
openapi 2.0: Should generate `parameters`


tstelzer:
Hey there, just found this project and greatly appreciate it! Specifically the philosophy of generating types closely resembling the original API.

One thing I'm missing is generated `parameters`. Given the following spec:

```yaml
swagger: "2.0"
info:
  description: This is some API.
  version: 1.0.0
  title: My API
basePath: /api/v1
consumes:
  - application/json
  - multipart/form-data
produces:
  - application/json
schemes:
  - http
  - https
paths:
  /thing/{id}:
    get:
      summary: Get a Thing
      description: Get a Thing
      parameters:
        - $ref: '#/parameters/ThingParams'
      responses:
        200:
          description: The thing
          schema:
            $ref: "#/definitions/Thing"

parameters:
  ThingParams:
    in: path
    name: id
    required: true
    type: string
    description: The id of the thing.

definitions:
  Thing:
    type: object
    description: A thing
    properties:
      id:
        type: string
        description: Uniquely identifies this thing.
```

I would expect this output:

```ts
/**
 * This file was auto-generated by openapi-typescript.
 * Do not make direct changes to the file.
 */

export interface definitions {
  /**
   * A thing
   */
  Thing: {
    /**
     * Uniquely identifies this thing.
     */
    id?: string;
  };
}

export interface parameters {
    ThingParams: {
        /**
         * The id of the thing.
         */
        id: string;
    };
}
```

But the actual result is missing `parameters`. Would love this addition. Cheers.

/edit:

I guess one could think about grouping `parameters` on `in`, e.g. 

```ts
export interface parameters {
  path: {
    ThingParams: {/* ... */};
  };
}
```

drwpow:
This was added in v3 of this library! 🎉 Closing, but if you encounter any issues with the new generation, please feel free to open a more specific issue.

PR #448, opened 2021-01-13
Refactor, add parameters for V2

drwpow:
This is a half-rewrite of the library. Reason is that originally when I released 2.x of this library, I figured that OpenAPI V2 and V3 are different enough I should keep the code separate. While that was a good gut feeling, in practice they can be transformed very similarly, with few differences.

This combines the OpenAPI V2 & V3 code together so we get more V3 features ported into V2. For the most part we can use the same code for both, an overall reduction! Only a few places where V2 & V3 diverge.

The goal is to have full feature-set parity from before, but with the addition of parameter generation for V2 (#436). I also figured since we were releasing 3.x anyway, what better time to do it than now?

**TODO**
- [x] get V2 & V3 tests passing again
- [x] add additional parameters tests (esp. for V2)
- [ ] more testing

tagging @gr2m if you’re curious. No pressure at all, but if you have a chance to test out `npx openapi-typescript@3.0.0-alpha.1` on your GitHub schemas and let me know if there are any regressions, that’d be helpful! I’ve been going through the diffs, but sometimes it‘s tough to see the true diffs among the noise


gr2m:
It's [a huge diff](https://gist.github.com/gr2m/dbc593df378a2428702fd6199adf2ca3). But as far as I can tell no regressions.

The changes I see

- comments added to the paths object
- single line comments
- change of order
- `components["responses"]` is actually used (I didn't caught this one before, that is great :)
- operation responses values now include a `headers` key (that might actually be a regression, I'm not sure if that is correct)
  
  Example before

  ```ts
  "activity/list-repos-watched-by-user": {
    parameters: {
      path: {
        username: components["parameters"]["username"];
      };
      query: {
        per_page?: components["parameters"]["per_page"];
        page?: components["parameters"]["page"];
      };
    };
    responses: {
      "200": {
        "application/json": components["schemas"]["minimal-repository"][];
      };
    };
  };
  ``

  after

  ```ts
    };
  "activity/list-repos-watched-by-user": {
    parameters: {
      path: {
        username: components["parameters"]["username"];
      };
      query: {
        per_page?: components["parameters"]["per_page"];
        page?: components["parameters"]["page"];
      };
    };
    responses: {
      200: {
        headers: {};
        "application/json": components["schemas"]["minimal-repository"][];
      };
    };
  };
  ```

  See the `headers: {};` 

There might be other changes I didn't catch yet, and I didn't yet have the time to test the newly generated types with actual code, I ran out of time for now. I can have another look once you are done from your side, ok?

drwpow:
> It's a huge diff. But as far as I can tell no regressions.

Thanks so much for taking a look! 🙏🏻 I’m also trying it out on projects to see if there are any code changes

> comments added to the paths object

Right this was something that was needed, I think

> single line comments

Merged in #443 but still unreleased

> change of order

I went off of the order in the OpenAPI 3 spec. While normally I’d defer to the order in which a schema was written (we still do in many places, like components), there were areas where we _had_ to reorder things anyway. So when we do reorder, we reorder in schema order.

> `components["responses"]` is actually used (I didn't caught this one before, that is great :)

🙂 

> operation responses values now include a `headers` key (that might actually be a regression, I'm not sure if that is correct)

This was actually added intentionally, because it was missing before. You can let me know if this is useful or not. But to your point we can hide it if it’s an empty object.

In a related point, I noticed that in some places we’d generate `200: { 'application/json': …` and in other places `200: { content: { 'application/json' …`. In our tests you can see both ways. I decided to just always have `content` as a namespace because that matches the schema, it‘s more expected, and it allows for better namespacing with `headers`. I know it adds a few characters in length, though, so open to thoughts there (I think we should either always include it, or always remove it, not do both as we were).

gr2m:
> In a related point, I noticed that in some places we’d generate `200: { 'application/json': …` and in other places `200: { content: { 'application/json' …`. In our tests you can see both ways. I decided to just always have `content` as a namespace because that matches the schema, it‘s more expected, and it allows for better namespacing with `headers`. I know it adds a few characters in length, though, so open to thoughts there (I think we should either always include it, or always remove it, not do both as we were).

I like `content: { 'application/json' …`, as long as we are consistent. Do you need help with looking into the inconsistency? Do you know if there are any other keys besides `content`, could there be `headers`, too?

drwpow:
> I like content: `{ 'application/json' …`, as long as we are consistent. Do you need help with looking into the inconsistency?

👍🏻  OK that’s what we’re doing in this refactor. I should have clarified: that was behavior before that has been fixed in this PR. I think it was because this PR does a better job at abstracting transform functions; before we were just hitting different codepaths because it was starting to turn a little spaghetti-ish

drwpow:
Update: I’ve generated some schemas from the current beta from this PR, and I’m feeling like this is an improvement over where `2.x` is, and it even fixes a few bugs in addition to shipping features (also I kept in changes we were going to ship in `3.x` anyway, such as numeric status codes). Overall, starting to feel more confident in this as a release candidate

gr2m:
Can we maybe ship the change of the comments in `2.x`? That's not a breaking change, and it will greatly reduce the diffs when upgrading to `3.x`. Will make it easier to spot regressions. If it's too much work don't bother, only if it's easy to do

drwpow:
> Can we maybe ship the change of the comments in `2.x`?

Can do! I will ship that tonight, no prob.

drwpow:
`v2.5.0` just released.

gr2m:
> `v2.5.0` just released.

Thanks a lot! The version also sets the `response` types that are referenced directly at root. The problem is now that some of the response types have the `content` namespace while others do not. 

For example

```
          "404": {
            "$ref": "#/components/responses/not_found"
          },
```

gets transpiled to

```
404: components["responses"]["not_found"];
```

and `not_found` is

```
    not_found: {
      content: {
        "application/json": components["schemas"]["basic-error"];
      };
    };
```

while

```
          "200": {
            "description": "Default response",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/webhook-config"
                },
                "examples": {
                  "default": {
                    "$ref": "#/components/examples/webhook-config"
                  }
                }
              }
            }
          }
        },
```

gets transpiled to

```
      200: {
        "application/json": components["schemas"]["webhook-config"];
      };
```

and `webhook-config` does not have the `content` namespace because it's referenced below the `content` key in the OpenAPI spec

Do you think we could fix that in `2.x`? Can I help?

drwpow:
> Do you think we could fix that in `2.x`? Can I help?

Interesting. Is that a regression with the code I just pushed? Or just an unresolved bug?

The code for this PR changes things a bit. I guess I’d wonder whether that `content` namespace is “breaking“ behavior, or if we’re fixing a bug there. I’m open to either, but my preference would be to probably try and release these in `3.x`.

gr2m:
> > Do you think we could fix that in `2.x`? Can I help?
> 
> Interesting. Is that a regression with the code I just pushed? Or just an unresolved bug?
> 
> The code for this PR changes things a bit. I guess I’d wonder whether that `content` namespace is “breaking“ behavior, or if we’re fixing a bug there. I’m open to either, but my preference would be to probably try and release these in `3.x`.

At least in case of GitHub's OpenAPI spec, I'd say it's an unresolved bug. The latest 2.x version fixed a bug with `$ref`s for operation responses, but it introduced, or surfaced, a new one. The `components["responses"]` schemas are used now, before they have just been ignored and the type was set to `unknown`. But the schemas `components["responses"]` have the `content` namespace, while other response types that don't use references at the same level or don't use references at all do not have the `content` namespace.

Maybe this is not fixable in 2.x. Introducing the `content` namespace to response types is a breaking change. I'd say don't worry about it too much, we can just get it fixed in 3.x.

drwpow:
> Maybe this is not fixable in 2.x. Introducing the `content` namespace to response types is a breaking change. I'd say don't worry about it too much, we can just get it fixed in 3.x.
 
Yeah as much as I‘d like to fix it in 2.x, I just don’t think I have the time to. Not for OpenAPI V2 and V3. I’m on the same page: I think now that this PR merges the V2 and V3 code, we can fix both in the same places. I think this new code is closer to fixing that bug.

In that regard, I’ll merge this, and create a `v2` branch in case we need it. And I’ll look specifically into your issue with the GitHub schema later tonight/this weekend and see if we can fix it before releasing 3.x. That way when we release it, we can not only provide the much-requested features for OpenAPI V2, but we can ship those without shipping the bug.

tshelburne:
@drwpow FYI, tested this in one of my projects, no regressions.

MeLlamoPablo:
Hi! When can we expect this to release? Could we have at least a beta/rc version published?

I'm getting a `TypeError: Cannot convert undefined or null to object` with the current  `v2.5.0` but this commit fixes it, so I would really appreciate if this was released :smile:

drwpow:
> Hi! When can we expect this to release? Could we have at least a beta/rc version published?

You can try out the `@next` release and it has all the latest changes. Been meaning to release the full version; just trying to complete all my testing and life keeps getting in the way. Full release coming very soon (next couple days)!

MeLlamoPablo:
Thank you so much! Shame on me for not checking that an alpha already existed :sweat_smile:

PR #489, opened 2021-02-26
Fix V2 parameter types

### LINKED ISSUE #470: openapi 2: parameters with known type parsed to `unknown`
**Given** a spec such as:

```yml
swagger: "2.0"
info:
  description: This is some API.
  version: 1.0.0
  title: My API
basePath: /api/v1
consumes:
  - application/json
  - multipart/form-data
produces:
  - application/json
schemes:
  - http
  - https
paths:
  /thing/{id}:
    post:
      summary: Some summary
      description: Some description
      operationId: postThing
      parameters:
        - in: path
          name: id
          type: string
          description: The id of the thing.
          required: true
        - in: header
          name: X-Something
          type: string
          minLength: 1
          description: Some additional header.
          required: true
        - in: query
          name: foo
          type: string
          description: some foo
          required: true
        - in: body
          name: something
          required: true
          description: Some body.
          schema:
            type: string
      responses:
        200:
          description: The thing.

definitions:
  Thing:
    type: object
    description: Some thing.
```

I get this output:
```typescript
/**
 */

export interface paths {
  "/thing/{id}": {
    /** Some description */
    post: operations["postThing"];
  };
}

export interface definitions {
  /** Some thing. */
  Thing: { [key: string]: any };
}

export interface operations {
  /** Some description */
  postThing: {
    parameters: {
      path: {
        /** The id of the thing. */
        id: unknown;
      };
      header: {
        /** Some additional header. */
        "X-Something": unknown;
      };
      query: {
        /** some foo */
        foo: unknown;
      };
      body: {
        /** Some body. */
        something: string;
      };
    };
    responses: {
      /** The thing. */
      200: unknown;
    };
  };
}
```

In `operations.postThing.parameters`, I would **expect** `path.id`, `header['X-Something']` and `query.foo` to reflect the types in the spec (all `string`, in this case).

What we **actually** see is that all of those types are `unknown`.

Version: 3.0.1

yamacent:
close #470


drwpow:
Thanks so much! I’ll review it soon

yamacent:
@drwpow Thanks for you reply! I'll modify tests as well if the direction of this PR is okay.

src/transform/type.ts
```diff
@@ -0,0 +1,45 @@
+import { ItemsObject, ParameterObject } from "types";
+import { tsUnionOf } from "../utils";
+
+function transformEnum(enumArray: string[]): string {
+  return tsUnionOf(enumArray.map((e) => (typeof e === "string" ? `'${e.replace(/'/g, "\\'")}'` : e)));
+}
+
+function transformItemsObj(obj: ItemsObject): string {
+  switch (obj.type) {
```
drwpow, line 9: This code actually already exists in [utils.ts#L29](https://github.com/drwpow/openapi-typescript/blob/main/src/utils.ts#L29). So by re-adding different logic, we could introduce bugs.

drwpow:
Thanks so much for your PR! I think the fix is much easier than expected. We’re using the same code for OpenAPI 3.0 and Swagger 2.0, and the `unknown` is happening because it’s trying to parse everything as 3.0. All we need to change is one line on [parameters.ts#L43](https://github.com/drwpow/openapi-typescript/blob/main/src/transform/parameters.ts#L43).


```diff
- paramObj.schema ? transformSchemaObj(paramObj.schema) : "unknown"
+ if (version === 2) {
+  paramObj.type ? transformSchemaObj(paramObj) : "unknown"
+ } else if (version === 3) {
+   paramObj.schema ? transformSchemaObj(paramObj.schema) : "unknown"
+ } 
```

So that’s step 1. Step 2 would be to pass that `version` from `transform/index.ts` into `transform/paths.ts` and finally into `transform/parameters.ts`. That should fix it!

Also, I find it best to start by actually making a new test in `parameters.test.ts` first, and then working until it passes.

yamacent:
@drwpow 
Thank you for your review and kind explanation!
I added tests to `parameters.test.ts` and fixed the code to pass the test as you say.
Please review again.
Also, could you please tell me how to fix tests in `tests/v2` directory?

drwpow:
> Also, could you please tell me how to fix tests in `tests/v2` directory?

Ah so you know what? This problem _is_ a bit trickier than expected. I seemed to miss this critical part of the [Open API 2.0 specs](https://swagger.io/specification/v2/#parameterObject) that states that we only use the top-level object **if `in` is NOT `body`**. Otherwise, we do use `schema` if `in` is `body`.

Will comment a fix on your PR and I think we can wrap this up!

src/transform/parameters.ts
```diff
@@ -16,9 +17,10 @@ export function transformParametersArray(
       if (globalParams[paramName]) {
         const reference = globalParams[paramName] as any;
         if (!mappedParams[reference.in]) mappedParams[reference.in] = {};
+        const schema = { $ref: paramObj.$ref };
         mappedParams[reference.in][reference.name || paramName] = {
           ...reference,
-          schema: { $ref: paramObj.$ref },
+          ...(version === 2 ? schema : version === 3 ? { schema } : {}),
```
drwpow, line 23: So because this is a bit difficult to follow, let’s take up a little more space:

```ts
if (version === 2) {
    ..reference,
    $ref: paramObj.$ref,
  } as any;
} else if (version === 3) {
    ..reference,
    schema: { $ref: paramObj.$ref },
  } as any;
}
```

Your version takes up fewer lines, and is good in that regard. But you have a nested ternary + a spread operator, and it’s very easy for someone to make a mistake (if not you, the next person to touch this).

src/transform/parameters.ts
```diff
@@ -39,9 +41,13 @@ export function transformParametersArray(
       if (paramComment) output += comment(paramComment);
 
       const required = paramObj.required ? `` : `?`;
-      output += `    "${paramName}"${required}: ${
-        paramObj.schema ? transformSchemaObj(paramObj.schema) : "unknown"
-      };\n`;
+      let type = ``;
+      if (version === 2) {
+        type = paramObj.type ? transformSchemaObj(paramObj) : "unknown";
```
drwpow, line 46: So according to the [OpenAPI 2.0 spec](https://swagger.io/specification/v2/#parameterObject), this will cause a regression, which is why the V2 tests are failing (to your credit, I think the specs are **very** confusing here, and I suspect it’s why they changed it for 3.0 🙂 ).

Instead, this needs to be:

```suggestion
  if (paramObj.in === 'body' && paramObj.schema) {
    type = transformSchemaObj(paramObj.schema);
  } else if (paramObj.type) {
    type = transformSchemaObj(paramObj);
  } else {
    type = 'unknown';
  }
```

src/transform/parameters.ts
```diff
@@ -39,9 +41,13 @@ export function transformParametersArray(
       if (paramComment) output += comment(paramComment);
 
       const required = paramObj.required ? `` : `?`;
-      output += `    "${paramName}"${required}: ${
-        paramObj.schema ? transformSchemaObj(paramObj.schema) : "unknown"
-      };\n`;
+      let type = ``;
```
drwpow, line 44: nit: can we use another name, other than `type`? Only because `type` is a special word in TypeScript (it doesn’t seem to be causing any problems here, but just for clarity)

src/transform/parameters.ts
```diff
@@ -39,9 +41,13 @@ export function transformParametersArray(
...
+      if (version === 2) {
+        type = paramObj.type ? transformSchemaObj(paramObj) : "unknown";
+      } else if (version === 3) {
+        type = paramObj.schema ? transformSchemaObj(paramObj.schema) : "unknown";
```
drwpow, line 48: 👍🏻 I believe this is still correct.

tests/parameters.test.ts
```diff
@@ -1,59 +1,112 @@
...
-        {
-          description: "Specifies which fields in the response should be expanded.",
-          in: "query",
-          name: "expand",
-          required: false,
-          schema: {
-            items: {
-              type: "string",
+describe.only("transformParametersArray()", () => {
+  describe("v2", () => {
```
drwpow, line 4: Nice tests! 💯

tests/parameters.test.ts
```diff
@@ -1,59 +1,112 @@
...
-          name: "three_d_secure",
-          required: true,
-          schema: {
-            type: "string",
-          },
-        },
-      ]).trim()
-    ).toBe(
-      `query: {
+            { in: "path", name: "three_d_secure", required: true, type: "string" },
```
drwpow, line 19: Suggestion: with our new knowledge, let’s try adding a `body` param here, like: `{ in: "body", schema: { type: "string" } }`. Then we can test if it generated correctly.

src/transform/parameters.ts
yamacent: I agree with you. Fixed in https://github.com/drwpow/openapi-typescript/pull/489/commits/8c02bd78cbc0ac13d83632ff358991056c54bfa2

Perhaps we need to consider the case where `in` is `body` here as well??

src/transform/parameters.ts
yamacent: Thank you! I missed the spec.. fixed in https://github.com/drwpow/openapi-typescript/pull/489/commits/fe3b69b4a0682f9799ffef29ce57dcc498cc47f4

src/transform/parameters.ts
yamacent: I see! I renamed it to `paramType`. https://github.com/drwpow/openapi-typescript/pull/489/commits/bb668eddc144b54ad422a2245eb54af4e2f63bf0

yamacent:
@drwpow  By the way, do we need to keep the changes I did in `src/types/index.ts`?

src/types/index.ts
```diff
@@ -60,13 +60,22 @@ export interface OperationObject {
...
+
 export interface ParameterObject {
   name?: string; // required
   in?: "query" | "header" | "path" | /* V3 */ "cookie" | /* V2 */ "formData" | /* V2 */ "body"; // required
   description?: string;
   required?: boolean;
   deprecated?: boolean;
   schema?: ReferenceObject | SchemaObject; // required
+  type?: "string" | "number" | "integer" | "boolean" | "array" | "file"; // V2 ONLY
+  items?: ItemsObject; // V2 ONLY
```
drwpow, line 77: This should be `ReferenceObject | SchemaObject` I believe.

src/types/index.ts
```diff
@@ -60,13 +60,22 @@ export interface OperationObject {
   responses?: Record<string, ReferenceObject | ResponseObject>; // required
 }
 
+export interface ItemsObject {
```
drwpow, line 63: I believe we can remove this; this seems to be a duplication of `SchemaObject` but with fewer fields (see `SchemaObject["items"]` as a reference

drwpow:
This is looking great! It looks like the only thing keeping the tests from passing are to take `tests/v2/generated/*.ts` and copy them over to `tests/v2/expected/*.ts`. Basically these are “snapshot” tests that test very, very large schemas.

These snapshot tests are bad at testing individual features because there are so many lines. But they are good at making sure that complete schemas are generating correctly. I believe all the changes I see point to your PR working as intended; it looks like we are now at an improvement with no other regressions!

I’ll go ahead and approve this, and as soon as you copy the files to `tests/v2/expected/*.ts` the tests should be working. Then we can merge & release this!

yamacent:
I fixed types where you pointed out ( https://github.com/drwpow/openapi-typescript/pull/489/commits/a5b9c791c7da64ae871ec6a20ce7960a1069f1fe ) and updated `tests/v2/expected/*.ts` as you said ( https://github.com/drwpow/openapi-typescript/pull/489/commits/2e9f00428ff375f4569b6f1c91af135d628907e6 ).
Please confirm.

drwpow:
This looks great, thank you!

drwpow:
@all-contributors please add @yamacent for code, bug, test

yamacent:
@drwpow Once again, thank you for your kind review!

drwpow:
This was just released in `3.1.0`! 🚀 Thank you again for your contribution.

PR #509, opened 2021-03-07
Remove @pika/pack 😢

drwpow:
Removes @pika/pack as the primary npm package builder & publish tool. Though it was a great project, it’s sadly on hiatus. And I, being partially to blame for that, don’t have the time to prioritize it & give it the updates it needs. For now, this just uses a manual build process using TypeScript.

This has been published to `openapi-typescript@next` on npm for testing, because changing the npm build could potentially break some things for people.


drwpow:
Aaand merging because this seems to do just fine (but will test once again as a release candidate before the next `latest`).

Issue #501, opened 2021-03-04
Enhancement: readonly arrays


laurence-myers:
It would be nice if we could use `ReadonlyArray` in place of `[]`. Currently, trying to return a `ReadonlyArray` in a response will fail type checking, because the response requires a mutable array.

e.g.

```
    ErrorResponse: {
      //errors: components["schemas"]["Error"][]; // old way
      errors: ReadonlyArray<components["schemas"]["Error"]>; // new way
    };
```

drwpow:
Interesting! This makes sense to me.

Just for clarification: should _all_ arrays be `Readonly` in TS generation? My gut instinct would be yes, but I’d like to confirm that’s what you’d think.

laurence-myers:
Yes, that'd be my preference. 🙂 But it would depend on the codebase, and
whether it generally uses immutable data. Some apps might want to add items
to an array while building a response, or "pop" items from an input array
(e.g. if it's an intermediate proxy), particularly if performance is a
concern.

Maybe it should be configurable, via CLI option etc.

drwpow:
That’s a fair point. My initial assumption was to treat it as immutable, but now that you say that, I myself have used this library to **generate new structures** before sending to an API (e.g. a `ReadonlyArray<>` would be a pretty bad structure if you were POST-ing something).

I like the idea of a CLI flag that just turns all arrays into `ReadonlyArray<>`. Looking at the OpenAPI spec, I see [readOnly](https://swagger.io/specification/#schemaReadOnly) as a schema option, but reading the description of “it MAY be sent as part of a response but SHOULD NOT be sent as part of the request” makes me think that this is something different than your idea. I think your idea is more of a TypeScript concern than a property of your OpenAPI schema, which makes me think a flag may be best.

dnalborczyk:
I second this. it would be great if we could also make all types immutable, not only arrays.

e.g. [graphql-code-generator](https://github.com/dotansimha/graphql-code-generator) for typescript has an option to generate immutable types: https://graphql-code-generator.com/docs/plugins/typescript

> immutableTypes
> type: boolean default: false
> 
> Generates immutable types by adding readonly to properties and uses ReadonlyArray


```ts
type Foo {
	readonly bar: readonly foobar[]
}
```

PR #522, opened 2021-03-15
feat: generate immutable types

dnalborczyk:
fixes: https://github.com/drwpow/openapi-typescript/issues/501


bin/cli.js
```diff
@@ -15,6 +15,7 @@ Options
   --help                display this
   --output, -o          Specify output file (default: stdout)
   --auth                (optional) Provide an authentication token for private URL
+  --immutableTypes      (optional) Generates immutable types (readonly properties and readonly array)
```
drwpow, line 18: nit: `meow` allows camelCase or kebab-case flags; WDYT about just using kebab-case here? Both should be supported behind-the-scenes

src/transform/headers.ts
```diff
@@ -1,15 +1,23 @@
...
+  { immutableTypes }: { immutableTypes: boolean }
+): string {
   let output = "";
 
   Object.entries(headerMap).forEach(([k, v]) => {
     if (!v.schema) return;
 
     if (v.description) output += comment(v.description);
+
+    const readonly = immutableTypes ? "readonly " : "";
```
drwpow, line 15: 👍🏻

drwpow:
Awesome! Love the direction of this PR; I would have done it the same way.

My only suggestion would be to add at least 1 test for this. But the nice part of adding a new feature is that we’re not breaking behavior for anybody. So even if we missed a type, we could add it after-the-fact.

src/transform/headers.ts
drwpow: I personally am also a fan of `readonly string[]` over `ReadonlyArray<string>`. AFAIK this looks easier to implement, and has no downsides. I’m in favor of this approach.

bin/cli.js
dnalborczyk: oh, yeah, definitely. I didn't even notice `prettier-config` and `raw-schema` being `kebab-case`.

dnalborczyk:
> Awesome! Love the direction of this PR; I would have done it the same way.

Thank you! 😊 

> My only suggestion would be to add at least 1 test for this.

yeah, I'll add some tests. just wanted to see the current tests passing.

> But the nice part of adding a new feature is that we’re not breaking behavior for anybody. So even if we missed a type, we could add it after-the-fact.

I agree.

src/transform/headers.ts
dnalborczyk: the "downside" - if you want to call it such - would be a minimum requirement of TypeScript v3.4, which I think should be fine considering v4.2 is currently latest.

drwpow: TS 4 was such an easy update I don’t know of any reason why people would still be on an old version. But given that this is opt-in too, this is more than fine.

dnalborczyk:
@drwpow the PR is pretty much good to go. I just want to merge some duplicate interfaces I created for passing around options: { immutableTypes }. gonna try to get to it tomorrow.

drwpow:
Awesome. I’ll take a look tomorrow. It’s looking good to me already.

src/transform/index.ts
```diff
@@ -7,11 +7,14 @@ import { transformResponsesObj, transformRequestBodies } from "./responses";
...
 
 interface TransformOptions {
+  immutableTypes: boolean;
   rawSchema?: boolean;
   version: number;
 }
 
-export function transformAll(schema: any, { version, rawSchema }: TransformOptions): string {
+export function transformAll(schema: any, { immutableTypes, rawSchema, version }: TransformOptions): string {
+  const readonly = immutableTypes ? "readonly " : "";
```
drwpow, line 16: Not a blocker for this PR, but I’d suggest moving this into `src/utils.ts` as `tsReadonly()` util. Similar to the existing `tsUnionOf()` util and others.

It’s a minor thing, but it makes modifications quicker, and helps prevent typos.

tests/operation.test.ts
```diff
@@ -72,5 +99,22 @@ describe("requestBodies", () => {
...
+          /** Pet request body */
+          Pet: {
+            readonly content: {
+              readonly "application/json": {
+                readonly test?: string;
+              };
+            };
+          };
+        };`)
+    );
```
drwpow, line 118: 💯

tests/v3/index.test.ts
```diff
@@ -1,18 +1,33 @@
...
         cwd: __dirname,
       });
-      const expected = fs.readFileSync(path.join(__dirname, "expected", output), "utf8");
-      const generated = fs.readFileSync(path.join(__dirname, "generated", output), "utf8");
+      const expected = readFileSync(join(__dirname, "expected", output), "utf8");
+      const generated = readFileSync(join(__dirname, "generated", output), "utf8");
+      expect(generated).toBe(expected);
+    });
+
+    it(`reads ${schema} spec (v3) from file (immutable types)`, () => {
```
drwpow, line 20: Love all the immutable tests!

drwpow (approved):
This PR is looking good! I have no changes requested. Ping me when you feel you’re happy with it and I can merge.

I can also either release an RC if you’d like to do more testing, or we can just cut a minor release. Your call. Again, since this is a new flag, I’m not as concerned with just releasing because we won’t be breaking anyone.

src/transform/index.ts
dnalborczyk: no problem, I can do that quick as well. I had it set up originally as a function, but changed it back.

dnalborczyk:
> This PR is looking good! I have no changes requested. Ping me when you feel you’re happy with it and I can merge.
> 
> I can also either release an RC if you’d like to do more testing, or we can just cut a minor release. Your call. Again, since this is a new flag, I’m not as concerned with just releasing because we won’t be breaking anyone.

should be good to go! 👍 **minor** release should be fine 🤞

dnalborczyk:
@drwpow just want to point out a couple things quick:

drwpow:
Thanks so much! Agree with all of these changes. Thanks again for contributing

drwpow:
@all-contributors please add @dnalborczyk for code, test

dnalborczyk:
> Thanks so much! Agree with all of these changes. Thanks again for contributing

@drwpow  thank you for this great project and thank you for accepting the PR!

laurence-myers:
@drwpow @dnalborczyk Thanks for implementing & reviewing this feature! I really appreciate your time & efforts. 🎉 

(I just wanted to join the "thanks" party! 😁)

PR #530, opened 2021-03-18
Inherit parameters from path item to its operations.

henhal:
A path item may declare common parameters to all operations outside of the operation map.
If this is the case, all operations should inherit those parameters and merge them with
any operation specific parameters.
In case of conflicting parameters, the one defined in the operation has precedence over
path item parameters.

To fix this, operations need a reference to their path item, so modifying the record of operations built during path item parsing to hold { operation, pathItem } objects.

### DIFF
--- a/src/transform/operation.ts
+++ b/src/transform/operation.ts
@@ -1,4 +1,4 @@
-import { OperationObject, ParameterObject, RequestBody } from "../types";
+import { OperationObject, ParameterObject, PathItemObject, RequestBody } from "../types";
 import { comment, isRef, transformRef, tsReadonly } from "../utils";
 import { transformParametersArray } from "./parameters";
 import { transformResponsesObj } from "./responses";
@@ -9,8 +9,10 @@ export function transformOperationObj(
   {
     globalParameters,
     immutableTypes,
+    pathItem,
     version,
   }: {
+    pathItem?: PathItemObject;
     globalParameters?: Record<string, ParameterObject>;
     immutableTypes: boolean;
     version: number;
@@ -20,8 +22,9 @@ export function transformOperationObj(
 
   let output = "";
 
-  if (operation.parameters) {
-    output += `  ${readonly}parameters: {\n    ${transformParametersArray(operation.parameters, {
+  if (operation.parameters || pathItem?.parameters) {
+    const parameters = (pathItem?.parameters || []).concat(operation.parameters || []);
+    output += `  ${readonly}parameters: {\n    ${transformParametersArray(parameters, {
       globalParameters,
       immutableTypes,
       version,
--- a/src/transform/operation.ts
+++ b/src/transform/operation.ts
@@ -9,7 +9,7 @@ export function transformOperationObj(
   {
     globalParameters,
     immutableTypes,
-    pathItem,
+    pathItem = {},
     version,
   }: {
     pathItem?: PathItemObject;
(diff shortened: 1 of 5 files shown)


src/transform/operation.ts
```diff
@@ -8,16 +8,19 @@ export function transformOperationObj(
...
     version: number;
+    pathItem?: PathItemObject;
     globalParameters?: Record<string, ParameterObject>;
   }
 ): string {
   let output = "";
 
-  if (operation.parameters) {
-    output += `  parameters: {\n    ${transformParametersArray(operation.parameters, {
+  if (operation.parameters || pathItem?.parameters) {
```
drwpow, line 21: This line seemed to throw ts-jest. If you could rebase off latest `main` and push again, we should see the tests run.

drwpow:
I think this looks good! But please run the tests locally yourself, and make sure that they’re all passing (and please update any `expected/*.ts` files, as you’ll very likely find the generation changed for those)

tests/operation.test.ts
```diff
@@ -74,3 +74,112 @@ describe("requestBodies", () => {
...
+    ).toBe(`parameters: {
+      path: {
+    "p2"?: number;
+    "p3"?: string;
+    "p1"?: string;
+  }
+
+  }`);
+  });
+});
```
drwpow, line 185: Great tests! 💯

src/transform/operation.ts
dnalborczyk: `optional chaining` is not supported by the current typescript settings.

henhal: Thanks @dnalborczyk. But weird, I obviously ran this locally and `ts-jest` should have used `tsconfig.json` with the correct transpile options. Could this possibly be related to Node 12 being used in CI while I use Node 14?
But no problem changing it obviously.

tests/operation.test.ts
henhal: Thanks! The tests obviously did work when I ran this locally, but not in CI for some reason. Checking if removing Node 14 stuff solves it.

src/transform/operation.ts
dnalborczyk: good question. not quite sure. the test run also failed in the ci with node.js v14: https://github.com/drwpow/openapi-typescript/runs/2141031952 . I'd have to pull your PR with that commit and run it locally as well.

tests/operation.test.ts
```diff
@@ -74,3 +74,112 @@ describe("requestBodies", () => {
     );
   });
 });
+
+describe.only("parameters", () => {
```
henhal: :facepalm: `only`

src/transform/operation.ts
dnalborczyk: there seems to be a bug in the ci setup. v14.x and v12.x are essentially running node v10.x

image

image

dnalborczyk: fixed in: https://github.com/drwpow/openapi-typescript/pull/539

drwpow: Thanks so much for improving this.

drwpow (approved):
Looks great! Thanks for contributing

drwpow:
@henhal I rebased your branch and fixed the conflict, but kept everything in your code the same. Hope that’s OK. Wanted to merge this to release soon! 🚀

drwpow:
@all-contributors please add @henhal for code, test

henhal:
@drwpow Thanks for rebasing and accepting the PR. 🎈

PR #535, opened 2021-03-22
chore: remove unused function

dnalborczyk:


drwpow (approved):
Thanks for the cleanup! Will have to mentally bookmark this in case we use `Object.fromEntries()` again.

dnalborczyk:
> Thanks for the cleanup! Will have to mentally bookmark this in case we use `Object.fromEntries()` again.

once node **v10** is being removed, it's supported in **v12+** https://node.green/#ES2019-features--Object-fromEntries

drwpow:
> once node **v10** is being removed

SOON

PR #539, opened 2021-03-24
ci: use node.js version from matrix

dnalborczyk:
fixes: https://github.com/drwpow/openapi-typescript/pull/530#discussion_r599828545

also included node.js v15

v10 can probably dropped end-of-April: https://nodejs.org/en/about/releases/ and replaced with v16 (while v15 to be removed)


drwpow (approved):
D‘oh! Thanks for fixing this. Also thanks for adding Node 15. I’ve been using it locally, but always good to have it in CI too.

PR #579, opened 2021-05-03
chore: use node.js v16

dnalborczyk:


drwpow (approved):
Thank you! Locally I’ve started using 16; thanks for updating CI too.

PR #585, opened 2021-05-05
Fixed openapi 3 spec to default additionalProperties to true

### LINKED ISSUE #583: additionalProperties is not correctly defaulted to true
I am running into an issue where some of my openapi specifications are not generating to the correct typescript object because they do not include the `additionalProperties`. According to the OpenAPI specification this is default to true (https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md):

> additionalProperties - Value can be boolean or object. Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema. Consistent with JSON Schema, additionalProperties defaults to true.

With this being the case, shouldn't it automatically add the union with `& { [key: string]: any }` unless `additionalProperties` is set to something else i.e. a schema or `false`?

When I run the following openapi definition:

```
{
    "openapi": "3.0.0",
    "info": {
        "title": "Generated by cue.",
        "version": "no version"
    },
    "paths": {},
    "components": {
        "schemas": {
            "Blueprint": {
                "$ref": "#/components/schemas/Contract"
            },
            "Contract": {
                "type": "object",
                "required": [
                    "slug"
                ],
                "properties": {
                    "slug": {
                        "type": "string"
                    }
                }
            }
        }
    }
}
```

I get the typescript:

```
export interface paths {}

export interface components {
	schemas: {
		Blueprint: components['schemas']['Contract'];
		Contract: {
			slug: string;
		};
	};
}

export interface operations {}
```

where it should be `Contract: { slug: string; } & { [key: string]: any }`

Thanks for your help and all the work put into this project!

mehalter:
Resolves #583 . Let me know if this seems correct or if you have any comments/critiques!

Signed-off-by: Micah Halter <micah@balena.io>

### DIFF
--- a/src/transform/schema.ts
+++ b/src/transform/schema.ts
@@ -14,6 +14,7 @@ interface TransformSchemaObjMapOptions {
   formatter?: SchemaFormatter;
   immutableTypes: boolean;
   required?: string[];
+  version: number;
 }
 
 /** Take object keys and convert to TypeScript interface */
@@ -34,6 +35,7 @@ export function transformSchemaObjMap(obj: Record<string, any>, options: Transfo
     output += transformSchemaObj(value.schema || value, {
       formatter: options.formatter,
       immutableTypes: options.immutableTypes,
+      version: options.version,
     });
 
     // 4. close
@@ -46,6 +48,7 @@ export function transformSchemaObjMap(obj: Record<string, any>, options: Transfo
 interface Options {
   formatter?: SchemaFormatter;
   immutableTypes: boolean;
+  version: number;
 }
 
 /** transform anyOf */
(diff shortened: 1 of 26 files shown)


src/transform/schema.ts
```diff
@@ -112,13 +115,14 @@ export function transformSchemaObj(node: any, options: Options): string {
...
           required: node.required,
+          version: options.version,
         });
 
         // if additional properties, add an intersection with a generic map type
         let additionalProperties: string | undefined;
-        if (node.additionalProperties) {
-          if (node.additionalProperties === true || Object.keys(node.additionalProperties).length === 0) {
-            additionalProperties = `{ [key: string]: any }`;
+        if (node.additionalProperties || (node.additionalProperties === undefined && options.version === 3)) {
```
drwpow, line 123: 👍🏻

drwpow:
Reasoning makes sense! All on board for generating types that follow the spec. This looks great!

mehalter:
Ah I forgot to run the linter, let me do that to fix this test.

mehalter:
@drwpow updated the commit with linted code

drwpow:
@all-contributors please add @mehalter for code, test, bug

drwpow:
Released in `v3.2.4`! 🎉

radist2s:
@mehalter, @drwpow,  
seems, this release broke a lot of specifications that were already working.
According to OpenAPI v [3.0.2](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#properties): _Consistent with JSON Schema, additionalProperties defaults to true._
At the same time, according to [3.0.1](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.1.md#properties) `additionalProperties` is optional.

By the way with `oneOf` generated result could be useless with non object types:

```json
{
  "pictures": {
    "type": "array",
    "items": {
      "oneOf": [
        {
          "type": "integer"
        },
        {
          "$ref": "#/components/schemas/ItemsArtworksPictures"
        }
      ]
    }
  }
}
```

```typescript
{
  pictures?: ((number | components["schemas"]["ItemsArtworksPictures"]) & {
    [key: string]: any;
  })[];
}
```

drwpow:
Due to feedback, we’re going to revert this behavior in `2.4.0`, back to the way it was. However, you can get this behavior by opting in via the `--additional-properties` flag. So if you need that, simply add that flag (it’ll be a Node.js API option, too)!

PR #586, opened 2021-05-06
Fix required attribute by considering schemas.

### LINKED ISSUE #584: Property values with defaults are being incorrectly typed as possibly undefined when they shouldn't be.
Hi Team,

Thank you for a great module.

It seems the logic in the link below is insufficient to handle if a property able to be defined or undefined.

https://github.com/drwpow/openapi-typescript/blob/4c51afc45f449943bc9e2c58639606b15aaa88b8/src/transform/parameters.ts#L58

If a property isn't required in an object but that property has a default value specified, that property will never be undefined.  See the docs at:

https://swagger.io/docs/specification/describing-parameters/

Rusty

rustyconover:
Resolves #584

### DIFF
--- a/src/transform/parameters.ts
+++ b/src/transform/parameters.ts
@@ -1,5 +1,5 @@
-import { ParameterObject, ReferenceObject } from "../types";
-import { comment, tsReadonly } from "../utils";
+import { ParameterObject, ReferenceObject, SourceDocument } from "../types";
+import { comment, resolveRefIfNeeded, tsReadonly } from "../utils";
 import { transformSchemaObj } from "./schema";
 
 export function transformParametersArray(
@@ -7,10 +7,12 @@ export function transformParametersArray(
   {
     globalParameters,
     immutableTypes,
+    document,
     version,
   }: {
     globalParameters?: Record<string, ParameterObject>;
     immutableTypes: boolean;
+    document: SourceDocument;
     version: number;
   }
 ): string {
@@ -34,6 +36,8 @@ export function transformParametersArray(
         } else if (version === 3) {
           mappedParams[reference.in][reference.name || paramName] = {
             ...reference,
+            // They want to keep this ref here so it references
+            // the components, but it does violate the standards.
             schema: { $ref: paramObj.$ref },
           };
         }
@@ -55,20 +59,36 @@ export function transformParametersArray(
       if (paramObj.description) paramComment += paramObj.description;
       if (paramComment) output += comment(paramComment);
 
-      const required = paramObj.required ? `` : `?`;
+      let hasDefault = false;
+      if (paramObj.schema) {
+        const actualParam = resolveRefIfNeeded<ParameterObject>(document, paramObj.schema as ReferenceObject);
+
+        // Because the schema refs have been re-written to reference the global parameters
+        // need to look a little deeper to get the real schema.
+        if (actualParam != null && actualParam.schema != null) {
+          const actualParamSchema = resolveRefIfNeeded(document, actualParam?.schema);
+
+          if (actualParamSchema != null && actualParamSchema.default != null) {
+            hasDefault = true;
+          }
+        }
+      }
+
+      const required = (paramObj.required || hasDefault) ? `` : `?`;
       let paramType = ``;
       if (version === 2) {
         if (paramObj.in === "body" && paramObj.schema) {
-          paramType = transformSchemaObj(paramObj.schema, { immutableTypes, version });
+          paramType = transformSchemaObj(paramObj.schema, { immutableTypes, version, document });
         } else if (paramObj.type) {
-          paramType = transformSchemaObj(paramObj, { immutableTypes, version });
+          paramType = transformSchemaObj(paramObj, { immutableTypes, version, document });
         } else {
           paramType = "unknown";
         }
       } else if (version === 3) {
-        paramType = paramObj.schema ? transformSchemaObj(paramObj.schema, { immutableTypes, version }) : "unknown";
+        paramType = paramObj.schema ? transformSchemaObj(paramObj.schema, { immutableTypes, version, document }) : "unknown";
       }
       output += `    ${readonly}"${paramName}"${required}: ${paramType};\n`;
+
     });
     output += `  }\n`; // close in
   });
--- a/src/transform/schema.ts
+++ b/src/transform/schema.ts
@@ -1,7 +1,8 @@
-import { SchemaFormatter } from "types";
+import { SchemaFormatter, SchemaObject, SourceDocument } from "types";
 import {
   comment,
   nodeType,
+  resolveRefIfNeeded,
   transformRef,
   tsArrayOf,
   tsIntersectionOf,
@@ -13,6 +14,7 @@ import {
 interface TransformSchemaObjMapOptions {
   formatter?: SchemaFormatter;
   immutableTypes: boolean;
+  document: SourceDocument;
   required?: string[];
   version: number;
 }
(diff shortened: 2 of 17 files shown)


src/utils.ts
```diff
@@ -131,3 +132,26 @@ export function unrefComponent(components: any, ref: string): any {
   const [type, object] = ref.match(/(?<=\[")([^"]+)/g) as string[];
   return components[type][object];
 }
+
+function resolveDocumentReference<T>(document: any, reference: string): T | undefined {
+  if (reference[0] === "#") {
+    const parts = reference.replace(/^#\//, "").split("/");
+    const result = _.get(document, parts.join("."));
```
drwpow, line 139: I’d like to not add `lodash` to this library all for one function. Why not use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) instead?

src/transform/parameters.ts
```diff
@@ -34,6 +36,8 @@ export function transformParametersArray(
         } else if (version === 3) {
           mappedParams[reference.in][reference.name || paramName] = {
             ...reference,
+            // They want to keep this ref here so it references
+            // the components, but it does violate the standards.
```
drwpow, line 40: Could you explain a little more about how this violates the standards? Should this be removed?

src/transform/schema.ts
```diff
@@ -28,12 +31,22 @@ export function transformSchemaObjMap(obj: Record<string, any>, options: Transfo
...
+
+    let hasDefault = false;
+    if (value.schema != null) {
+      const schema = resolveRefIfNeeded<SchemaObject>(options.document, value.schema);
+      if (schema?.default != null) {
+        hasDefault = true;
+      }
+    }
+
+    output += `${readonly}"${key}"${required.includes(key) || hasDefault ? "" : "?"}: `;
```
drwpow, line 43: 👍🏻 Thanks for the fix!

src/transform/schema.ts
```diff
@@ -28,12 +31,22 @@ export function transformSchemaObjMap(obj: Record<string, any>, options: Transfo
     if (value.description) output += comment(value.description);
 
     // 2. name (with “?” if optional property)
-    output += `${readonly}"${key}"${required.includes(key) ? "" : "?"}: `;
+
+    let hasDefault = false;
+    if (value.schema != null) {
+      const schema = resolveRefIfNeeded<SchemaObject>(options.document, value.schema);
```
drwpow, line 37: Not a request for change, but is passing the entire document necessary here? Could we simply add refs that have default values to that `required[]` array? That way we’re figuring this stuff out in the context that has that information, and not at the schema object level. But happy for pushback if that‘s simply not possible.

src/transform/parameters.ts
rustyconover: Consider a document that has a reference as for a parameter to #/components/parameters/foobar.

This code requires the schema attribute to be that reference, but it turns out the code that I've written for determining if there is a default value for the parameter expects the schema for the parameter itself to be in the schema property that is getting written here.

So to access the true schema of the parameters the code has to follow .schema.schema, which seems quite awkward to reason about but this PR does implement it.

src/transform/schema.ts
rustyconover: The default value is set on the actual values themselves, as such it seems that the PR follows most closely to your goal.  I think manipulating the required array would certainly be not as clear.

Additionally, if the required array isn't specified there would be special code to handle that.

src/utils.ts
rustyconover: I've change to the single `get` function export of lodash.  

I'd rather not write some code to dynamically walk a JS object when lodash already has a function that already has it and is well tested.

drwpow: I know lodash isn’t gigantic, but aren’t you just using optional chaining here? It’s more about adding weight for something built into the language now.

src/transform/parameters.ts
drwpow: Ah yes for parameters we did try and flatten the schema, because in the TypeScript conversion we’re more interested in types, and we found the interface easier to use without nested `.schema` … `.schema` everywhere. I don’t know if I would say this “violates the standards” because TypeScript types are not bound to the standards. There are some opinions and interpretations we’re making here.

drwpow: I guess in a related question: what’s the purpose for this comment? Is this a TODO? Does this clarify the purpose of the existing code? I’m a little unclear why this was added when no other code here was changed.

src/transform/schema.ts
drwpow: If it’s a problem of the `required` array being nullable, we could fix that. From what I hear you saying, it would be possible to modify the `required` array. If that’s the case, I vote for doing that instead. I disagree that this is “clearer;“ this PR does take a module which previously was only worried about a single schema object at a time, and now this module has complete document context and so now its responsibilities are less clear.

Also, as an aside, this library does support [partial schemas](https://github.com/drwpow/openapi-typescript/#cli-options), so I’d like to avoid handling that logic in two places (here and at the top level) if possible.

If it’s simply an issue of `required` being too unclear a name, I vote for renaming that array. Or have it be an array with more data in there about default properties so that we don’t need to tunnel the document through multiple levels of modules.

src/utils.ts
rustyconover: Considering the paths are dynamic based on the contents of the ref, could you show me an example, I don't think optional chaining does what I think lodash's get method does.

It seems to me that together we are going going to iterate the parts of the path, with a current value then traverse down using each property name.  Optional chaining does not do that unless you write the property path explicitly in the code which we can't do because references are dynamic in both length and content.

src/transform/parameters.ts
rustyconover: It was a note to myself, you can merge it or not merge it as you like.  But it may help other users who want to edit this code understand what is going on.

drwpow:
@rustyconover sorry for the noise created from the recent changes. After digging further into the issue you raised, I align more and more with your approach. But I’ve decided I want to release remote schemas first (#602). So I’ve opened up #613 as a half-measure for what you’ve done here. I’d like for you to give it a look and see if you agree with it, pointing out that you handled `$ref` resolution whereas I didn’t.

How do you feel about releasing an incremental update for now, then coming back to resolving `$ref`s later in a world of remote schemas?

src/utils.ts
dnalborczyk: haven't looked at the details of the above, but you can use expressions with optional chaining:

```ts
const foo = 'x-foo'

obj?.[foo]?.['bar'].x
```

not sure how I'd approach the length, possibly with a loop ...

rustyconover:
Hi @drwpow I've just added fixes for headers being included by reference that weren't being properly handled.  

You can see more of the details in: #624

drwpow:
Sorry about the conflicts in the `main` branch; this PR just happened to get opened as remote $refs were getting merged.

As a general direction, I still don’t want to keep all the schemas in memory as you’re doing and passing them through to every function (especially as they’re getting transformed).

Instead, I’d like to see a solution that does that initial work up-front, and saves it on the property itself. For example, what if you did the work in `load.ts`, and saved everything to an `x-oapits-required` property like so?

```ts
JSON.parse(JSON.stringify(schemas[subschemaURL]), (k, v)) => {
  if (!v.$ref) return v; // select parent node of $ref;
  const { url: refURL, parts } = parseRef(v);
  if (refURL) return v; // don’t try and resolve remote schemas (for now)

  // find referenced node, if we can
  let foundRef: any = schemas[subschemaURL];
  for (const key of parts) {
    if (!foundRef[key]) return;
    foundRef = foundRef[key];
  }

  // check if the referenced node is required
  if (foundRef.required) v['x-oapits-required'] = true;
  return v;
});
```

Then in `transform/schema.ts` you should only need to check for that `x-oapits-required` property. This should be less far-reaching, and easier to work with with remote schemas.

Does this seem like a good suggestion?

fkowal:
@drwpow @rustyconover

this fix resolves the issue i've got in my project

I wanted to ask if there is anything I can help with to resolve this issue and get this released

PR seem stagnant

PR #602, opened 2021-05-19
Resolve remote schemas

drwpow:
Resolves #321 and adds support for:

```yaml
  $ref: "some-remote-schema.yaml#/components/schemas/Thing
```

The goal was pretty straightforward, but quite a bit of code had to change. The major change was moving the loader from the CLI into the Node API. Before, the Node API had to receive a JSON object. Now, the Node API can handle a string pointing to a local file or remote URL (README updated). There was a lot more prep work done, but that was merged in #612 to prepare for this PR (and make v3 much easier to patch, if needed).

This PR will act as a bit of an RFC to see if this is preferable before moving forward (see comments for generated schema).

---

**Merging this will mean version 4.0!** 🎉 It will disrupt Node.js API users by changing the main method from sync to async. Other than that, no other breaking changes, which means v4 will be backwards-compatible with v3 for all CLI users.

~~Before merging, I’d like to resolve #607 and #584 in v3. And this feature will be the first release of v4. So this PR will be merged as soon as those are resolved and published.~~

Last outstanding item before merging: #616. This will follow that!

PR #612, opened 2021-05-25
Change additionalProperties default

### LINKED ISSUE #607: New additionalProperties default breaks `oneOf`
I experience exactly the problem that @radist2s described in https://github.com/drwpow/openapi-typescript/pull/585#issuecomment-840149431:

```json
{
  "pictures": {
    "type": "array",
    "items": {
      "oneOf": [
        {
          "type": "integer"
        },
        {
          "$ref": "#/components/schemas/ItemsArtworksPictures"
        }
      ]
    }
  }
}
```
results in 
```typescript
{
  pictures?: ((number | components["schemas"]["ItemsArtworksPictures"]) & {
    [key: string]: any;
  })[];
}
```
which is useless (and wrong I hope?).

I'm reverting to `3.2.3` for now but it'd be great to have this fixed soon! :)

drwpow:
Resolves #607. A recent change (#585) allowed for the automatic addition of additional properties, so `& { [key: string]: any }` was added to all schema objects. The intent was to better reflect OpenAPI’s specification, and there is precedent for that. However, based on feedback, and previous expectations this library set, it was a little too disruptive.

And so, this PR reverts that behavior, turning this **off by default**`, but adding a new `--additional-properties` flag to enable this behavior (outlined in the README). This is a good compromise of allowing that behavior, but requiring it to be opt-in (which is also probably better to be stricter by default).


src/transform/index.ts
```diff
@@ -1,165 +1,137 @@
...
+    const required = new Set(Object.keys(schema));
+    switch (ctx.version) {
       case 2: {
-        return `export interface definitions {\n  ${transformSchemaObjMap(schema, {
-          formatter,
-          immutableTypes,
-          required: Object.keys(schema),
-          version,
-        })}\n}`;
+        output.definitions = transformSchemaObjMap(schema, { ...ctx, required });
```
drwpow, line 22: Changing `transformAll()`’s default output to be an object rather than a string is required for the upcoming #602. But I wanted to get that change in before we release v4, so that v3 patches will be easier to work on (less codebase distance between v3 and v4)

PR #613, opened 2021-05-25
Mark schema objects with default values as non-nullable

### LINKED ISSUE #584: Property values with defaults are being incorrectly typed as possibly undefined when they shouldn't be.

drwpow:
Partially fixes #584.

This is a very good fix pointed out by @rustyconover, and handled in PR #584, but that currently would be blocking remote schemas, as `$ref`s would no-longer be resolvable (or at least, not easily). This PR does half of #584, where `$ref`s are not resolved. But it does improve the handling of `default: ` without blocking remote schemas.

### DIFF
--- a/src/transform/schema.ts
+++ b/src/transform/schema.ts
@@ -15,10 +15,14 @@ interface TransformSchemaObjOptions extends GlobalContext {
   required: Set<string>;
 }
 
+function hasDefaultValue(node: any): boolean {
+  if (node.hasOwnProperty("default")) return true;
+  // if (node.hasOwnProperty("$ref")) return true; // TODO: resolve remote $refs?
+  return false;
+}
+
 /** Take object keys and convert to TypeScript interface */
 export function transformSchemaObjMap(obj: Record<string, any>, options: TransformSchemaObjOptions): string {
-  const readonly = tsReadonly(options.immutableTypes);
-
   let output = "";
 
   for (const k of Object.keys(obj)) {
@@ -27,7 +31,9 @@ export function transformSchemaObjMap(obj: Record<string, any>, options: Transfo
     if (v.description) output += comment(v.description);
 
     // 2. name (with “?” if optional property)
-    output += `${readonly}"${k}"${options.required.has(k) ? "" : "?"}: `;
+    const readonly = tsReadonly(options.immutableTypes);
+    const required = options.required.has(k) || hasDefaultValue(v.schema || v) ? "" : "?";
+    output += `${readonly}"${k}"${required}: `;
 
     // 3. transform
     output += transformSchemaObj(v.schema || v, options);
--- a/tests/bin/expected/prettier-js.ts
+++ b/tests/bin/expected/prettier-js.ts
@@ -72,7 +72,7 @@ export interface components {
       shipDate?: string
       /** Order Status */
       status?: 'placed' | 'approved' | 'delivered'
-      complete?: boolean
+      complete: boolean
     }
     Category: {
       id?: number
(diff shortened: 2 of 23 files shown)


rustyconover:
Looks awesome!

src/transform/schema.ts
```diff
@@ -15,10 +15,14 @@ interface TransformSchemaObjOptions extends GlobalContext {
   required: Set<string>;
 }
 
+function hasDefaultValue(node: any): boolean {
+  if (node.hasOwnProperty("default")) return true;
```
drwpow, line 19: This is the only change here: check if a schema object has a `default` value. But like the description states, we can’t currently look this up for `$ref` (which might be a remote schema)

tests/bin/expected/prettier-js.ts
```diff
@@ -72,7 +72,7 @@ export interface components {
       shipDate?: string
       /** Order Status */
       status?: 'placed' | 'approved' | 'delivered'
-      complete?: boolean
+      complete: boolean
```
drwpow, line 75: Updated snapshots show it’s improved a lot of schemas

gr2m:
Do you mind if we revert this change until we figure out to differentiate between request parameter types and response types? This change made a bunch of `requestBody` parameters from GitHub's OpenAPI spec required because there is a `defaults` key, but that is incorrect. The parameters are not required, the spec just states what value the parameter will get if it is not set.

rustyconover:
Hi Gregor,

I disagree a bit with you, I believe this change is correct for my use
cases but I’d love to understand your perspective more.

I’m considering the types from the perspective of a server implementing the
API (a handler function that runs after parsing and validating) and from
that perspective the values are required because they will be defined when
passed into the handler. The default values will be set.

If we’re using the types from a client you’re right the values don’t have
to be required, since the server should set the default values.

Maybe there needs to be a switch for the perspective of the interface?
Indicating client or server based types would be the description of the
option.

Rusty

drwpow:
@gr2m thanks for your feedback! And apologies for the changes to the GitHub schema. I too am a bit on the fence about this one, because there’s a lot of overlap with your business logic, and where these values get initialized. Do these types sit between the API response and your business logic? If so, they should probably be stricter (err on the side of nullable). Or do these types sit between the business logic and your frontend? If so, the change to respect `default` may be preferred.

If this is causing issues for people, I’d rather err on the side of doing what the library was doing before, and make this behavior opt-in via a flag (e.g. `--default-non-nullable` or something). Would that be doable?

gr2m:
GItHub's OpenAPI Spec is for public consumption. They use it internally as well, but it's also used to generate the docs at https://docs.github.com/en/rest/reference. So if a default value is set on a request parameter, it cannot be interpreted as a required parameter. 

I use the OpenAPI spec to generate several SDKs (e.g. https://github.com/octokit/plugin-rest-endpoint-methods.js/), and they types (github.com/octokit/openapi-types.ts and github.com/octokit/types.ts), so I have the API consumer perspective.

> If this is causing issues for people, I’d rather err on the side of doing what the library was doing before, revert the behavior, and make this behavior opt-in via a flag (e.g. `--default-non-nullable` or something). Would that be doable?

yes that sounds like a good plan

drwpow:
@gr2m Fixed in `v3.4.0` and a `--default-non-nullable` flag was added to opt into this behavior

Issue #526, opened 2021-03-17
Feature request: Run on glob pattern, output to directory


icopp:
A simple example of possible syntax:

```
openapi-typescript "api/*.yml" src/generated
```

This would run `openapi-typescript` once for each `.yml` file in `api`, and output a file with a matching name (but with `.ts` extension) to `src/generated`.

Alternately, this could use something like a `-m` (for 'match') flag to avoid any backwards compatibility questions:

```
openapi-typescript -m "api/*.yml" src/generated
```

Other command line options would be identical to the single-file usage, but it might be useful to also have a `-p` flag to indicate running in parallel, where the default would be sequential (to play nice with limit-resource environments).

drwpow:
Thanks for the suggestion! You know, I thought of this, but the main reason I didn’t implement this was I’d be scared that in auto-naming everything as you suggested, someone would ask for a different way to name the output files, and it would just be a pain.

How do you feel about the current [Generating multiple schemas](https://github.com/drwpow/openapi-typescript#generating-multiple-schemas) section of the README? Would you suggest an alternate approach? I’d like to give people as many options as possible without locking them into a restriction of this library.

icopp:
Doing it in that style rapidly becomes pretty clunky as you accumulate multiple schema files (as you pretty rapidly can if you're trying to make specific bits of documentation on a big project for multiple external integrations), which is why I made the request to do it in a more integrated way.

drwpow:
That‘s fair. I think it‘s something we can probably support, with the given restriction that renaming simply isn‘t allowed; filenames and folders will be carried over exactly 1:1. And for anything else, using the Node API is recommended.

benkroeger:
I'm using a shell script for generating from multiple schemas (also prepending with some eslint rules)

```sh
#!/bin/sh
MYPWD=${PWD}
PACKAGE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." &> /dev/null && pwd )"

cd $PACKAGE_DIR

for spec_file in openapi-specs/*.yaml
do
  file_name="${spec_file##*/}"
  api_name="${file_name%.*}"
  typescript_file="src/apigee-service/@generated-types/${api_name}.d.ts"

  # `openapi-typescript` generates interface names in all lowercase - which doesn't comply with the project's lint rules
  echo "/* eslint-disable @typescript-eslint/naming-convention */\n" > $typescript_file

  npx openapi-typescript $spec_file \
[...]

PR #615, opened 2021-05-26
adds support for glob-style input spec paths and directory output paths

sharmarajdaksh:
This PR adds support for glob-style inputs and directory output paths for the openapi-typescript cli, as mentioned in #526.
With this, something like the following should be possible:
```
npx openapi-typescript specs/*.yaml -o outdir/
```
- Maintains backwards compatibility, and throws reasonable helpful errors.
- If the input is a glob path and -o flag it specified, it must be a directory
- Cannot specify remote resources using glob-type paths


sharmarajdaksh:
Needs some fixes! sorry

PR #616, opened 2021-05-26
adds support for glob-style input spec paths and directory output paths

sharmarajdaksh:
This PR adds support for glob-style inputs and directory output paths for the openapi-typescript cli, as mentioned in #526.
With this, something like the following should be possible:
```
npx openapi-typescript specs/*.yaml -o outdir/
```
- Maintains backwards compatibility, and throws reasonable helpful errors.
- If the input is a glob path and -o flag it specified, it must be a directory
- Cannot specify remote resources using glob-type paths


tests/bin/cli.test.ts
```diff
@@ -31,4 +31,14 @@ describe("cli", () => {
...
+
+  it("supports glob paths", () => {
+    const expectedPetstore = fs.readFileSync(path.join(__dirname, "expected", "petstore.ts"), "utf8");
+    const expectedManifold = fs.readFileSync(path.join(__dirname, "expected", "manifold.ts"), "utf8");
+    execSync(`../../bin/cli.js \"specs/*.yaml\" -o generated/`, { cwd: __dirname }); // Quotes are necessary because shells like zsh treats glob weirdly
+    const outputPetstore = fs.readFileSync(path.join(__dirname, "generated", "petstore.ts"), "utf8");
+    const outputManifold = fs.readFileSync(path.join(__dirname, "generated", "manifold.ts"), "utf8");
+    expect(outputPetstore).toBe(expectedPetstore);
+    expect(outputManifold).toBe(expectedManifold);
+  });
```
drwpow, line 43: Nice test!

bin/cli.js
```diff
@@ -104,15 +143,19 @@ async function main() {
...
-    const timeEnd = process.hrtime(timeStart);
-    const time = timeEnd[0] + Math.round(timeEnd[1] / 1e6);
-    console.log(green(`🚀 ${pathToSpec} -> ${bold(cli.flags.output)} [${time}ms]`));
-  } else {
-    // output option 2: stdout
-    process.stdout.write(result);
+  let result = "";
+  for (const specPath of inputSpecPaths) {
+    // append result returned for each spec
+    result += await generateSchema(specPath);
```
drwpow, line 158: This is clever. I hadn’t thought about the `stdout` case but this seems pretty handy.

drwpow: Oh wait—we’re logging to `stdout` elsewhere. Truth be told, I don’t know why we’re returning the `result` here; it doesn’t seem to be helpful (and possibly wasting memory?). Either way, this is fine to leave in this PR.

sharmarajdaksh: Hey, I was initially confused as well about what the reason to return the result was. In user usage, the returned result doesn't matter, but I think it gets used in certain tests so it should be kept in.

drwpow: OK. Yeah leave it in for now, and I can revisit that post-merge.

sharmarajdaksh: Alright! 
Rebasing and updating expected schemas fixed the broken test, so I think this PR should be ready now

drwpow:
@all-contributors please add @sharmarajdaksh for code

Issue #554, opened 2021-04-06
Use unknown instead of any


andrenarchy:
It would be great if the resulting ts would not use `any` but `unknown` instead.

drwpow:
Thanks for the suggestion. Do you have an example of a schema you’re using, what it’s generating, and what you’d like it to be instead?

Also would you be open to opening a PR for this?

andrenarchy:
One example is a schema I have with the following property somewhere:
```
          "presets": {
            "description": "JSON structure containing the preset value for created/updated items.",
            "type": "object",
            "nullable": true
          },
```
This gets transformed into:
```
presets?: { [key: string]: any } | null
```

The use of `any` is generally discouraged since it can be accessed in any way. But that's not really what we want for a type where we don't have further knowledge. So the better option would be to use `unknown` for all types that are currently generated as `any` because then it's up to the user to do further checks on the type. In the above example I'd expect:
```
presets?: { [key: string]: unknown } | null
```

See also [eslint](https://github.com/typescript-eslint/typescript-eslint/blob/v4.24.0/packages/eslint-plugin/docs/rules/no-explicit-any.md) (which is complaining about `any` in the `recommended` rule set) and the [TypeScript announcement](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type) that says 

> `unknown` is the type-safe counterpart of `any`

which really is the essence I think. :)

drwpow:
Thanks for providing more info. I also see that #625 was opened to address this. If it’s all the same to people, I don’t see why `unknown` isn’t an improvement. I’m on board with this change.

drwpow:
Added in #769!

PR #620, opened 2021-05-28
Always checks for `required` properties, even if they are missing from `properties`

### LINKED ISSUE #619: `required` properties are not respected when they are missing from `properties`
Example schema
```
type: object
properties: 
  email: 
    type: string
required: ["user"]
```
should generate a type that requires `user` to be present. Today it generates a type with only `email` being the optional argument.

This is part of spec and commonly used for when reusing an object and making optional fields required, for example

```
LogInReq:
  type: object
  allOf: 
    - $ref: '#/someCommonObject' 
    - type: object
       required: ["user", "password"]
```

`LogInReq` now requires `user` and `password` to be present while all other fields in `someCommonObject` to be their default setting.

Working on a PR.

This will change a lot of behavior for the library and curious about people's thoughts.

shuluster:
Fixes #619


src/transform/schema.ts
```diff
@@ -45,6 +45,19 @@ export function transformSchemaObjMap(obj: Record<string, any>, options: Transfo
...
 
+/** make sure all required fields exist **/
+export function addRequiredProps(properties: Record<string, any>, required: string[]): string[] {
+  const missingRequired = required.filter((r: string) => !(r in properties));
+  if (missingRequired.length == 0) {
+    return [];
+  }
+  let output = "";
+  for (const r of missingRequired) {
+    output += `${r}: any;\n`;
```
drwpow, line 56: In anticipation of #625 (not merged yet), let’s use `unknown` instead of `any` here

src/transform/schema.ts
```diff
@@ -45,6 +45,19 @@ export function transformSchemaObjMap(obj: Record<string, any>, options: Transfo
   return output.replace(/\n+$/, "\n"); // replace repeat line endings with only one
 }
 
+/** make sure all required fields exist **/
+export function addRequiredProps(properties: Record<string, any>, required: string[]): string[] {
```
drwpow, line 49: This is a minor thing, but I’d recommend using `Set<string>` rather than `string[]`, so that duplicates will automatically be removed for us. (e.g. `required: ["foo", "foo", "foo", "bar"]`). I don’t think it’s invalid to repeat properties in `required`, but the generated TypeScript will be incorrect if this happens (I know that this is on the schema author, but the error will be really weird so I’d rather just avoid it).

src/transform/schema.ts
drwpow:
```suggestion
    output += `${r}: unknown;\n`;
```

src/transform/schema.ts
```diff
@@ -95,14 +108,16 @@ export function transformSchemaObj(node: any, options: TransformSchemaObjOptions
...
-
+        const missingRequired = addRequiredProps(node.properties || {}, node.required || []);
         // if empty object, then return generic map type
         if (
           !isAnyOfOrOneOfOrAllOf &&
           (!node.properties || !Object.keys(node.properties).length) &&
           !node.additionalProperties
         ) {
-          output += `{ ${readonly}[key: string]: any }`;
+          const emptyObj = `{ ${readonly}[key: string]: any }`;
```
drwpow, line 118: Same here: 
```suggestion
          const emptyObj = `{ ${readonly}[key: string]: unknown }`;
```

tests/schema.test.ts
```diff
@@ -9,7 +9,7 @@ const defaults = {
   immutableTypes: false,
   required: new Set<string>(),
   rawSchema: false,
-  version: 3,
+  version: 3
```
drwpow, line 12: Not blocking for this PR, but if you could use the local Prettier settings for this project that’d cut down on some unnecessary diff here

tests/schema.test.ts
```diff
@@ -342,10 +342,44 @@ describe("SchemaObject", () => {
 }>)) & ({
 "a"?: string;
 
+})`);
+    });
+
+    it("empty object with required fields", () => {
```
drwpow, line 348: Awesome tests! ✨

drwpow:
Requested a few minor changes; once those are in I’d be happy to merge

drwpow (changes requested):
A few minor changes requested in comments

shuluster:
Will address those later today.

src/transform/schema.ts
shuluster: Thinking about this, this might be a breaking change for a lot of folks, who use empty projects and rely on the any.  In my project, this wouldn't be a problem. 

I will include this change for now. Let me know your thoughts

shuluster:
Updated the branch based on feedbaack

src/transform/schema.ts
drwpow: I’m really sorry but I think you do raise a good point and you’ve changed my mind on this—this actually would cause disruption for people. And #554 didn’t really get a lot of support from people, so I’m starting to second-guess it. This PR shouldn’t be blocked by another that may or may not merge.

It seems there is another merge conflict (apologies); if you’re able to update that would you mind changing this back to `any` again? Again I’m sorry and this is 100% on me. If I don’t hear from you soon, I can update this PR myself and still give you credit for it. Consider this PR ready-to-go!

drwpow (approved):

shuluster:
fixed the merge conflict.

drwpow:
@all-contributors please add @shuluster for code

Issue #604, opened 2021-05-20
Support "readOnly" and "writeOnly" OpenAPI properties


Rycochet:
See https://swagger.io/docs/specification/data-models/data-types/#readonly-writeonly

These modifiers are on specific properties and allow for a single definition to be used in both read & write endpoints, but provide a different shape to each -

```js
// ... "Data": { ...
"properties": {
  "id": { // Returned by GET, not used in POST/PUT/PATCH
    "type": "integer",
    "readOnly": true
  },
  "username": {
    "type": "string"
  },
  "password": { // Used in POST/PUT/PATCH, not returned by GET
    "type": "string",
    "writeOnly": true
  }
}
```

Currently this just generates a single interface similar to -

```ts
export interface components {
  schemas: {
    Data: {
      id: number;
      username: string;
      password: string;
    };
  };
}
```

Unfortunately that makes it unusable for any endpoint as GET doesn't show the `password` and POST etc doesn't want an `id`. One way to change this is to use multiple interfaces (alternatively using a suffix for the schema might be suitable - but there is a possibility of a name clash that way) - possibly something like -

```ts
export interface components {
  read_schemas: {
    Data: {
      id: number;
      username: string;
    };
  };
  write_schemas: {
    Data: {
      username: string;
      password: string;
    };
  };
  schemas: {
    Data: components["read_schemas"]["Data"] | components["read_schemas"]["Data"];
  }
}
```

Alternatively it could go the other way around, although I feel this makes the typing slightly less useful (adding type guards to code is pretty simple, but they like the union types more) -

```ts
export interface components {
  read_schemas: {
    Data: components["schemas"]["Data"] & {
      id: number;
    };
  };
  write_schemas: {
    Data: components["schemas"]["Data"] & {
      password: string;
    };
  };
  schemas: {
    Data: {
      username: string;
    };
  }
}
```

If there's no `readOnly` or `writeOnly` properties then nothing would need to change for any schemas that don't use them.

drwpow:
Interesting! I had been wondering if `readOnly` and `writeOnly` needed representation in TypeScript somehow, but I had personally never used them in a schema. Your example makes sense.

[...]

PR #621, opened 2021-05-28
`readOnly` and `writeOnly` properties

Rycochet:
This is a basic implementation for https://github.com/drwpow/openapi-typescript/issues/604

Globally this adds a comment to the end of fields that only exist in `readOnly` or `writeOnly` form, but it also adds an optional `-s` / `--splitSchema` flag that will instead create a slightly different structure.

Data -

```js
// ... "Data": { ...
"properties": {
  "id": { // Returned by GET, not used in POST/PUT/PATCH
    "type": "integer",
    "readOnly": true
  },
  "username": {
    "type": "string"
  },
  "password": { // Used in POST/PUT/PATCH, not returned by GET
    "type": "string",
    "writeOnly": true
  }
}
```

Without the flag -

```ts
export interface components {
  schemas: {
    Data: {
      id: number; // GET requests only
      username: string;
      password: string; // POST/PUT/PATCH responses only
    };
  };
}
```

With `-s` / `--splitSchema`

```ts
export interface components {
  "x-requestSchemas": {
    Data: {
      id: number; // GET requests only
      username: string;
    };
  };
  "x-responseSchemas": {
    Data: {
      username: string;
      password: string; // POST/PUT/PATCH responses only
    };
  };
  "schemas": {
    Data: components["x-requestSchemas"]["Data"] | components["x-responseSchemas"]["Data"];
  }
}
```

Not shown, but with the flag any request or response fields with a $ref will point at the correct interface (and anything outside of there will point at the base `schemas`) - Typescript will understand what to do if you point at those rather than the base `schemas` data (which can be used with typeguards for simpler source).

There are minimal tests just to check it is working in general (something testing the cli flag would be far better).

There is also no checking of if it is required - that is fully up to the user.

Ideally a better solution would check if it is needed anywhere in the data tree, and then only use the split schema in that case, and keep it in the main `schemas` structure otherwise, however given the simple structure of the code this would add a **lot** of overhead, and would be best with a rewrite of the logic being used (not something I really want to think about given the amount of time it could take).

Technically `readOnly` is supported in both v2 & v3, however `writeOnly` is only supported in v3 (similar to how `additionalProperties` gets a slightly different meaning between the versions) - yet another reasons to leave this behind a flag rather than as a default.


drwpow:
This is really clever! I like your implementation. And the usage of a flag is great. I don’t see any conflict that could happen, since technically `components["responseSchemas"]` is an unused space, I think that will generate nicely.

Bikeshedding a little, I was seeing if the OpenAPI spec allows unknown fields in the `components` object. I think it does, as it doesn’t specifically disallow it, but it does give this hint:

> This object MAY be extended with Specification Extensions (`x-*`)

What do you think about `x-requestSchemas` and `x-responseSchemas` instead for `--splitSchema`? Though I think your implementation may work, the `x-*` prefix would seem to make it a little more OpenAPI compliant. Open to your thoughts there.

Rycochet:
Adding the `x-` prefix makes a lot of sense - will fix the conflicts and update for that :-)

mrdannael:
Hi! What is the status of this PR? :smiley:  This is something that i'm looking for currently :1st_place_medal:

YuJianrong:
Any update on this PR? It's created almost one year ago...

Fi1osof:
Waiting too(((

prevostjules:
Hi,

Would it be possible to have updates on that great feature ? Is a merge planned soon ? :) 

Thanks all for your terrific work

PR #625, opened 2021-05-30
Change from using "any" to "unknown" for unspecified types.

rustyconover:
In accordance with Typescript guides the use of the "any" type doesn't
enforce many guarantees, it will match any type.  In situations where
the type is unspecified it is better to use the "unknown" which is
closely related to the "any" type but requires a type assertion or
control flow based narrowing before the type can be used.  The use of
the "unknown" type seems much more appropriate to use then "any" for
these cases.

See:

https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type

For more details.

### DIFF
--- a/src/transform/schema.ts
+++ b/src/transform/schema.ts
@@ -80,7 +80,7 @@ export function transformSchemaObj(node: any, options: TransformSchemaObjOptions
       case "string":
       case "number":
       case "boolean": {
-        output += nodeType(node) || "any";
+        output += nodeType(node) || "unknown";
         break;
       }
       case "enum": {
@@ -102,7 +102,7 @@ export function transformSchemaObj(node: any, options: TransformSchemaObjOptions
           (!node.properties || !Object.keys(node.properties).length) &&
           !node.additionalProperties
         ) {
-          output += `{ ${readonly}[key: string]: any }`;
+          output += `{ ${readonly}[key: string]: unknown }`;
           break;
         }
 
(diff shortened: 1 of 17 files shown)


drwpow (approved):
I’m on board with this! Thank you for making this improvement.

drwpow:
I’m actually starting to second-guess this PR. Though I do think `unknown` may be a technically-better choice, it would be disruptive for people and I don’t think for most people there’s a clear benefit. I’d be OK in revisiting this as a breaking change later down the line, but perhaps we delay on this for now.

rustyconover:
Hi @drwpow,

I appreciate your reluctance to "break" things for users.   While this change is increasing the safety of the generated interfaces it doesn't really do anything "wrong".   If people are already relying on the behavior of "any" in their code, they're at risk right now of getting a type they don't actually expect.   With "unknown", they need to be explicit with their types.  This goes back to the [Robustness Principal](https://en.wikipedia.org/wiki/Robustness_principle).

This change matters more for implementers of OpenAPI interfaces compared to clients.  If you're sending data the use of "any" is no big deal because it is not your responsibility to handle the data (you're merely sending it on), but if you're the server/implementor of the OpenAPI interface it matters quite a bit.

It doesn't seem to me that many people use `openapi-typescript` for the server implementation side of things, and since I'm doing that I'm running into these issues.

Rusty

dnalborczyk:
@rustyconover generally I'm in favor of using `unknown` instead of `any` as well. I agree with @drwpow that this change could cause some breakage for some users of this project, although I haven't discovered any usage of `any` types in my own schema yet.

to get around this issue you could put this functionality behind a flag, which can then be removed with the next major bump.

drwpow:
Sorry for the delay on this—I had not had time the past few months to properly maintain the project. I will make this change in #769 and it will go out for the next release! And though this PR won’t be merged I’d still like to attribute credit for this code-writing and improvement.

drwpow:
@all-contributors please add @rustyconover for code

Issue #617, opened 2021-05-27
References to arbitrary data


Peteck:
So because of the missing ability in the OpenAPI standard to mark a property to be required independently of it being read-only or write-only, I'm making references to properties directly in cases we need a property that has been marked as required to be optional. 

A quick example could be that you have a product with a name. The name is always required in responses. You might want it to be optional in a patch operation because only the given properties should be patched.

So my work-around was to add every writable property as the request body and then don't make them required. Just like so

```yaml
requestBody:
  content:
    application/json:
      schema:
        type: object
        properties:
          name:
            $ref: '#/components/schemas/Product/properties/name'
```

This is working with all the other tools we're using, but after the `additionalProperties` default has been set to `false`, then the generated schema fails, as it should because of the missing `properties` object around the properties. This in general might be a problem with the structure of the generated schema types at the moment, because it may not be 1:1 with the provided API specification. I'm not quite sure if this problem might occur in other places as well.

I've made a small hack to the utility function `transformRef` but might become a unmaintainable solution moving forward. See it here [https://github.com/venue-pos/openapi-typescript/commit/ad6be15bbaeddf93171dbda34683e3d47ad308cc](https://github.com/venue-pos/openapi-typescript/commit/ad6be15bbaeddf93171dbda34683e3d47ad308cc)

What is a better general solution that could actually be implemented?

drwpow:
I think you’re asking something different of #604, but I also see some overlap there. In your example about required-ness being conditional of response vs PATCH (request), that would be `readOnly` / `writeOnly`, no? Or are you saying you need even more control than that?

Peteck:
> Or are you saying you need even more control than that?

Yes. In a response I want the type `product.name` to be required, as in it can not be undefined. But when sending a request using the same component as the body I don't want it to be required.

read-only and write-only is good for let's say `product.id`, as the ID may never change and may be generated by the API, so one would set it as required and read-only. But in this case, I want to be able to change the name of the product by using a patch operation, but I don't want it to be required in the request body, as I only want the operation to change the defined properties in the request body.

And I don't really want to use `anyOf` and make all the properties nullable, because this will not fulfill my need in the patch operation request body, as one would need to define all other untouched properties as `null` and it would also conflict with properties that actually is nullable.

Rycochet:
[...]

PR #626, opened 2021-05-31
Support for making references to schema properties

Peteck:
Fixes issue #617


drwpow:
Cool. This looks great, but I wouldn’t want to merge this without adding a test first. Can you write a unit test for this?

I think a great place would be in `schema.test.ts`, there is a `$ref` and `$ref (external)` test; maybe this could be a `$ref (properties)` test? Just in one place should be enough to make sure this is working as intended.

drwpow:
Actually, on second thought, now that #602 is merged, we can keep your approach, but we’ll have to move it to [load.ts#L152](https://github.com/drwpow/openapi-typescript/blob/main/src/load.ts#L152). Something like:

```diff
  let [base, ...rest] = parts;
+ if (rest[rest.length - 2] === 'properties') rest.splice(rest.length - 2, 1); // if "properties" is next-to-last segment, remove
  return `${base}["${rest.join('"]["')}"]`;
```

The change was due to the fact that because we resolve remote schemas, we have to resolve `$ref`s at the very beginning in order to figure out what to load. And we have to transform them at the very beginning, too, because we only have that context during resolution. But other than that, this should look familiar to your approach here; it just moved to a different spot in the codebase.

Peteck:
> Cool. This looks great, but I wouldn’t want to merge this without adding a test first. Can you write a unit test for this?

Sure

> Actually, on second thought, now that #602 is merged, we can keep your approach, but we’ll have to move it to [load.ts#L152](https://github.com/drwpow/openapi-typescript/blob/main/src/load.ts#L152).

All right, I will move it :slightly_smiling_face:

drwpow:
Just checking in. Are you still able to make this change? Then we can get this merged.

Sorry again for the v4 disruption on this!

drwpow (approved):
Approved pending moving to `load.ts` (as this utility no longer exists)

Peteck:
I chose to make the test as a schema-load-test in v2 and v3. That is because the check and change is made at the load phase, and the test in `schema.test.ts` is not for tests in the loading phase.

drwpow:
> I chose to make the test as a schema-load-test in v2 and v3. That is because the check and change is made at the load phase, and the test in `schema.test.ts` is not for tests in the loading phase.
> 
> And sorry about the late response, I've been busy 🙂

No worries at all! I totally understand—I only work on this project when I have time as well. 🙂

avaly:
Is there anything blocking merging this PR?

I also ran into this issue recently and would like to get this patch available.

drwpow:
> Is there anything else blocking merging this PR?
> 
> I also ran into this issue recently and would like to get this patch available.

Nope! Sorry—I fell a little behind on maintenance for this project and am just now catching up. Will publish this in the next release!

drwpow:
@all-contributors please add @Peteck for code, bug

PR #632, opened 2021-06-03
Add docs for glob feature to README

drwpow:
Adds glob docs to README, (#616), along with credit.

Also fixes an unreleased bug where specifying a glob pattern wouldn’t create the folder (it used the file behavior, where it would stop at the _parent_ directory, as with single files, however with globs we need that last directory, too)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .cache
 .DS_Store
-**/generated/*.ts
+**/generated/
 coverage
 dist
 node_modules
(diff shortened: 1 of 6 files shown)


drwpow:
cc @sharmarajdaksh please take a look at the README, and please submit a PR if you’d like to change anything 🙂

sharmarajdaksh:
Looks good, @drwpow, I don't really see anything that needs to be changed. 
Cool work on parallelizing the spec generation for globs though!

Issue #571, opened 2021-04-19
Generate typings from any JSON schema


codeclown:
It would be really useful if openapi-typescript would support generating typings from any JSON schema specification, not necessarily an OpenAPI spec with endpoints etc.

For example, with Helm charts it's possible to define a simple `values.schema.json` file which Helm will then use to validate variable values at runtime. This is just one example of where one might have a JSON schema spec that's not an OpenAPI spec.

Would this be possible to implement? Would you accept a PR if so?

---

## Steps to reproduce

1. Save a basic JSON schema in file `values.schema.json`:

```json
{
  "$schema": "http://json-schema.org/schema#",
  "type": "object",
  "required": [
    "image"
  ],
  "properties": {
    "image": {
      "type": "object",
      "required": [
        "repository",
        "pullPolicy"
      ],
      "properties": {
        "repository": {
          "type": "string",
          "pattern": "^[a-z0-9-_]+$"
        },
        "pullPolicy": {
          "type": "string",
          "pattern": "^(Always|Never|IfNotPresent)$"
        }
      }
    }
  }
}
```

2. Run openapi-typescript:

```bash
npx openapi-typescript values.schema.json --output schema.ts
```

## Expected results

openapi-typescript would create file `schema.ts` with typings of the specified schema.

## Actual results

```
npx: installed 72 in 6.314s
✨ openapi-typescript 3.2.3
🔭 Loading spec from values.schema.json…
Error: 🚏 version missing from schema; specify whether this is OpenAPI v3 or v2 https://swagger.io/specification
    at Object.swaggerVersion (/Users/martti/.npm/_npx/21878/lib/node_modules/openapi-typescript/dist/cjs/utils.js:57:11)
    at swaggerToTS (/Users/martti/.npm/_npx/21878/lib/node_modules/openapi-typescript/dist/cjs/index.js:31:61)
    at main (/Users/martti/.npm/_npx/21878/lib/node_modules/openapi-typescript/bin/cli.js:75:18)
```

drwpow:
I don’t see why we couldn’t support this! Because OpenAPI is based on JSON schema this seems like a natural fit. And yes I’d be open to a PR adding this.

LironHazan:
I'd tried to utilize another tool for that but I'm not 100% sure it currently fits the need, the output is as follows:

```
export interface Simple {
  image: {
    repository: string;
    pullPolicy: string;
    [k: string]: unknown;
  };
  [k: string]: unknown;
}
```
[...]

PR #636, opened 2021-06-04
[Feat] Added a choice to CLI of generating types from plain json like…

LironHazan:
… in json-schema-to-typescript

1. I supplied a quick fix for using jsons which aren't part of a the open-api, can be tested as follows:
` ./bin/cli.js --jsonToTs=true examples/simple.json --output schema.ts`
2. I Ran the existing tests, all green
3. Documented the option in readme

In general you can decide if to make a more "plugable" design, giving an easy way to developers to add their own "plugins" or running pre-procedures/processing :)

Screen Shot 2021-06-04 at 18 25 41

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ dist
 node_modules
 yarn.lock
 yarn-error.log
+.idea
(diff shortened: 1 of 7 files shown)


LironHazan:
addressing this issue --> https://github.com/drwpow/openapi-typescript/issues/571

drwpow:
Hey thanks for opening this PR. I think when I agreed to the work, I assumed that someone else hadn’t already solved this problem. But seeing that `json-schema-to-typescript` already exists, I’d rather not have this repo just be a wrapper around an existing, actively-maintained project. I believe the experience for users would be worse than just using that, and I’d rather not introduce that dependency to the project.

Instead I’d recommend users use that package instead.

LironHazan:
@drwpow thanks for replying, I totally understand your decision!

PR #637, opened 2021-06-05
571 dtsgenerator

LironHazan:
Added an ability to parse and create types from json schema by using **dtsgenerator**


LironHazan:
another alternative for --> https://github.com/drwpow/openapi-typescript/issues/571

drwpow:
I’m going to close this PR because a project goal is not to ship `dtsgenerator` inside this package.

Perhaps in the future, if we all decide that `dtsgenerator` does everything this library does, and then some, then I’m happy to recommend people switch. But until then, my advice is to use `dtsgenerator` for your needs if it fits! At this time I’m not interested in shipping an overlapping library as a core dependency of this package.

PR #643, opened 2021-06-07
fix: properties + anyOf with only required properties

### LINKED ISSUE #642: Empty `Partial<{ [key: string]: any }>` in case of `schema.anyOf` with only `required` keys
I think the best will be to just filter out `anyOf` items that only have `required` keys. I'll try a PR to explore the effect of such a change for further discussion, ok?

This is currently causing TypeScript problems in @octokit, see https://github.com/octokit/types.ts/issues/319#issuecomment-856305469

gr2m:
fixes #642

### DIFF
--- a/src/transform/schema.ts
+++ b/src/transform/schema.ts
@@ -39,7 +39,19 @@ export function transformSchemaObjMap(obj: Record<string, any>, options: Transfo
 
 /** transform anyOf */
 export function transformAnyOf(anyOf: any, options: TransformSchemaObjOptions): string {
-  return tsIntersectionOf(anyOf.map((s: any) => tsPartial(transformSchemaObj(s, options))));
+  // filter out anyOf keys that only have a `required` key. #642
+  const schemas = anyOf.filter((s: any) => {
+    if (Object.keys(s).length > 1) return true;
+
+    if (s.required) return false;
+
+    return true;
+  });
+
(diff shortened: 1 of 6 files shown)


gr2m:
looks like the `14.x` CI fails due to some weird caching of a previous `prettier` version?

gr2m:
I've published `@gr2m/openapi-typescript@4.0.0-fork.2` and use it in @octokit for the time being, feel free to give a try if you have the same problem

drwpow:
> looks like the `14.x` CI fails due to some weird caching of a previous `prettier` version?
> 
> image

Hm. Maybe try rebasing off latest `main` and repushing? Maybe that will clear the cache

drwpow:
Actually I can try that real quick myself; forgot that you have access and this was a local branch in the project 🙂

gr2m:
doesn't look like it helped? Super odd, I don't know what's up sorry :(

drwpow:
Oh I figured it out—the snapshots were just updated with an older local version of Prettier. No big deal—easy fix. No code changes necessary.

drwpow (approved):

gr2m:
> the snapshots were just updated with an older local version of Prettier

🤦🏼 sorry for that

gr2m:
> the snapshots were just updated with an older local version of Prettier

🤦🏼 sorry for that

PR #649, opened 2021-06-09
Add an explicit description of reason why remote globbing doesn't work

taras:
## Motivation

I came across `for obvious reasons` but then I realized that it wasn't very obvious to me. I wanted to make sure it's easier for others to understand in the future.

## Approach

I replaced it with the reason why I believe this might be the reason, but I could be wrong. If I'm wrong, please suggest the actual reason so I can update it.


drwpow (approved):
Thank you for the addition! I will always approve PRs that make language more inclusive and explicit. I try to avoid “if you know you know“ insider language; sorry that it came across this way.

taras:
@drwpow all good. I'm glad this was helpful.

PR #678, opened 2021-07-01
refactor: simplify fs promise import

dnalborczyk:
the original PR imported from `fs/promises`, but unfortunately this path was not back-ported to node.js v12 😞


drwpow:
Thanks for raising this, but does this improve anything? I personally prefer the `fs` namespace as it’s predictable and easily-referenceable. The original code is intentional.

dnalborczyk:
> Thanks for raising this, but does this improve anything? I personally prefer the `fs` namespace as it’s predictable and easily-referenceable. The original code is intentional.

just less clutter. the original commit is not supported in node.js v12: https://github.com/drwpow/openapi-typescript/pull/678/commits/ea05226ce235427d6b0dc8a3967e66cfd8032f9a

Issue #680, opened 2021-07-01
v4 creates empty "interface external"


dnalborczyk:
issue: if we don't have an external/remote schema, **openapi-typescript** generates an `empty interface`.

```typescript
export interface external {}
```

which generally speaking is not a problem, but the eslint typescript rule [@typescript-eslint/no-empty-interface](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-empty-interface.md) complains (for the right reasons):

```cli
An empty interface is equivalent to `{}`  @typescript-eslint/no-empty-interface
```

drwpow:
Thanks for raising. But the empty interfaces are intentional, even though they do throw a lint error. I think it’s a better experience to consistently export the exact same wrappers no matter the schema, and only have the properties vary. I’d like to keep doing it this way.

dnalborczyk:
@drwpow could we instead put that behavior behind a flag? since this file is generated, it's not simple to ignore eslint/typescript in that case.

I also had another issue, potentially a breaking change, also with linting. type (aliases) and interfaces are usually [almost?] always uppercased, whereas `operations`, `external` etc is not. would you also accept a PR for this? with a flag as well?

drwpow:
> curious what the use case of exporting an empty interface would be?

Imagine if you were generating a ton of schemas programmatically, and also importing several at once. Isn’t it easier to simply count on the fact that `import { external } from 'my-schema.ts'` always works, no matter the contents? Sure, people can adapt, but I find that consistency helpful—that you won’t get an error just because a schema does or does not have external references (which could always change as the schema changes).

> since this file is generated, it's not simple to ignore eslint/typescript in that case.

Could you place an `.eslintrc` file only in your generated schema folder, and only disable the rules you need to? I find that to be helpful for generated files, because there are a lot of lint rules we just can’t realistically generate for.

> type (aliases) and interfaces are usually [almost?] always uppercased, whereas `operations`, `external` etc is not. would you also accept a PR for this? with a flag as well?

This fits into [Project Goal #2](https://github.com/drwpow/openapi-typescript#-project-goals) as outlined in the README: matching your schema. We want capitalization to perfectly match your schema, because OpenAPI supports more characters than TypeScript does. I agree that TypeScript usually prefers `PascalCase` but I’d personally prefer to mirror your schema.

[...]

PR #681, opened 2021-07-01
fix: only generate "external interface" if we have properties

dnalborczyk:
fixes: https://github.com/drwpow/openapi-typescript/issues/680


drwpow:
Thanks for raising, but as I said in #680 this is intentional to have empty interfaces, because it makes exports consistent across all schemas. I’d like to keep the empty interfaces.

Issue #645, opened 2021-06-08
Pass full header value with --auth arg


mbelsky:
Hey,

Thanks for this solution to make types from oapi jsons.

As I see this tool already supports `--auth` option and this arg expects to get a token for the `Authorization` header. However in my case I need to pass a header like `Cookie: auth=xxx` to get a private schema. 

What is your opinion about changing the `--auth` argument  logic from handling just a value for the `Authorization` header to handle full header value? This implementation will be more flexible for users, but this will be a breaking change for existing users.

The current syntax: `npx openapi-typescript --auth "Basic YWxhZGRpbjpvcGVuc2VzYW1l" https://oapi.com/secret-schema.json`
The suggested syntax: `npx openapi-typescript --auth "Authorization: Basic YWxhZGRpbjpvcGVuc2VzYW1l" https://oapi.com/secret-schema.json`

If you are open to make this change, I will be glad to help with this logic implementation.

drwpow:
Hm. I think in that case I’d probably rather add a `--header` flag with a `curl`-like syntax, like so: `--header "Authorization: …" --header "Content-Type: application-json`, etc. That way you just have full control over the headers. And we can keep `--auth` just as a shortcut.

Would that work?

mbelsky:
Thank you for sharing this solutions. It's a great alternative.

Just one more question, for a command like `npx openapi-typescript schema.yaml --auth "token123" --header "Authorization: token789"` should the tool use `token123` or `token789` as an auth token?

In my opinion it should use the `token789` token because it comes from a `--header` flag.

ericzorn93:
Added in this [`PR`](https://github.com/drwpow/openapi-typescript/pull/713)

PR #713, opened 2021-08-06
Feature:  Add Custom HTTP Headers and Method (via CLI and Programmatically)

ericzorn93:
This feature allows the user to send custom HTTP headers to the JS API in order to introspect and parse the remote URL Swagger/OpenAPI schema. This has been updated for the CLI format as well. These changes will not cause any breaking changes to the API and/or user.

All tests are passing.


ericzorn93:
Many of these tests seem to be failing from a combination of failed network calls to introspect the schema and some indentation issues with type output in the format of strings.

ericzorn93:
Hey @drwpow, it won't let me add reviewers to this PR for some reason.

drwpow:
> Hey @drwpow, it won't let me add reviewers to this PR for some reason.

I believe you have to have repo access to do that. But feel free to just @- mention anyone, or I can add them myself.

tests/empty-definitions.test.ts
```diff
@@ -30,62 +30,8 @@ describe("allow empty definitions", () => {
...
-      /** Invalid input */
-      readonly 405: unknown;
-    };
-  };
-}
-
-export interface external {}
-`);
+    // We are using snapshots to enforce consistency in string output and indentation
+    expect(await openapiTS(schema as any, { version: 2 })).toMatchSnapshot();
```
drwpow, line 34: 👍🏻 I’m fine with snapshots here. As you may have guessed I’m partial to having the full-file “snapshots” just so they could be more-easily readable. Also in a project like this where nearly everything is a snapshot, I didn’t want a careless `-u` flag causing a huge regression and liked the added friction of having to edit a file.

But that’s for the full-schema tests; I agree that the inline snapshots here were noisy, and not as useful. I’m OK with this change.

drwpow:
@ericzorn93 this is great! Very well-thought out, great execution, and thanks for improving the tests, too.

I’m happy to approve & merge this but let me know if you wanted additional reviewers as well.

gr2m:
I'm sorry I don't have the bandwidth to review. But we gotta trust the tests. If something breaks, that's the fault of the test setup, not your PR :) Looks like great work, thank you!

ericzorn93:
@gr2m Thank you. If we are good, do you or another maintainer mind merging?

ericzorn93:
@drwpow Sorry to bother you again, but if this PR is safe to merge, do you mind merging to master? I think this would be a great help for my personal use case and others as well 👍

bin/cli.js
```diff
@@ -31,6 +33,15 @@ Options
       auth: {
         type: "string",
       },
+      httpHeaders: {
+        type: "string",
```
mbelsky, line 37: What do you think about passing header values as an array like curl does that? I guess it will be easier to write headers one by one in cli instead of making one json.

How curl does that: `curl -header "X-First-Name: John" --header "X-Last-Name: Doe" http://example.com/`

Also probably with this way we won't need to parse headers because `meow` could handle multiple values for a flag: https://github.com/sindresorhus/meow#flags

src/utils.ts
```diff
@@ -139,3 +139,17 @@ export function tsUnionOf(types: Array<string | number | boolean>): string {
...
+/**
+ * Determines if the http method being evaluated
+ * is permitted for the fetch request to the server
+ * to load the schema.
+ *
+ * @export
+ * @param {string} httpMethod
+ * @return {boolean} isValid
+ */
+export function isValidHTTPMethod(httpMethod: string | HTTPVerb): boolean {
```
mbelsky, line 152: I found this function returns true for values like `123POST, POST123` etc. What about using [http.METHODS](https://nodejs.org/api/http.html#http_http_methods) to validate a value?

```suggestion
import {METHODS} from 'http'

export function isValidHTTPMethod(httpMethod: string | HTTPVerb): boolean {
  METHODS.includes(httpMethod.toUpperCase())
}
}
```

src/load.ts
```diff
@@ -89,9 +140,28 @@ export default async function load(
...
       headers.set("User-Agent", "openapi-typescript");
-      if (options.auth) headers.set("Authorization", options.auth);
-      const res = await fetch(schemaID, { method: "GET", headers });
+      if (options.auth) {
+        headers.set("Authorization", options.auth);
+      }
+
+      // Fetch Swagger Schema by schemaID in cache/URL via GET request
+      let httpMethod: HTTPVerb = "GET";
+      if (options.httpMethod && typeof options.httpMethod === "string" && isValidHTTPMethod(httpMethod)) {
```
mbelsky, line 161: `typeof undefined !== "string "` so let's simplify this condition a bit:
```suggestion
      if (typeof options.httpMethod === "string" && isValidHTTPMethod(httpMethod)) {
```

src/load.ts
```diff
@@ -89,9 +140,28 @@ export default async function load(
     } else {
       // load remote
       const headers = new Headers();
+
+      // Add custom parsed HTTP headers
+      if (options.httpHeaders) {
+        const parsedHeaders = parseHttpHeaders(options.httpHeaders);
+        Object.keys(parsedHeaders).forEach((headerKey) => {
```
mbelsky, line 147: What about using `Object.entries` here instead of `Object.keys`?

mbelsky (changes requested):
This is a great feature that I'd really would like to have in this tool.

src/load.ts
ericzorn93: I know that we can access the value using the entries method, but is there a preference of why we would do this?￼

src/load.ts
ericzorn93: Fair enough, this was an oversight on my end, thank you for pointing this out.￼

bin/cli.js
ericzorn93: Is this a block or to getting merged? Or can we add this at a later point?￼￼

src/utils.ts
ericzorn93: Great point! I will fix this as well￼

bin/cli.js
ericzorn93: I definitely feel that this is a very useful feature and it has a way cleaner API, but can this be an additional option? I think both ways would be valid ways of inputting HTTP headers￼￼

ericzorn93: @mbelsky I just realized a unique edge case on why this might not be the best API for the user. Currently, the values are the only thing stored in the array. Therefore, we would have to enforce a strict string schema of ‘-x "headerKey:headerValue‘ and then split on the string. However, compared to a JSON object (yes, definitely more verbose), they will potentially run into issues with spacing and casing. We would need much more strict validation

src/load.ts
ericzorn93: Complete

src/load.ts
ericzorn93: Completed

bin/cli.js
ericzorn93: @drwpow What do you think about this implementation? I personally think that we should be okay to merge and then think about this additional feature and implementation next.

mbelsky: We have [discussed](https://github.com/drwpow/openapi-typescript/issues/645#issuecomment-868938899) with @drwpow about this feature. Just wanna to link the issue to this pr.

src/load.ts
mbelsky: There is only one preference why I've asked about this: we rarely use the classic `for (let i=0; i<…)` statement to iterate a list without making any index magic because we have better version of that: for in, for of, forEach. These modern versions are less verbose and more semantically clear. So same point for `Object.entries`.

bin/cli.js
ericzorn93: I can add this as an additional alternative￼

.nvmrc
```diff
@@ -0,0 +1 @@
+v16
```
gr2m, line 1: This change seems out of scope of this pull request

gr2m:
Once you two are in agreement and confirmed that it's working, can you please ping Drew? It's up to him whether the added complexity is something he wants to maintain in the long term. You have my blessing, but I'm just helping out :)

gr2m:
sounds like you'll have to bump the coverage ☝🏼 it shouldn't drop

.nvmrc
ericzorn93: Np, this can be removed. I was trying to enforce consistency but this should not be an issue for most people using Node 14>=

bin/cli.js
ericzorn93: @mbelsky and @drwpow We now support both methods. Please check the updated PR.

.nvmrc
ericzorn93: Removed :)

ericzorn93:
@gr2m I have added a number of small tests for utility functions to attempt to bump coverage. I don't know how much more I can really increase unit test coverage of the load function without adding insignificant granular unit tests

ericzorn93:
@gr2m all test coverage has been updated

ericzorn93:
- [x] Hey @gr2m, @mbelsky, and @drwpow. Sorry to be pushy, but my PR should be good to go with all requested changes and increased test coverage. Now, you can add headers individually like the CURL CLI, or via a JSON string. 

Do you mind merging?

gr2m:
> This functionality would be great in my current project 🎉

To take the pressure of, maybe publish your own fork for the time being? I don't know when @drwpow will have the time to review, and it's bad etiquette in independent Open Source projects to pressure maintainers like that, please don't do that.

ericzorn93:
@gr2m I completely understand and did not think of that tbh. I will do this in the meantime. I apologize for any pressure I have put on the maintainers

drwpow:
@ericzorn93 sorry for the delay on this! This library has been difficult for me to maintain the past few months. I went ahead and approved and merged this in #764, with a few changes as-noted. See that PR for more details.

This approach of merging your work in another PR is a bit unorthodox, and I apologize. Please check that off as a failure on my part to give you a timely review! If you’d like to adjust anything from my revisions, please open up another PR.

Overall I wanted to say again **what a quality PR this was!** Your JS Docs are excellent, your comments are excellent, and I am amazed at the amount of thought & care you put into this PR. And I apologize again for not merging this in sooner because it is quality work.

I’ll be shipping a new minor version with this feature soon!

drwpow:
@all-contributors please add @ericzorn93 for code, test, docs

ericzorn93:
@drwpow No worries at all. I completely understand. I apologize for coming across pushy, I was just trying to incorporate this on a script that I have been working on. 

I completely understand that maintaining widely used open source software can be a difficult task and very time consuming. This is a great repo!

Thanks for the kind words and I look forward to contributing more in the future 😀.

Thanks,
Eric

Issue #718, opened 2021-08-11
Use formatter before node processing


ar53n:
Hello, I have a suggestion to apply a formatter before defining nullable for flexible schema editing. 

https://github.com/drwpow/openapi-typescript/blob/db04503309ee1e7e7f6b8c8cd97d642ec1216dfa/src/transform/schema.ts#L82-L88

ar53n:
#720

PR #720, opened 2021-08-11
Fix formatter position

ar53n:
Change formatter position before string builder
#718


drwpow (approved):
Sure thing! I think this is fine to move here.

PR #747, opened 2021-09-16
feat: add @deprecated comment on schema objects

bunkscene:
This adds support for deprecated schema objects.

```
"components": {
    "schemas": {
      "Table": {
        "type": "object",
        "properties": {
          "Description": {
            "type": "string",
            "description": "The description field.",
            "deprecated": true
```

### DIFF
--- a/src/transform/schema.ts
+++ b/src/transform/schema.ts
@@ -19,7 +19,10 @@ export function transformSchemaObjMap(obj: Record<string, any>, options: Transfo
     const v = obj[k];
 
     // 1. JSDoc comment (goes above property)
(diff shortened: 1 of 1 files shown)


bunkscene:
Hey @drwpow, I added this PR to help support deprecated schema objects.  I used the same logic you use elsewhere in the repo.  Thoughts on pulling this in?

drwpow:
@bunkscene This is great! Definitely in-favor of merging this.

Could you add a quick test just to make sure output is correct? There’s a comment test in `schema.test.js` and I think you could add a simple test right beneath that.

PR #763, opened 2021-09-30
Chore: update deps

drwpow:
Update deps

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -1,7 +1,7 @@
+import fetch from "node-fetch";
 import fs from "fs";
 import path from "path";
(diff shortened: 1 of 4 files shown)


src/load.ts
```diff
@@ -88,10 +88,13 @@ export default async function load(
       contentType = mime.getType(schemaID) || "";
     } else {
       // load remote
-      const headers = new Headers();
```
drwpow, line 91: Minor change: ditching a library-generated `Headers()` class not shipped by core Node just makes TypeScript a bit happier. Replaced with just a standard object which works just as well.

PR #764, opened 2021-09-30
Feature add custom http headers

drwpow:
Supercedes https://github.com/drwpow/openapi-typescript/pull/713.

## Changes

Other than the commented code below, the following changes were made from #713 

- The `load` unit test was removed. Reason being that didn’t give much assurance that the entire CLI was behaving as desired (in my limited testing, it wasn’t)
- The `util.test.ts` unit test was also removed. This was a little too implementation detail-y, and it only retested what existed in other tests.


bin/cli.js
```diff
@@ -69,6 +86,22 @@ function errorAndExit(errorMessage) {
 async function generateSchema(pathToSpec) {
   const output = cli.flags.output ? OUTPUT_FILE : OUTPUT_STDOUT; // FILE or STDOUT
 
+  // Parse incoming headers from CLI flags
+  let httpHeaders = {};
+  // prefer --headersObject if specified
+  if (cli.flags.headersObject) {
+    httpHeaders = JSON.parse(cli.flags.headersObject); // note: this will generate a recognizable error for the user to act on
```
drwpow, line 93: Minor change here: we had a `getCLISchemaHeadersJSON()` function that did a lot of setup around `JSON.parse()`, and in the end obscured the error from the user. Instead, I’d like to expose the error to the user so it’s more clear what’s not parsing correctly.

bin/cli.js
```diff
@@ -69,6 +86,22 @@ function errorAndExit(errorMessage) {
...
+  }
+  // otherwise, parse --header
+  else if (Array.isArray(cli.flags.header)) {
+    cli.flags.header.forEach((header) => {
+      const firstColon = header.indexOf(":");
```
drwpow, line 98: This was using a RegEx instead (`/(\w+)(\:)\s((\w+)[\-|\:]?.+)/gi;`), but in testing I found it to be brittle and a little awkward. Instead I opted for more of a `curl`-like approach where the first `:` designates the key–value relationship. It’s then up to the user to format their header correctly.

src/types.ts
```diff
@@ -134,6 +136,21 @@ export interface SwaggerToTSOptions {
   silent?: boolean;
   /** (optional) OpenAPI version. Must be present if parsing raw schema */
   version?: number;
+  /**
+   * (optional) List of HTTP headers that will be sent with the fetch request to a remote schema. This is
+   * in addition to the authorization header. In some cases, servers require headers such as Accept: application/json
+   * or Accept: text/yaml to be sent in order to figure out how to properly fetch the OpenAPI/Swagger document as code.
+   * These headers will only be sent in the case that the schema URL protocol is of type http or https.
+   */
+  httpHeaders?: Headers;
```
drwpow, line 145: This is an opinionated change, but in the previous PR we allowed `httpHeaders?:` to be a polymorphic type, from `null` to a JS object to a `Map`. We then had to implement quite a bit of runtime logic to check to make sure everything lined up correctly.

Instead, I opted to go with just a plain JS object for headers because that’s what gets passed to `fetch()` anyway. Servers are lax in what they accept. And while I completely get the author’s original desire to strongly type their request headers, I believe that’s outside the scope of this library, and could only introduce possible bugs/unnecessary maintenance.

src/types.ts
```diff
@@ -134,6 +136,21 @@ export interface SwaggerToTSOptions {
...
+  /**
+   * HTTP verb used to fetch the schema from a remote server. This is only applied
+   * when the schema is a string and has the http or https protocol present. By default,
+   * the request will use the HTTP GET method to fetch the schema from the server.
+   *
+   * @default {string} GET
+   */
+  httpMethod?: string;
```
drwpow, line 153: Similar approach: rather than maintain a discriminated union of all valid HTTP methods, I opted instead to allow a simple `string` like most every other library allows. Perhaps your server allows more!

Issue #726, opened 2021-08-18
`TypeError: boolean true is not iterable (cannot read property Symbol(Symbol.iterator))` at 4.0.2


Charlie91:
I have a problem on a last `4.0.2` version that I solved rolling back to `4.0.0`.

`TypeError: boolean true is not iterable (cannot read property Symbol(Symbol.iterator))`.

The error also dissapears if I change property `required` in parameter of `"type": "file"` to `false`

swagger.json
```
{
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
  ],
  "swagger": "2.0",
  "info": {
    "description": "API for Portal",
    "title": "Portal API",
    "version": "0.1.0"
  },
  "host": "portal.test",
  "basePath": "/api",
  "paths": {
    "/activity/transaction/{counterparty}/{transaction_number}/attachment": {
      "post": {
        "security": [
          {
            "Oauth": [
              "readaccount"
            ]
          }
        ],
        "consumes": [
          "multipart/form-data"
        ],
        "produces": [
          "application/json"
        ],
        "tags": [
          "activity_transaction_attachment"
        ],
        "summary": "Upload attachment for the transaction",
        "operationId": "postTransactionAttachment",
        "parameters": [
          {
            "type": "file",
            "description": "The file to upload",
            "name": "file",
            "required": true,
            "in": "formData"
          }
        ],
        "responses": {
          "200": {
            "description": "Success upload an attachment for the instruction"
          },
          "default": {
            "description": "Default error",
            "schema": {
              "type": "object",
              "$ref": "#/definitions/ErrorResponse"
            }
          }
        }
      }
    }
  },
  "definitions": {},
  "securityDefinitions": {}
}
```

eirikur-grid:
I just ran into this error as well. Will give 4.0.0 a try.

drwpow:
Hm. I don’t think we know how to support `"type": "file"`. Let me push a fix for that and see if that resolves it.

drwpow:
I’m seeing an improvement to this in `4.1.0`. Please give it a try and let me know if the problem is improved.

mmounts:
Not the OP here, but I was hitting this issue in `4.0.2` today, found this issue while researching, and update to `4.1.0`. For my use case the error has now gone away and the result is as expected. Thanks!

mitchell-merry:
I believe this is fixed in version 6 as well, closing...

PR #766, opened 2021-09-30
Feat: add support for "type": "file" in parameters

drwpow:
Related to https://github.com/drwpow/openapi-typescript/issues/726. When `type: "file"` is encountered in parameters, weird results occur.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -32,7 +32,17 @@ export function isRef(obj: any): obj is ReferenceObject {
 }
 
 /** Return type of node (works for v2 or v3, as there are no conflicting types) */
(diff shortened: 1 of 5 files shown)


src/utils.ts
```diff
@@ -67,6 +77,11 @@ export function nodeType(obj: any): SchemaObjectType | undefined {
     return "array";
   }
 
+  // file
+  if (obj.type === "file") {
+    return "unknown";
```
drwpow, line 82: At least for now, we’ll return `unknown` for a file type. Because we’re not sure where this is running—server or browser—it would seem risky to simply assume a `Buffer` or `string` type for a file. Instead, we’ll rely on the implementer to specify how to deal with the file.

PR #769, opened 2021-09-30
Improve generated types for unknown schema objects

### LINKED ISSUE #693: Schema without a type should be "any" or "unknown" instead of a generic map type
According to the [OpenAPI spec](https://swagger.io/docs/specification/data-models/data-types/#any), when a schema is defined without a type it is the equivalent of:

```
      anyOf:
        - type: string
        - type: number
        - type: integer
        - type: boolean
        - type: array
          items: {}
        - type: object
```

but openapi-typescript uses [`object` type as the default](https://github.com/drwpow/openapi-typescript/blob/main/src/types.ts#L102) and [generates](https://github.com/drwpow/openapi-typescript/blob/main/src/transform/schema.ts#L116-L121) `{ [key: string]: unknown }`. Instead it should generate just `unknown`.

drwpow:
Fixes #693, #554

## Background

According to #693, I believe this is correct:

>According to the [OpenAPI spec](https://swagger.io/docs/specification/data-models/data-types/#any), when a schema is defined without a type it is the equivalent of:
>
>```
>      anyOf:
>        - type: string
>        - type: number
>        - type: integer
>        - type: boolean
>        - type: array
>          items: {}
>        - type: object
>```
>
>but openapi-typescript uses [`object` type as the default](https://github.com/drwpow/openapi-typescript/blob/main/src/types.ts#L102) and [generates](https://github.com/drwpow/openapi-typescript/blob/main/src/transform/schema.ts#L116-L121) `{ [key: string]: unknown }`. Instead it should generate just `unknown`.

This PR addresses that, fixing some bugs in the process. And also switches from `any` to `unknown` in other areas (#554).

I had originally thought that changing to `unknown` would be disruptive for people, but after looking again at the output I don’t believe it will be too bad for most. It should actually surface errors that people weren’t aware of before, which is the whole goal! So this will be released as a **minor version** rather than a breaking major version.

### DIFF
--- a/tests/v2/expected/stripe.ts
+++ b/tests/v2/expected/stripe.ts
@@ -8991,7 +8991,7 @@ export interface operations {
     parameters: {
       query: {
         /** Specifies which fields in the response should be expanded. */
-        expand?: any[];
+        expand?: unknown[];
       };
       path: {
         three_d_secure: string;
@@ -9013,7 +9013,7 @@ export interface operations {
     parameters: {
       query: {
         /** Specifies which fields in the response should be expanded. */
-        expand?: any[];
+        expand?: unknown[];
       };
     };
     responses: {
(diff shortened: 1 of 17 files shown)


tests/v2/expected/stripe.ts
```diff
@@ -9151,7 +9151,7 @@ export interface operations {
               state?: string;
               town?: string;
             };
-            dob?: { [key: string]: unknown };
```
drwpow, line 9154: This is a big improvement! `unknown` should be the correct type here.

PR #797, opened 2021-11-05
feat: new jsdoc comments formatter

sadfsdfdsa:
Feat more powerful comments formatter with jsdoc tags supports.

**TYPE: FEATURE**
**AFFECTS: ONLY COMMENTS**

Now supports:

- `title` as simple description comment (as description in actual version)
- `format` from yaml with prefix (example: `Format: date-time`), think it's better to have it)
- `@deprecated` as in previous version
- `@description something` as JSDOC tag (actual version: without tag, but now title use that format and description switched to tag)
- `@default something` as JSDOC tag
- `@example something` as JSDOC tag

Generated example with full possibilities of new version:
image

Also added json (ping me if need to go to yaml) with all new jsdoc config example.

Of course, I broke tests that are designed for the old description format, thinks it's better to change expects with new generated. Snapshots different only in comments body.


sadfsdfdsa:
@drwpow sorry for pushing, but we need to merge this pr, or we'll only have to use the fork version of your package (

drwpow (approved):
This is fantastic! Thank you for adding.

Also, sorry for the delay on this. I kinda was maxxed out on open-source maintenance for a while this year.

drwpow:
I won’t push a new release today, as I want to do a little local testing. But this will go out in a minor release this week

sadfsdfdsa:
No problem, thanks for reaction and merging, hope work with this package in future as well!

drwpow:
@all-contributors please add @sadfsdfdsa for code

drwpow:
Released in `4.5.0`!

PR #854, opened 2022-01-03
Update types.ts to allow "produces" and "consumes" in Operation Object per OpenAPI v2 spec.

btelles:
Allow "produces" and "consumes" mime types that are defined in the v2 spec: https://swagger.io/specification/v2/#operation-object

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -58,6 +58,8 @@ export interface OperationObject {
   description?: string;
   tags?: string[]; // unused
   summary?: string; // unused
+  consumes?: string; // unused
+  produces?: string; // unused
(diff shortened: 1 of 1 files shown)


drwpow:
Thanks very much for contributing! However, the types inherent to this library are more “functional” types and not really meant to be consumed, and by no means are complete. For people that need complete types, I’d recommend the [openapi-types](https://www.npmjs.com/package/openapi-types) package.

I’d probably be against merging this because it doesn’t really add a feature or fix a bug, unless the added types are something that you’re looking into generating? (in which case, I’d love an issue first to discuss how these may be added)

Issue #869, opened 2022-02-01
Feature request: Input yaml from stdin


kgtkr:
Example:
```
# Input from stdin
$ openapi-typescript -
```

Currently `openapi-typescript /dev/stdin` doesn't work.

```
file:///home/kgtkr/path/to/node_modules/openapi-typescript/bin/cli.js:38
  throw new Error(errorMessage);
        ^

Error: ❌ Could not find any specs matching "/dev/stdin". Please check that the path is correct.
    at errorAndExit (file:///home/kgtkr/path/to/node_modules/openapi-typescript/bin/cli.js:38:9)
    at main (file:///home/kgtkr/path/to/node_modules/openapi-typescript/bin/cli.js:149:5)
```

PR #870, opened 2022-02-01
schema input from stdin

kgtkr:
fix: #869


drwpow (changes requested):
Thank you! I’m in favor of this change and would like to see this merged. However, before we do, could you please add a test to `cli.test.js`?

There is a test for `stdout` already; please add one there for `stdin` so we can be sure there are no regressions with this in the future.

kgtkr:
@drwpow
Thank you for your review. I added a test.

Richard87:
I would love this feature as well! I saw the readline-sync library on npmjs.org with 760'000 weekly downloads, but the logic seemed pretty complicated to get right in all environments, maybe pull in that library if you want stdin support for windows (and powershell +++) instead of managing it yourself? 

https://www.npmjs.com/package/readline-sync
https://github.com/anseki/readline-sync/blob/master/lib/readline-sync.js

kgtkr:
Sorry I'm late. I was able to fix it.
But in CI, `pnpm install` fails in one place. I think it will be sucuceed if you re-execute.

drwpow:
Thanks for fixing! 🙏 Re-running the test now to see if it’s just flaky (and will fix in a followup)

drwpow (approved):

PR #872, opened 2022-02-01
Supports minItems / maxItems

### LINKED ISSUE #871: Supports minItems / maxItems
example:
```ts
// minItems=undefined, maxItems=undefined
type A = X[];

// minItems=1, maxItems=undefined
type B = [X, ...X[]];

// minItems=1, maxItems=3;
type C = [X] | [X, X] | [X, X, X];

// minItems=undefined, maxItems=2;
type D = [] | [X] | [X, X];
```

* Have to be careful about code sizes that are too large
* The generated code is incompatible and should add an option
* openapi-typescript can put an array in items, but not in openapi3.0

kgtkr:
fix #871


test/core/schema.test.js
```diff
@@ -150,6 +151,34 @@ describe("SchemaObject", () => {
       );
     });
 
+    it("array (supportArrayLength)", () => {
```
drwpow, line 154: Awesome tests!

drwpow (approved):
This is great, thank you!

drwpow (changes requested):
I’m in favor of merging this, but on second thought, I’d like to see the following added:

- There’s no documentation in the README, nor the `--help` flag on how to use this new feature. How would a user use this? When would they want to use this?
- There are no tests in the CLI to test if this flag is working (there are good TypeScript tests, but we’re not checking the `--support-array-length` flag is passed correctly)

Could you address both of the following?

kgtkr:
Thanks. I added a test. 
The difference between with and without the option is, for example, `test/v3/expected/github.support-array-length.ts:16633`.

PR #895, opened 2022-04-07
Fix anyOf transformation

### LINKED ISSUE #894: anyOf should not result in intersection type
`anyOf` is equivalent to `OR`, i.e. TypeScript's union `|` operator.

The current behavior is incorrect.

---

Related Note:

`oneOf` is equivalent to `XOR`, no direct equivalent for TypeScript. Currently treated as `OR`.

MunifTanjim:
Fixes #894


mull:
Anything I can do to help move this along?

MunifTanjim:
> Anything I can do to help move this along?

Additional tests are added and previous tests were updated. So I guess we just need approval from one of the maintainers. 🤞🏼

test/core/schema.test.js
```diff
@@ -304,7 +304,7 @@ describe("SchemaObject", () => {
           { ...defaults }
         )
       ).to.equal(
-        `(Partial<components["schemas"]["StringType"]>) & (Partial<components["schemas"]["NumberType"]>) & (Partial<components["schemas"]["BooleanType"]>)`
```
drwpow, line 307: 👍 this seems more correct to me

drwpow: Bah actually, I re-read the spec and I’m second-guessing it now. See other comment.

drwpow:
Hm after revisiting the spec, I’m not so sure this is an improvement on the existing behavior. Copying a comment [I made here](https://github.com/drwpow/openapi-typescript/issues/894#issuecomment-1107702682):

Hm, from the [OpenAPI specification](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/):

> OpenAPI 3.0 provides several keywords which you can use to combine schemas. You can use these keywords to create a complex schema, or validate a value against multiple criteria.
>
>    - [`oneOf`](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#oneof) – validates the value against exactly one of the subschemas
>    - [`allOf`](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#allof) – validates the value against all the subschemas
>    - [`anyOf`](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#anyof) – validates the value against any (one or more) of the subschemas

I don’t believe `anyOf` is `OR`, as it may be either **one** of the subschemas, or **all** of the subschemas. To get `OR` you would want `oneOf`, no?

I think the current behavior is correct (`Partial<A> & Partial<B> & Partial<C>`) unless I’m missing something.

MunifTanjim:
Replied in https://github.com/drwpow/openapi-typescript/issues/894#issuecomment-1107714019 @drwpow

drwpow:
So as discussed in #894, I’m in favor of this change and I’ve been convinced this is proper behavior. Even though this is a _fix_, this may disrupt enough people I’m going to release this in a breaking change.

There are some other minor bugfixes I’ll merge first, then I’ll ship a prerelease candidate for the next major version with this for testing. If that looks good, we’ll cut a new major for this 😎

MunifTanjim:
Rebased on latest `main`.

drwpow:
Just released a new version at `openapi-typescript@next` that has this behavior. If anyone has time to test this out in their existing schemas and see if it works as intended it’d be much appreciated 🙏. I’ll do the same in mine.

If the behavior works as expected, I’ll cut a stable release of this.

PR #905, opened 2022-04-26
feat(flags): add flag to force never response body

duncanbeevers:
The OpenAPI specs state that in order to describe an Empty Response Body, the response `content` section should be omitted.
https://swagger.io/docs/specification/describing-responses/

The current behavior generates the expected `never` type for a handful of status code which should always return an empty response body.
However, when an app returns an empty response body for a status code outside of this short list, we lose type safety.

This flag allows developers to generate types with the `never` type used for every response with no `content`


bin/cli.js
```diff
@@ -21,6 +21,7 @@ Options
   --header, -x                 (optional) Provide an array of or singular headers as an alternative to a JSON object. Each header must follow the key: value pattern
   --httpMethod, -m             (optional) Provide the HTTP Verb/Method for fetching a schema from a remote URL
   --immutable-types, -it       (optional) Generates immutable types (readonly properties and readonly array)
+  --content-never              (optional) If supplied, an omitted reponse \`content\` property will be generated as \`never\` instead of \`unknown\`
```
drwpow, line 24: Nice! Thank you for extending the help message

test/bin/expected/content-never.ts
```diff
@@ -0,0 +1,21 @@
...
+ * This file was auto-generated by openapi-typescript.
+ * Do not make direct changes to the file.
+ */
+
+export interface paths {
+  "/test": {
+    get: {
+      responses: {
+        /** A list of types. */
+        200: never;
```
drwpow, line 11: 👍 great test!

drwpow (approved):
Looks great!

One of the tests is failing because of a simple copy/paste error (I think the comment should be `/** OK */` rather than `/** A list of types */`. Once that’s fixed, happy to merge.

PR #906, opened 2022-04-27
fix: Serialize falsey values

### LINKED ISSUE #890: Falsy example value do not generate comments in output
When we do the below to generate interface from schema, when `example` value is false, it does not generate the comment for the example value but when `example` value is true, it does. Could we have an option to allow this or even make it default? Thanks

Schema
```
openapi: 3.0.3
components:
  schemas:
    Config:
      type: object
      properties:
        IsEnable:
          type: boolean
          example: false // this is the value I am talking about
```

Expected output
```
export interface components {
  readonly schemas: {
    readonly Config: {
        /** @example false */
        readonly IsEnable: boolean
    }
  }
}
```

Actual output
```
export interface components {
  readonly schemas: {
    readonly Config: {
        readonly IsEnable: boolean // comment is not generated
    }
  }
}
```

duncanbeevers:
Closes #890

I went through and looked for all the spots where an empty string made sense and would be translated across to the TS files.
This includes `example`, `description`, `title`, `format`, and `default`.

I tried to fix this up for both v2 and v3 schemas.


src/transform/request.ts
```diff
@@ -6,7 +6,7 @@ export function transformRequestBodies(requestBodies: Record<string, RequestBody
   let output = "";
 
   for (const [name, requestBody] of Object.entries(requestBodies)) {
-    if (requestBody && requestBody.description) output += `  ${comment(requestBody.description)}`;
+    if (requestBody && requestBody.description !== undefined) output += `  ${comment(requestBody.description)}`;
```
drwpow, line 9: 👍 Great fix, thank you!

drwpow:
I’m in favor of this approach, but there are some test failures now where empty JSDoc comments are being inserted into the schema. I’d like to avoid this if possible.

But when that’s fixed, happy to approve & merge!

duncanbeevers:
Are you referring to the empty JSDoc comments like [this](https://github.com/drwpow/openapi-typescript/pull/906/files#diff-0936c19deb046277e4bd1f06f738f1a9a1ee617c53680fb56f6fb7f2f559b742R8)?

```ts
/** */
```
They definitely seem kind of weird, but I was going with the perspective "The developer put an empty string in here, for some reason…, so respect that."

https://youtu.be/OIVIDmuJZLI?t=40
u-g-F5F5P30

One would really have to go out of their way in order to introduce these empty comments.

Are you saying you'd like me to ignore descriptions and such that would generate such comments, or that you'd like me to fix the tests to match the PR behavior?

duncanbeevers:
I see that in the spec, for many fields like response-code definitions, that the `description` field is mandatory and can't be simply elided.

I agree that these comments are unsightly in this case when there's no mechanism available to avoid them, and I'll see about cleaning up those instances.

duncanbeevers:
I've re-worked this PR to only serialize falsey `example` and `default` values, and not all these other fields.
The PR depends-on, and builds-on https://github.com/drwpow/openapi-typescript/pull/909, although I was unable to change the base branch of the PR.

drwpow (approved):
Looks great, thank you!

drwpow:
@all-contributors please add @duncanbeevers for code, bug, test

Issue #915, opened 2022-06-06
Schema escape issues for $ref


liangskyli:
version: openapi-typescript@5.3.0
system: windws
node: v14.19.3

my openapi3 json file
```
{
  "openapi": "3.0.0",
  "info": {
    "title": "bff to openapi3",
    "version": "1.0.0"
  },
  "paths": {
    "/v1/getUser2": {
      "get": {
        "operationId": "UserController.getUser2.DjHhjSamaD",
        "tags": ["User"],
        "summary": "",
        "description": "UserController.getUser2",
        "parameters": [
          {
            "in": "query",
            "name": "param1",
            "description": "",
            "required": false,
            "schema": {
              "type": "number"
            }
          },
          {
            "in": "query",
            "name": "param2",
            "description": "",
            "required": false,
            "schema": {
              "type": "object",
              "oneOf": [
                {
                  "type": "string"
                },
                {
                  "$ref": "#/components/schemas/IUserWithId"
                }
              ]
            }
          },
          {
            "in": "query",
            "name": "param3",
            "description": "",
            "required": false,
            "schema": {
              "type": "object",
              "oneOf": [
                {
                  "type": "string"
                },
                {
                  "$ref": "#/components/schemas/IUserWithName"
                }
              ]
            }
          }
        ],
        "responses": {
          "200": {
            "content": {
              "application/json": {
                "schema": {}
              }
            },
            "description": "Successful response"
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "IUserWithId": {
        "oneOf": [
          {
            "$ref": "#/components/schemas/Partial<Pick<IUser,\"id\">>"
          },
          {
            "type": "string"
          }
        ]
      },
      "Partial<Pick<IUser,\"id\">>": {
        "type": "object",
        "properties": {
          "id": {
            "type": "number"
          }
        }
      },
      "IUserWithName": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string"
          }
        },
        "required": ["name"]
      },
      "IUser": {
        "type": "object",
        "properties": {
          "id": {
            "type": "number"
          },
          "name": {
            "type": "string"
          }
        },
        "required": ["id"]
      }
    }
  }
}

```
I use this openapi3 json file, generate ts file error

error info:
```
(node:9836) UnhandledPromiseRejectionWarning: SyntaxError: Property or signature expected. (17:1)
  15 |   schemas: {
[...]

PR #916, opened 2022-06-06
fix: schemas key double quote replace (#915)

liangskyli:
fix bug: #915 , use schemas key double quote replace


drwpow (approved):
Great fix, and great tests! No feedback; happy to merge this.

Issue #898, opened 2022-04-20
Support nullable as type arrays for OpenAPI 3.1


timomeh:
In OpenAPI v3.1 `nullable: true` was removed in favor of type arrays with the value `"null"`. Reference: https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

```yaml
# OpenAPI v3.0
type: string
nullable: true

# OpenAPI v3.1
type:
- "string"
- "null" 
```

The current version of openapi-typescript generates those properties as `unknown`, instead of `string | null`.

drwpow:
I think there are 2 things here worth addressing:

### 1. This feels like a good change that should happen for both 3.0 and 3.1

I think the `unknown` behavior is probably improved by a `string | null` union as you suggested. I _think_ I remember some previous changes to the `nullable` flag, but I don’t recall an argument against having `string | null` generated (but if I’m wrong or if someone finds a previous decision in the issues this ticket disagrees with, please remind me!).

### 2. This library should probably distinguish between 3.0 and 3.1

This doesn’t block the previous solution from happening, but it may be needed in the future.

<details>
  <summary>Extra info</summary>
  
  There is an [internal version number](https://github.com/drwpow/openapi-typescript/blob/main/src/index.ts#L46) that gets passed around for the parts of the codegen where Swagger 2.x differs from OpenAPI 3.x. But where OpenAPI 3.0 differs from 3.1, this library could do better.

I think this wouldn’t be a major change—probably just switching that internal `version` number to a discriminated union (e.g. `2 | 3 | 3.1` or perhaps a string equivalent). That should probably happen at some point.

_Sidenote: previous versions of this library did have more of a separation between 2.x and 3.x schemas, but all-in-all they share so much in common that it was easier to maintain one codebase and hard-code in the deviations rather than maintain 2 completely-isolated generators_
</details>

Anyways, would love if someone could open a PR to address this!

BrandonGoren:
@drwpow I have a PR open that I _think_ solves this issue by handling array types recursively: https://github.com/drwpow/openapi-typescript/pull/917

PR #917, opened 2022-06-06
Support Type Arrays for OpenAPI 3.1

BrandonGoren:
https://github.com/drwpow/openapi-typescript/issues/898

This PR allows arrays of types to be parsed properly, as per the 3.1 specification. 

## Changes
- Adds recursive functionality to parse "arrays" of types
- Supports the "null" type 

## Open questions
- I tested this locally using my company's openapi spec at https://developers.arcadia.com/openapi.yaml. I feel  a little weird about including this in a test because openapi-typescript does not support OpenAPI 3.1 - **How should I add a test for this?**

### DIFF
--- a/src/transform/schema.ts
+++ b/src/transform/schema.ts
@@ -105,8 +105,9 @@ export function transformSchemaObj(node: any, options: TransformSchemaObjOptions
   } else {
     // transform core type
     switch (nodeType(node)) {
-      case "multiple-types":
-        output += tsUnionOf((node.type as any[]).map(type => transformSchemaObj({ ...node, type }, options)))
+      case "type-array":
+        // This is an array of types as of the 3.1 specification - we should recursively evaluate them
+        output += tsUnionOf((node.type as any[]).map((type) => transformSchemaObj({ ...node, type }, options)));
         break;
       case "ref": {
         output += node.$ref; // these were transformed at load time when remote schemas were resolved; return as-is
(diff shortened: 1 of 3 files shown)


drwpow:
> I feel a little weird about including this in a test because openapi-typescript does not support OpenAPI 3.1

I feel that there are several outstanding tasks to get this library to support OpenAPI 3.1! Yes that is beyond the scope of this PR

> How should I add a test for this?

Simplified, minimal examples should go in `test/core/*.test.js`. Those tests are central to making sure each individual operation works as expected.

A goal of this project is to support _every_ schema, so for that reason we do test transforming entire schemas, but that is only done to make sure that all the individual operations work in tandem, and those are secondary (which, interestingly enough, is why this library was created in the first place—so many typegen libraries only tested against the PetStore schema but failed at almost every other real-world schema).

BrandonGoren:
@drwpow I added some tests to the "core" module - let me know what you think!

Newbie012:
Sorry for bumping this, but are there any plans to merge this PR soon?

drwpow (approved):
Great test, thank you!

drwpow:
@Newbie012 no worries; merged!

There’s an upcoming `6.x` release that was caused by a breaking change in #895 that required a little more followup work to release. We also have some additional `3.1` features that would be nice to add as well. So this will go out with `6.0.0`.

PR #919, opened 2022-06-07
Add support for anyOf/allOf/oneOf when node has explicit primitive type

michael-gillett:
## Problem
Currently the generated schema will simply mark any node with `type: string` as string in the output, regardless of whether these is additional specifiers like `anyOf`/ `allOf`/ `oneOf`

### Input
```yaml
openapi: 3.0
[...]
components:
  schemas:
    Example:
      description: Object with one property that is a string enum
      type: object
      properties:
        status:
          type: string
          allOf:
            - $ref: '#/components/schemas/ExampleStatus'
    ExampleStatus:
      type: string
      enum:
        - ACTIVE
        - INACTIVE

```

### Output
```typescript
[...]
export interface components {
  schemas: {
    /** @description Object with one property that is a string enum */
    Example: {
      status?: string;
    };
    /** @enum {string} */
    ExampleStatus: "ACTIVE" | "INACTIVE";
  };
}
```

## Desired Solution
The generate schema already has good support from string unions. Ideally these union types would be used when referenced in an `anyOf`/ `allOf`/ `oneOf` even if the property has `type: string`. I have added a new example spec outlining the desired behavior for the tests

### Output
```typescript
[...]
export interface components {
  schemas: {
    /** @description Object with one property that is a string enum */
    Example: {
      status?: components["schemas"]["ExampleStatus"];
    };
    /** @enum {string} */
    ExampleStatus: "ACTIVE" | "INACTIVE";
  };
}
```

## The Fix
This PR adds support for `anyOf`/ `allOf`/ `oneOf` nodes that have an explicit primitive type e.g. `type: string` by treating them as `object` internally.


drwpow (approved):
Great fix. Thank you! No feedback; happy to merge this.

michael-gillett:
Thanks a bunch!

PR #942, opened 2022-09-23
New `--alphabetize` switch: Sorts output file consistently

MrLeebo:
Screen Shot 2022-09-24 at 3 47 34 PM

This PR adds a new `--alphabetize` switch that sorts the entries in the typescript output file alphabetically. This ensures the typescript file appears in a consistent sort order. Sometimes with dynamically generated docs, the order of the operations can change just depending on the timing of when each controller gets parsed. This can create trouble if you're checking your types into git. It can be difficult to review what changes were actually made when the only thing you can see in git diff is all of the operations being shuffled around.


src/transform/parameters.ts
```diff
@@ -19,11 +19,16 @@ export function transformParametersArray(
 
   // sort into map
   let mappedParams: Record<string, Record<string, ParameterObject>> = {};
-  for (const paramObj of parameters as any[]) {
```
MrLeebo, line 22: I factored out the cast to `as any[]` and made some changes below to satisfy the type violations caused by that.

src/transform/parameters.ts
```diff
@@ -36,7 +41,7 @@ export function transformParametersArray(
           case 2: {
             mappedParams[reference.in][reference.name || paramName] = {
               ...reference,
-              $ref: paramObj.$ref,
+              ...("$ref" in paramObj ? { $ref: paramObj.$ref } : null),
```
drwpow, line 44: Good improvement!

src/utils.ts
```diff
@@ -285,3 +285,9 @@ export function replaceKeys(obj: Record<string, any>): Record<string, any> {
     return obj;
   }
 }
+
+export function getEntries<Item>(obj: ArrayLike<Item> | Record<string, Item>, options: GlobalContext) {
+  const entries = Object.entries(obj);
+  if (options.alphabetize) entries.sort(([a], [b]) => a.localeCompare(b, "en"));
```
drwpow, line 291: 👍 This is a smart place to alphabetize. Sorting too often could really slow things down. But for practical purposes I’d guess this sorts as little as possible if done on retrieval like this.

test/v3/index.test.js
```diff
@@ -71,6 +71,17 @@ describe("cli", () => {
       const expected = eol.lf(fs.readFileSync(new URL(`./expected/${filename}`, cwd), "utf8"));
       expect(generated).to.equal(expected);
     });
+
+    it(`reads ${schema} spec (v3) from file (alphabetize)`, async () => {
```
drwpow, line 75: These CLI tests are “snapshots,” where the output is so long and complicated it’s impossible to read through everything and see if it’s working as expected. But that’s their design—they’re mostly “smoke tests” that test that complex behavior can still work given a complex, real-world schema.

Alphabetization shouldn’t live here IMO, because it’s unclear if the original schemas were even unsorted in the first place (some examples are!).

Instead, I’d rather see the alphabetization tests live the spec tests themselves, testing that for every major type of OpenAPI object (`paths`, `components`, `operations`), and even adding tests for specific types (do object `properties` keys get sorted? Do endpoints? Do content types? `enum`s? etc. Because I think it’s very likely that some structures may be unsorted yet, and digging further into each data type would assure that.

src/utils.ts
drwpow, line 291:
```suggestion
  if (options.alphabetize) entries.sort(([a], [b]) => a.localeCompare(b, "en", { numeric: true }));
```

Small little creature comfort: add [numeric sort option](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#numeric_sorting) (sorts `1`, `2`, `10` rather than the default `1`, `10`, `2`)

drwpow (changes requested):
Love to see PRs like this! I’m in favor of this merging, but I’d like to see the tests moved out of snapshots and into unit tests where it makes sense.

MrLeebo:
@drwpow is there anything else I can do to help move this along?

drwpow:
This looks great! Sorry for the delay; will merge this today.

Just a heads up: this will be released in the `6.0` version along with some OpenAPI 3.1 additions and a breaking change.

drwpow (approved):

drwpow:
Merged exactly as-is in #966. Just fixed the test error & merge conflict but made no changes.

drwpow:
@all-contributors please add @MrLeebo for code, test

PR #943, opened 2022-09-29
Initial version generating express types from the base types

djMax:
I don't expect this to be merged, but (a) wanted to say :wave: and thanks and (b) perhaps you can think of some way to fit this concept inside the existing "intent" of openapi-typescript. Basically I want to define fully typed Express handlers easily from an OpenAPI spec. I could consume the output of openapi-typescript and then spit out another file, but this seemed cumbersome for what amounted to a "small number" of changes. I wonder if some sort of functional extension might work to amend the basic output.

Oh - the other change in here which I actually think might be useful is outputting operations even without operationIds. We've had mixed experience over the years with operationIds and have come to think they are generally a bad thing - that a deterministic mapping of method and path to method name is clearer in a microservices environment. Perhaps that's a spaces vs tabs thing, but definitely something we wanted to support.


drwpow:
This is a cool concept! Yeah I’d probably say releasing a separate package (`openapi-typescript-express`) and maintaining a separate repo would probably be best, especially for generating Express-style things.

But the improvements to operation could possibly merge into this project, provided there are some tests / there’s some clear win there.

PR #946, opened 2022-10-03
update GitHub Actions

drwpow:
Dropping Node 14 from the automated testing to just speed things up. This doesn’t mean this library is officially dropping support; just that speeding up CI is preferable right now.

If there are any Node 14 regressions, please file a bug! But no part of the API is expected to cause breakages at any time while Node 14 is still within LTS.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          version: 7
+      - uses: actions/setup-node@v3
+      - run: pnpm i
+      - run: npm run lint
+  test:
+    strategy:
+      matrix:
+        node-version: [16.x, 18.x]
+        os: [macos-12, ubuntu-latest, windows-latest]
```
drwpow, line 21: `macos-12` is actually more recent than `latest` 🤔

PR #951, opened 2022-10-05
[transform/paths] Remove parameters replacing

### LINKED ISSUE #950: --make-paths-enum generate enum with another paths
For example, it generates
```ts
export interface paths {
  '/download/{p1}': {
    // ...
  },
}

export enum ApiPaths {
  getDownload: '/download/:p1',
}
```

So I can't choose type:
```ts
type X = paths[ApiPaths.getDownload] // error
```

---

As I see, that is why paths in `export interface paths {` and `export enum ApiPaths {` are different:
https://github.com/drwpow/openapi-typescript/blob/a4e05fa005277ea1e42eb191921aae5d9f51293a/src/transform/paths.ts#L127-L128

@berzi hi! Could you answer if this change is needed?

Semigradsky:
Close #950


drwpow:
@all-contributors please add @Semigradsky for bug, code

PR #963, opened 2022-10-30
Update deps, swap tiny-glob for fast-glob

drwpow:
## Changes

Updates deps, swaps tiny-glob for fast-glob (former hasn’t been updated in quite some time).

⚠️ Note: `main` is failing CI; will fix in a following PR.

## How to Review

See comments

## Checklist

- [x] Unit tests updated
- [x] README updated


bin/cli.js
```diff
@@ -114,11 +116,11 @@ async function generateSchema(pathToSpec) {
 
   // output
   if (output === OUTPUT_FILE) {
-    let outputFilePath = path.resolve(process.cwd(), flags.output); // note: may be directory
```
drwpow, line 117: Overdue cleanup: `URL()` is a vastly-preferred way to work with the filesystem in Node, because it’s not only faster, but it also works in both Windows and *nix systems without any fuss. And no more awkward `path.join()` or conditional `path.sep` logic; simply use forward slashes and It Just Works™

`path.resolve()` is a bit of a convoluted mess under-the-hood. It’s great when it works, but it’s a headache when it doesn’t.

tsconfig.json
```diff
@@ -4,14 +4,15 @@
     "declaration": true,
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
-    "lib": ["ES2018"],
```
drwpow, line 7: Another overdue cleanup: most modern versions of Node don’t need legacy codegen.

Issue #961, opened 2022-10-30
📣 v6.0 RFC and call for contributors


drwpow:
### 👯‍♂️ v6.0

openapi-typescript has an upcoming 6.0 version, which started as [a breaking change](https://github.com/drwpow/openapi-typescript/pull/895) but has now expanded to include the following as well:

- [x] OpenAPI 3.1 support
   - [x] polymorphism (#969)
   - [x] `discriminator` (#953)
- [x] Closer adherence to the spec
- [x] Better JSDoc comment generation
- [x] Library cleanup to make contributions easier
- [x] Improvements to the formatter (now called `transform()`)

### 🤝 Call for contributors

This library would love to have more contributors and maintainers help keep this project going! In service of this, I’ve recently added:

- Issue templates for better bug filing
- Improved [contributor docs](https://github.com/drwpow/openapi-typescript/blob/main/CONTRIBUTING.md) to make contributions friendlier
- Codebase cleanup that makes onboarding simpler.

If you’ve been interested in contributing to opensource, or if you’ve found this tool useful at your workplace or personal projects and would love to contribute, please raise a hand ✋! Anyone of any skill level and experience with the project is welcome 🙂

_Note that at this time, contributing monetarily isn’t necessary. Perhaps in the future, but for now this library is held back by time constraints and number of maintainers._

### ❓ v6.0 RFC

Would love some 👍 or 👎 (comments welcome) on the following changes for 6.0 and beyond:

#### ❓ 1. Dropping Swagger 2.x support

Swagger 3.0 has been out for 5 years now and already many (most?) APIs have fully made the leap to 3.x. Anecdotally I haven’t seen an active 2.0 schema in the wild for quite some time. Dropping support for 2.x would shed quite a bit of dead weight in the codebase and allow for faster releases. Of course, users of Swagger 2.x could always use `openapi-typescript@5` indefinitely to generate types for their 2.x schemas, but won’t receive any future updates.

#### ❓ 2. Dropping Prettier

Prettier was originally added as a lazy way to clean up the generated output. But with the in-progress code improvements, the output is much better formatted to the point Prettier isn’t needed (to be clear, the output doesn’t and never will match Prettier 100%, but it’s still formatted neatly to the point Prettier isn’t needed). Dropping Prettier would mean **much faster generation** (initial testing shows a 3× speedup on large schemas) and it would also shed this library’s weight by quite a bit. And users could use their formatting method of choice (#959).

[...]

PR #968, opened 2022-11-03
Refactor

drwpow:
## Changes

Refactors app, accomplishes many goals outlined in #961.


leppaott:
Just out of curiosity were those `test/v3` files deleted deliberately? Made me think when I saw there were `test/v2` files indeed deleted.

drwpow:
v6 drops support for Swagger 2.0 so the v2 folder was dropped and v3 just became `test`. And the tests were cleaned up and deduplicated to be easier to manage (the refactor required they all had to change anyway).

Issue #985, opened 2022-11-11
Invalid output when a parameter description contains a "."


axelhzf:
**Description**

When a parameter description contains a "." at the end, it produces an invalid typescript output. Here is a small schema that produce an invalid output

```yml
openapi: 3.0.0
info:
  title: A
  version: '1.0'
paths:
  '/programs/{id}/data':
    parameters:
      - name: id
        in: path
        schema:
          type: string
        required: true
        description: Description.
    get:
      responses:
        200:
          description: OK.
          content:
            "application/json":
              schema:
                type: string
```

The code generated is

```ts
export interface paths {
  "/programs/{id}/data": {
    get: {
      responses: {
        /** @description OK. */
        200: {
          content: {
            "application/json": string;
          };
        };
      };
    };
    {
    parameters: {
        /** @description Description. */
      path: {
        id: string;
      };
    };
  }
  };
}
```

There are an extra `{}` around `parameters`. If I remove the "." at the end of the description, the generation works correctly.

| Name                 | Version                      |
|:---------------------|:-----------------------------|
| `openapi-typescript` | `6.0.2`                      |
| Node.js              | `16.17.1`                      |
| OS + version         | `macOS 13` |

**Reproduction**

Generate the types from the previous schema

**Expected result**

_(in case it’s not obvious)_

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/CONTRIBUTING.md))

axelhzf:
I think the problem is related to this unwrap_obj_re here

https://github.com/drwpow/openapi-typescript/blob/main/src/transform/path-item-object.ts#L48
https://github.com/drwpow/openapi-typescript/blob/main/src/transform/path-item-object.ts#L12

axelhzf:
Fixed by #986

PR #986, opened 2022-11-12
Fix generation of parameters with a dot in the description

axelhzf:
## Changes

Fixes the generation of parameters containing a `.` in the description. See https://github.com/drwpow/openapi-typescript/issues/985

## How to Review

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/test/paths-object.test.ts
+++ b/test/paths-object.test.ts
@@ -20,7 +20,7 @@ describe("Paths Object", () => {
   test("basic", () => {
     const schema: PathsObject = {
       "/api/v1/user/{user_id}": {
-        parameters: [{ name: "page", in: "query", schema: { type: "number" } }],
+        parameters: [{ name: "page", in: "query", schema: { type: "number" }, description: "Page number." }],
         get: {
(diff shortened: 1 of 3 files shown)


test/paths-object.test.ts
```diff
@@ -20,7 +20,7 @@ describe("Paths Object", () => {
   test("basic", () => {
     const schema: PathsObject = {
       "/api/v1/user/{user_id}": {
-        parameters: [{ name: "page", in: "query", schema: { type: "number" } }],
+        parameters: [{ name: "page", in: "query", schema: { type: "number" }, description: "Page number." }],
```
axelhzf, line 23: I'm not sure if I should create a new test instead of updating an existing one

drwpow: Yeah that’s always a tough one and there’s no right answer. I’d say this one is fine here but if we started collecting a lot of these we should break them out into its own test.

drwpow (approved):

drwpow:
@all-contributors please add @axelhzf for code, bug, test

PR #990, opened 2022-11-15
Use lowercase headers

HiiiiD:
## Changes

Undici returns headers in lowercase as it can be seen [here](https://github.com/nodejs/undici/blob/a3ce745b0c5d7c837ecae2359434ab4550a6e0ea/lib/core/util.js#L214). This PR updates the `content-type` check in order to use the lowercase version.

## How to Review

## Checklist

- [ ] Unit tests updated
- [ ] README updated

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -131,9 +131,9 @@ export default async function load(
         }
       }
       const res = await request(schema, { method: (options.httpMethod as Dispatcher.HttpMethod) || "GET", headers });
-      const contentType = Array.isArray(res.headers["Content-Type"])
-        ? res.headers["Content-Type"][0]
-        : res.headers["Content-Type"];
+      const contentType = Array.isArray(res.headers["content-type"])
+        ? res.headers["content-type"][0]
+        : res.headers["content-type"];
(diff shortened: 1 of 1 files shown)


timothymctim:
This would fix an issue for me: my backend server sends the `Content-Type` header in lower case. Since key indices in JavaScript are case sensitive, this would fix it for me, but this fix would break the code for others as it wouldn’t match `Content-Type` any more. The following code would normalize the header key first before selecting and using it.
```
let contentType = res.headers[Object.keys(res.headers).find(key => key.toLowerCase() === 'content-type')]
contentType = Array.isArray(contentType)
    ? contentType[0]
    : contentType;
```

HiiiiD:
As I said in the first PR comment, Undici returns all headers in lowercase, so that's not an issue @timothymctim

timothymctim:
> As I said in the first PR comment, Undici returns all headers in lowercase, so that's not an issue @timothymctim

Indeed, that the case! Sorry, my bad :)

drwpow (approved):

drwpow:
@all-contributors please add @HiiiiD for code, test

Issue #976, opened 2022-11-09
Type helpers + @typescript-eslint/no-unused-vars


imagoiq:
**Description**

With the latest version (v6), the generated schema contains some type helpers even if there are not used which is catch as a unused variables.

```
12:5  error  'OneOf' is defined but never used. Allowed unused vars must match /_/u  @typescript-eslint/no-unused-vars
```

| Name                 | Version                      |
|:---------------------|:-----------------------------|
| `openapi-typescript` | `6.0.1`                      |
| Node.js              | `16.15.0`                      |
| OS + version         | `Windows 10` |

drwpow:
I’d be open to a PR if someone would like to tackle this!

PR #992, opened 2022-11-15
Add type helpers only when used

imagoiq:
Fix https://github.com/drwpow/openapi-typescript/issues/976 by checking if `OneOf` type is added to the schema. The fix uses the easiest path, but perhaps not the best for long term (if there are other type helpers added especially). Perhaps, the context should be aware if type helpers are required? (if it's not already the case and that I miss something).

### DIFF
--- a/test/index.test.ts
+++ b/test/index.test.ts
@@ -7,7 +7,9 @@ const BOILERPLATE = `/**
  * Do not make direct changes to the file.
  */
 
+`;
 
+const TYPE_HELPERS = `
 /** Type helpers */
 type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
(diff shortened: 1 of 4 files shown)


test/index.test.ts
```diff
@@ -442,6 +444,56 @@ export interface components {
 
 export type external = Record<string, never>;
 
+export type operations = Record<string, never>;
+`);
+      });
+    });
+
+    describe("OneOf type helpers", () => {
```
drwpow, line 452: 👏 Awesome test!

drwpow (approved):
LGTM! Yeah in general I’d like to avoid scanning text if possible, but profiling this it seems to be fast enough to not have any real performance impacts even on big schemas, so this works for me!

But if more things were added to text scanning we might want to move it into global context or something; this is fine but some other text scanning methods may have serious performance concerns.

Issue #925, opened 2022-07-26
Folder as output doesn't work


AzazelloB:
This command `npx openapi-typescript src/api/v1.0/schemas/**/*.json --output src/api/v1.0/types/` throws an error "src/api/v1.0/types/src/api/v1.0/schemas/schema.ts". Because in openapi-typescript/bin/cli.js:120 filename is not actually a filename.

Possible fix:
replace: ``const filename = pathToSpec.replace(new RegExp(`${path.extname(pathToSpec)}$`), ".ts");``
to: ``const filename = path.basename(pathToSpec.replace(new RegExp(`${path.extname(pathToSpec)}$`), ".ts"));``

drwpow:
I think there is a bug here, but I’m having a little trouble understanding how to reproduce it. Could you list out the contents of all your `.json` files so I could recreate?

AzazelloB:
Cannot reproduce this issue myself anymore. Don't know what was it. The only inconvenience now that in a folder structure like that
```
├── src
│   ├── api
│   │   ├── v1.0
│   │   │   ├── schemas
│   │   │   │   ├── file1.json
│   │   │   │   ├── file2.json
│   │   │   ├── types
```
it creates .ts file with full folder path in the name
```
├── src
│   ├── api
│   │   ├── v1.0
│   │   │   ├── schemas
│   │   │   │   ├── file1.json
│   │   │   │   ├── file2.json
│   │   │   ├── types
│   │   │    │  ├── src
│   │   │    │  │   ├── api
│   │   │    │  │   │   ├── v1.0
│   │   │    │  │   │   │   ├── schemas
│   │   │    │  │   │   │   │   ├── file1.ts
│   │   │    │  │   │   │   │   ├── file2.ts
```

Which is inconvenience but expected

BTMPL:
Was anyone able to sidestep this issue? Downgrading to 5.x solves the problem, but I'd like to use the native `fetch` helper, which was added in 6.x

PR #999, opened 2022-11-19
Fix bug in generating multiple spec to folder

BTMPL:
## Changes

Fixes https://github.com/drwpow/openapi-typescript/issues/925

The issue is that the original `outputDir` was being mutated, so each itteration would create more nested folders.

## How to Review

1. Create 2 directories holding 2 different OpenAPI contracts (e.g. `src/contracts/service1/openapi.yaml` and `src/contracts/service2/openapi.yaml`
2. Attempt to generate types to different target directory, e.g. `src/generated` via `openapi-typescript src/contracts/**/openapi.yaml --output ./src/generated/api/`

In the non-pathed version the command will fail, multiple invalid directories will be created.

## Checklist

- [ ] Unit tests updated
- [ ] README updated

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 import fs from "node:fs";
+import path from "path";
 import { URL } from "node:url";
 import glob from "fast-glob";
 import parser from "yargs-parser";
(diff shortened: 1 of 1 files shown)


drwpow:
Thank you!

drwpow (approved):

drwpow:
@all-contributors please add @BTMPL for code, bug, test

PR #1002, opened 2022-11-20
Use overridable fetch instead of unidici's request

### LINKED ISSUE #988: TypeError: Cannot read properties of undefined (reading 'schema')
**Description**

I have got an error in 6.0.3 version when I'm trying to generate types. When I downgraded to 5.4.1 version, everything works fine.

```
file:///Users/adrian/generator-test/node_modules/openapi-typescript/dist/load.js:140
    const currentSchema = options.schemas[schemaID].schema;
                                                    ^

TypeError: Cannot read properties of undefined (reading 'schema')
    at load (file:///Users/adrian/generator-test/node_modules/openapi-typescript/dist/load.js:140:53)
    at async openapiTS (file:///Users/adrian/generator-test/node_modules/openapi-typescript/dist/index.js:37:5)
    at async generateSchema (file:///Users/adrian/generator-test/node_modules/openapi-typescript/bin/cli.js:89:18)
    at async main (file:///Users/adrian/generator-test/node_modules/openapi-typescript/bin/cli.js:156:5)
```

| Node.js              | `18.12.1`                      |
| OS + version         | `macOS 11` |

**Reproduction**

**Expected result**

_(in case it’s not obvious)_

**Checklist**

version 6.0.3:
image

version 5.4.1:
image

### LINKED ISSUE #998: Option to generate class methods instead of static methods.
**Description**

Why?

If we generate classes with methods, We could use these generated classes as the implementation contract for the backend service.

By this, I can write an additional middleware to plug it with express easily.

For example.
```
interface MyBackendClass implements OpenApiGeneratedService {
    // override generated methods here
    async getUser(id: string): Promise<User> {
        ...
    }
}
```

**Proposal**

Provide an option to generate functions into class methods instead of static methods.
```
openapi --useClassMethod --input ./spec.yaml --output ./sdk
```

yacinehmito:
## Changes

When loading the schema, the `load` function used to only consider the Content-Type header when capitalized. However, the HTTP specification states that headers are to be treated as case-insensitive. HTTP/2 further enforces this by stating that all HTTP headers must be converted to lowercase when encoded. Therefore, many recent implementations move away from the conventional capitalization in favour of headers in lowercase, such as the Nest framework.

In issue #988, it is shown that loading the schema from a live server implemented in Nest fails when the endpoint does not include the desired extension.

This commit fixes the issue by replacing unidici's request by fetch. The request function returns headers as a record, where the keys are left exactly as the headers are received, meaning that the casing of headers affect the output. Fetch, on the other hand, returns headers in a special object whose accessor method `get` will lookup headers in a case-insensitive way.

We took this opportunity to make the fetch implementation overridable. This way, it is possible for the caller to swap it out for another. It is a step towards portability, as openapiTS does not run as-is in the browser, Deno or Cloudflare Workers at the moment.

Additionally, unit tests have been added for the `load` function.

Fixes #988.

## How to Review

I'll put comments where appropriate.

## Checklist

- [x] Unit tests updated
- [ ] README updated — _not applicable_
- [ ] `examples/` directory updated (if applicable) — _not applicable_

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,7 +9,7 @@ import transformParameterObject from "./transform/parameter-object.js";
 import transformRequestBodyObject from "./transform/request-body-object.js";
 import transformResponseObject from "./transform/response-object.js";
 import transformSchemaObject from "./transform/schema-object.js";
-import { error, escObjKey, getEntries, indent } from "./utils.js";
+import { error, escObjKey, getDefaultFetch, getEntries, indent } from "./utils.js";
 export * from "./types.js"; // expose all types to consumers
 
 const EMPTY_OBJECT_RE = /^\s*\{?\s*\}?\s*$/;
@@ -67,6 +67,7 @@ async function openapiTS(
     urlCache: new Set(),
     httpHeaders: options.httpHeaders,
     httpMethod: options.httpMethod,
+    fetch: options.fetch ?? getDefaultFetch(),
   });
 
   // 1. basic validation
--- a/src/load.ts
+++ b/src/load.ts
@@ -1,5 +1,6 @@
 import type {
   ComponentsObject,
+  Fetch,
   GlobalContext,
   OpenAPI3,
   OperationObject,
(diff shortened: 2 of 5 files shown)


yacinehmito:
I only saw #990 after the fact. My apologies.

drwpow (approved):

yacinehmito:
I just rebased.

drwpow:
@all-contributors please add @yacinehmito for bug

PR #1014, opened 2022-12-09
Escape constant string in generated schema

mvdbeek:
## Changes

Given the schema
```ts
const schema: SchemaObject = {
  type: "object",
  properties: { "constant": {"const": "a", "type": "string"} },
  required: ["constant"],
};
```
the generator would build this
```
  /** @constant */
  constant: a;
```
which isn't valid typescript. I think the same issue was properly taken care of for enums in https://github.com/drwpow/openapi-typescript/blob/main/src/transform/schema-object.ts#L80, so I've used the same approach here.

## How to Review

I hope this should be pretty uncontroversial. Maybe the string cast should also be applied if no type is specified, but not sure about that. This would be the minimal fix for the API we're documenting.

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/src/transform/schema-object.ts
+++ b/src/transform/schema-object.ts
@@ -62,7 +62,13 @@ export function defaultSchemaObjectTransform(
 
   // const (valid for any type)
   if (schemaObject.const) {
-    return transformSchemaObject(schemaObject.const as any, {
+    let schemaConst = schemaObject.const as any;
+    if ("type" in schemaObject) {
+      if (schemaObject.type === "string") {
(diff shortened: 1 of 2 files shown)


drwpow (approved):
Thank you! This looks great to me; just failing a tiny lint check.

nikeee:
Anything we can do to get this merged?

mvdbeek:
Ah, I had missed the notification. Thanks for the review @drwpow!

drwpow:
@all-contributors please add @mvdbeek for code, bug, test

PR #1022, opened 2023-01-04
Make params non-nullable even if all params are optional

### LINKED ISSUE #995: Property 'query' does not exist on type
**Property 'query' does not exist on type**

In v5 I could get the type of query parameters from `operations['myOperationId']['parameters']['query']`, but this no longer works with v6. Instead I get type error `Property query does not exist on type { query?: { orderBy?: "id" | "created" | "name" | undefined; } | undefined; } | undefined`

```
type myQueryType = operations['myOperationId']['parameters']['query']

query' does not exist on type '{ query?: { orderBy?: "id" | "created" | "name" | undefined; } | undefined; } | undefined'
```

| Node.js              | `v18.9.1`                      |
| OS + version         | `Ubuntu 22.04` |

**Reproduction**

Generate types with v6.0.3 from YAML file that has a GET path with query parameters. Specific YAML file I used [https://github.com/nsb/myjobplanner-api/blob/286941f03ef8fc496c985cc4035922f3fd7d81f8/openapi.yaml](https://github.com/nsb/myjobplanner-api/blob/286941f03ef8fc496c985cc4035922f3fd7d81f8/openapi.yaml)

**Expected result**
I should be able to get the type definition of query parameters without type error

`type myQueryType = operations['myOperationId']['parameters']['query']`

**Checklist**

sgrimm:
In 5.x, an optional query parameter's type could be referenced with an
expression such as

    type myQueryType = operations['myOperation']['parameters']['query']

In 6.0, that only works if there is at least one mandatory query
parameter in addition to the optional one. This makes query parameter
types much more awkward to reference in application code if all the
parameters are optional.

Revert to the 5.x behavior: if there are any parameters, optional or
not, the `parameters` object is always present. If there are any query
parameters, the `query` object is always present, same with `path` and
the other parameter types.

## Changes

Fixes #995

## How to Review

The actual code change here is trivial (the vast majority of changed lines 
are changes to the test cases). But per the discussion on #995, it's not
entirely clear why this behavior changed between 5.x and 6.0. Reverting to
the 5.x behavior will restore compatibility with existing application code,
but it may break code that relied on the 6.0 behavior depending on how that
code was constructing its type references.

## Checklist

- [X] Unit tests updated
- [ ] README updated - _no README changes needed_
- [X] `examples/` directory updated (if applicable)

### DIFF
--- a/examples/github-api.ts
+++ b/examples/github-api.ts
@@ -86825,8 +86825,8 @@ export interface operations {
      * 
      * The permissions the installation has are included under the `permissions` key.
      */
-    parameters?: {
-      query?: {
+    parameters: {
+      query: {
         outdated?: string;
       };
     };
@@ -87147,10 +87147,10 @@ export interface operations {
      * 
      * For more information on creating a personal access token, see "[Creating a personal access token](/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token)."
      */
-    parameters?: {
+    parameters: {
         /** @description A cursor, as given in the [Link header](https://docs.github.com/rest/overview/resources-in-the-rest-api#link-header). If specified, the query only searches for events after this cursor. */
         /** @description A cursor, as given in the [Link header](https://docs.github.com/rest/overview/resources-in-the-rest-api#link-header). If specified, the query only searches for events before this cursor. */
-      query?: {
+      query: {
         date_start?: string;
         date_end?: string;
       };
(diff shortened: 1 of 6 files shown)


mkosir:
Hey @drwpow thanks for this awesome library.
Any thoughts on this PR, if it can be merged as @sgrimm suggested, since this issues is quite blocking for many of us?

drwpow (approved):
Thank you! I agree with this change.

jmwalsh91:
I know this has made it's way into main already, but I just wanted to thank @sgrimm for this from the bottom of my heart.

PR #1027, opened 2023-01-21
feat: add with required utility to enforce properties of allOf

### LINKED ISSUE #657: Advanced `required` enforcement for `allOf` + `required` case
We have an OpenAPI schema where we combine `allOf` with `required` in the way that we enforce some attributes defined by `allOf` to be required, something along these lines:

```yaml
SomeEntity:
  required:
    - name
  allOf:
    - "$ref": "#/definitions/l10nName"
    - ...something else...
```

In this case `name` is defined in `l10nName` subschema and we enforce it to be required for `SomeEntity`. The generated code, however, does not take that into account, generating code with `name` property being optional.

This is something that still could be expressed via TypeScript as something like:

```
SomeEntity_raw: <whatever definition is being generated right now>,
SomeEntity: SomeEntity_raw & Required<Pick<SomeEntity_raw, "name">>,
```

What do you think? It seems to be relatively simple to detect with something like "if property is in required, but not explicitly defined in `properties` on current type, generate enforcing wrapper"

Swiftwork:
## Changes

Resolves #657

## How to Review

Resolves the issue described in #657 and updates the unit tests. Should be a fairly safe change as it is just additive to the current code generation without breaking changes. This solution will only work for OpenAPI 3.1 as described by the comment here https://github.com/drwpow/openapi-typescript/issues/657#issuecomment-1399274607

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/examples/digital-ocean-api.ts
+++ b/examples/digital-ocean-api.ts
@@ -4,7 +4,10 @@
  */
 
 
-/** Type helpers */
+/** WithRequired type helpers */
+type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
+
+/** OneOf type helpers */
 type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
 type XOR<T, U> = (T | U) extends object ? (Without<T, U> & U) | (Without<U, T> & T) : T | U;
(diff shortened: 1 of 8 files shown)


drwpow (approved):
This looks great. There may be some cases where we might need a little more code generation than this, but I think this is an improvement. We can come back to those cases as they become clearer.

drwpow:
@all-contributors please add @Swiftwork for code, bug, test

PR #1032, opened 2023-02-21
feat: Under-specified objects are Record<string, unknown>

### LINKED ISSUE #1031: type: "object" being parsed as empty object
**Description**

A brief description of the issue.
A schema of type: "object" with no other information specified is being parsed as an empty (Record<string, never>) object, but I would expect it to be an Record<string, any>

I think it would have to be changed in here, correct me if I'm wrong:
https://github.com/drwpow/openapi-typescript/blob/3ffb475d15352ebc1dd2f823b01d3ae8b18ef68b/src/transform/schema-object.ts#L244

<html><body>

Name | Version
-- | --
openapi-typescript | 6.1.0
Node.js | v16.13.2
OS + version | Windows 10

</body>
</html>

**OpenAPI**
image

**Generated TS**

```ts
    DocumentModel: {
      /** Type */
      type?: string;
      /** Content */
      content?: (components["schemas"]["DocumentModel"])[];
      /** Text */
      text?: string;
      /** Attrs */
      attrs?: Record<string, never>;
      /** Marks */
      marks?: (components["schemas"]["Mark"])[];
    };
```

**Expected TS*

```ts
    DocumentModel: {
      /** Type */
      type?: string;
      /** Content */
      content?: (components["schemas"]["DocumentModel"])[];
      /** Text */
      text?: string;
      /** Attrs */
      attrs?: Record<string, any>;
      /** Marks */
      marks?: (components["schemas"]["Mark"])[];
    };
```

**Checklist**

- [x ] My OpenAPI schema passes a validator
- [x ] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/CONTRIBUTING.md))

duncanbeevers:
## Changes

This changes the fallback behavior when generating types for `object` schemas without specified properties to be `Record<string, unknown>` instead of `Record<string, never>`

This change likely constitutes a major version bump, but more-closely resembles the v5 behavior.
We could hide it behind a CLI flag for backwards compatibility.

Closes #1031

## How to Review

The example snapshots are checked as part of the tests. Reviewing the changes to these snapshots is the best way to get a sense of the scope of the changes.

For many of the changes I also went back to the source OpenAPI yaml definitions to verify that `Record<string, unknown>` accurately represents the types returned from the actual API implementations.

## Checklist

- [x] Unit tests updated
- [ ] README updated (I'm already in there)

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,7 @@ The following flags can be appended to the CLI command.
 | `--http-method`                | `-m`  | `GET`    | Provide the HTTP Verb/Method for fetching a schema from a remote URL                                                         |
 | `--immutable-types`            |       | `false`  | Generates immutable types (readonly properties and readonly array)                                                           |
 | `--additional-properties`      |       | `false`  | Allow arbitrary properties for all schema objects without `additionalProperties: false`                                      |
+| `--empty-objects-unknown`      |       | `false`  | Allow arbitrary properties for schema objects with no specified properties, and no specified `additionalProperties`          |
 | `--default-non-nullable`       |       | `false`  | Treat schema objects with default values as non-nullable                                                                     |
 | `--export-type`                | `-t`  | `false`  | Export `type` instead of `interface`                                                                                         |
 | `--path-params-as-types`       |       | `false`  | Allow dynamic string lookups on the `paths` object                                                                           |
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -22,6 +22,7 @@ Options
   --export-type, -t            (optional) Export "type" instead of "interface"
   --immutable-types            (optional) Generates immutable types (readonly properties and readonly array)
   --additional-properties      (optional) Allow arbitrary properties for all schema objects without "additionalProperties: false"
+  --empty-objects-unknown      (optional) Allow arbitrary properties for schema objects with no specified properties, and no specified "additionalProperties"
   --default-non-nullable       (optional) If a schema object has a default value set, don’t mark it as nullable
   --support-array-length       (optional) Generate tuples using array minItems / maxItems
   --path-params-as-types       (optional) Substitute path parameter names with their respective types
@@ -46,6 +47,7 @@ const flags = parser(args, {
     "help",
     "version",
     "defaultNonNullable",
+    "emptyObjectsUnknown",
     "immutableTypes",
     "contentNever",
     "exportType",
@@ -89,6 +91,7 @@ async function generateSchema(pathToSpec) {
   // generate schema
   const result = await openapiTS(pathToSpec, {
     additionalProperties: flags.additionalProperties,
+    emptyObjectsUnknown: flags.emptyObjectsUnknown,
     auth: flags.auth,
     defaultNonNullable: flags.defaultNonNullable,
     immutableTypes: flags.immutableTypes,
--- a/src/index.ts
+++ b/src/index.ts
@@ -46,6 +46,7 @@ async function openapiTS(
     transform: options && typeof options.transform === "function" ? options.transform : undefined,
     postTransform: options && typeof options.postTransform === "function" ? options.postTransform : undefined,
     immutableTypes: options.immutableTypes || false,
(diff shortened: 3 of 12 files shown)


duncanbeevers:
I've been using the 5.4.0 branch for quite some time to great effect.
In order to migrate to 6.1.x, I needed this functionality, and also https://github.com/drwpow/openapi-typescript/pull/1022.

I published a private package with these changes and am using it happily in the existing app. (although it did reveal some places where I needed to make `requestBody` required.

image

leppaott:
Upvoted as seems also what we need to migrate to v6.

drwpow:
I’m not sure I agree with this change. Yes 5.x had that behavior, but this change does make the generated types for `"type": "object"` equivalent to `"type": "object", "additionalProperties": true`.

Is there a reason why `"type": "object", "additionalProperties": true` isn’t a better expression in the schema for generating the `Record<string, unknown>` type?

duncanbeevers:
I primarily based this change on schemas seen in the wild.

For example, on the Stripe API, a notoriously-well-documented schema, all those `Record<string, never>` values mean those parts of the API are not just unknown to TypeScript, but basically off-limits.

In my opinion, this proposed behavior better-matches the intent of real-world schema authors.

duncanbeevers:
The `--additional-properties` flag already exists. However, using it causes the types to be widened across a huge number of objects, not just those where no specific schema is specified.

This change only affects the fallback behavior where *no* schema for an object is specified.

If there were an additional command-line flag, I'd imagine it would be something like `--bare-objects-unknown`.

duncanbeevers:
@drwpow I've updated this PR to put the functionality behind the `--empty-objects-unknown` flag.

I tested the node-js generated pieces, but I didn't see a straightforward way to test CLI flag permutations against the example OpenAPI schema corpus.

I did go ahead and regenerate the examples in `main`, and opened [this PR](https://github.com/drwpow/openapi-typescript/pull/1037) for it.
Then I used the "clean" base to verify the CLI flag behavior when regenerating the examples.

mitchell-merry:
According to [the OpenAPI docs](https://swagger.io/docs/specification/data-models/data-types/#free-form):

"A free-form object (arbitrary property/value pairs) is defined as:
```yml
type: object
```

This is equivalent to
```yml
type: object
additionalProperties: true
```
and

```yml
type: object
additionalProperties: {}
```
"

so I believe `Record<string, unknown>` is the correct default behaviour here, and `Record<string, never>` when `additionalProperties: false`.

examples/digital-ocean-api/resources/apps/apps_get_logs_active_deployment.yml
```diff
@@ -0,0 +1,47 @@
...
+  default:
+    $ref: ../../shared/responses/unexpected_error.yml
+
+x-codeSamples:
+  - $ref: 'examples/curl/apps_get_logs_active_deployment.yml'
+
+security:
+  - bearer_auth:
+    - 'read'
+
```
drwpow, line 47: 👍 Thanks for updating this example!

drwpow:
@mitchell-merry ah thank you for providing that detail! Yes it does seem to be explicit in the Swagger 3.0 documentation. It’s much more ambiguous in the 3.1 specification.

> I think I disagree that a flag is necessary here. The correct behaviour is `Record<string, unknown>`, and this isn't a breaking change, since it doesn't widen/narrow any existing types.

I’d still like for this to be a flag for now, because it would be a behavior change for `6.x` of this library, and it would break types for people already using this version.

Because the discussion over generated types is stemming from **ambiguity** in an OpenAPI schema, we can could take 1 of 2 approaches: either err toward being too strict (you can’t access any properties that aren’t defined until you update your OpenAPI schema) or too lax (you can freely access any properties not in your OpenAPI schema by just asserting types). Just speaking personally, I‘d rather err on the side of being more strict, and requiring the OpenAPI schema to accurately describe all shapes rather than TypeScript. You’ll always get better mileage out of that approach, if you ask me!

PR #1036, opened 2023-02-23
Fix readonly nullable properties edge case

mtth:
## Changes

The current logic generates invalid TypeScript code for nullable properties which include at least one union (potentially nested) when the `immutableTypes` option is set.

Simple example:

```yaml
  type: array
  nullable: true
  items:
    oneOf:
      - type: number
      - type: string
```

Generated code before and after this PR:

```typescript
  readonly before: readonly ((number | string)[]) | null; // Invalid
  readonly after: (readonly (number | string)[]) | null; // OK
```

## How to Review

## Checklist

- [x] Unit tests updated
- [x] README updated


drwpow (approved):
Awesome fix, thank you!

drwpow:
@all-contributors please add @mtth for code, bug, test

PR #1049, opened 2023-03-12
If no type is specified, type as unknown

### LINKED ISSUE #1039: Property without type should be declared as unknown
**Description**

| `openapi-typescript` | `6.1.1`                      |
| Node.js              | `18.12.2`                      |
| OS + version         | `Ubuntu 22.04.2` |

**Reproduction**

1. Create a file named _api.json_ with the following content:

_api.json_
```json
{
  "openapi": "3.0.0",
  "components": {
    "schemas": {
      "User": {
        "type": "object",
        "additionalProperties": false,
        "properties": {
          "Info": {
            "nullable": false
          }
        }
      }
    }
  }
}
```

2. Run `openapi-typescript`:

```console
$ npx openapi-typescript api.json --output type.ts
✨ openapi-typescript 6.1.1
🚀 api.json → file:///path/to/type.ts [12ms]
```

3. You should get the following result:

_type.ts_
```ts
/**
 */


export type paths = Record<string, never>;

export interface components {
  schemas: {
    User: {
      Info?: Record<string, never>;
    };
  };
  responses: never;
  parameters: never;
  requestBodies: never;
  headers: never;
  pathItems: never;
}

export type operations = Record<string, never>;
```

As you can see, `schemas.User.Info` is defined as `Record<string, never>` which is incorrect because `Info` might contain a `string`, `number` or `boolean` and not necessarily an `object`.

**Expected result**

```ts
export interface components {
  schemas: {
    User: {
      Info?: unknown;
    };
  };
  responses: never;
  parameters: never;
  requestBodies: never;
  headers: never;
  pathItems: never;
}
```

**Checklist**

mitchell-merry:
## Changes

Fixes #1039.

## How to Review

Related docs: https://swagger.io/docs/specification/data-models/data-types/#any

I wish I could just update the snapshots without updating the specs.

## Checklist

- [x] Unit tests updated
- [ ] README updated


drwpow:
Thanks so much for doing the work on this, but I believe this conflicts with #1032 and the `--empty-objects-unknown` flag. I also gave some [additional context here](https://github.com/drwpow/openapi-typescript/issues/1039#issuecomment-1496742003). But please let me know if I’m mistaken.

mitchell-merry:
It doesn't conflict with that PR, since this is specifically when `type` is not specified. I've left my comments on that issue.

drwpow:
My mistake. I agree this is a good change. Sorry for the misunderstanding on my part! If the merge conflicts are fixed I’m happy to merge this and agree this is a bugfix and not a breaking change.

mitchell-merry:
Awesome, thank you! Will do once I get a chance

test/utils.test.ts
```diff
@@ -18,4 +19,26 @@ describe("utils", () => {
       return expect(tsUnionOf(...[0, true, "string", '"hello"'])).toBe('0 | true | string | "hello"');
     });
   });
+
+  describe("tsIntersectionOf", () => {
```
mitchell-merry, line 23: I happened to also add a utils.test.ts before this one made it into main :D but our test cases use different approaches. Would you like me to convert my cases into the pre-existing ones?

examples/digital-ocean-api.ts
```diff
@@ -3483,7 +3483,7 @@ export interface external {
      * @example Sammy Shark
      */
     user_name?: string;
-    user_billing_address?: Record<string, never> & external["resources/billing/models/billing_address.yml"];
```
mitchell-merry, line 3486: This is an intentional change - since `unknown & AnyType === AnyType` is a tautology, I've added code to tsIntersectionOf to remove these redundant bits of the intersection.

These extraneous unknowns are caused by the definitions in the schema:
```yaml
user_billing_address:
  allOf:
    - description: some description
    - $ref: 'billing_address.yml'
```

We (correctly) interpret this as two schemas, the first registers to be a schema with no type defined and therefore `unknown`. Why they did this escapes me.

examples/github-api-next.ts
```diff
@@ -95046,7 +95046,7 @@ export interface operations {
              */
             path: "/" | "/docs";
           });
-        }) & (Record<string, never> | Record<string, never> | Record<string, never> | Record<string, never> | Record<string, never>);
```
mitchell-merry, line 95049: wonder if we should de-duplicate unions

mitchell-merry: Let's keep it out of this PR

mitchell-merry: #1069

test/utils.test.ts
drwpow: No, the opposite—we can move the ones there into this one. Thanks for adding these!

drwpow (approved):
Awesome, thank you!

PR #1050, opened 2023-03-12
docs: Update transform date-time example

duncanbeevers:
## Changes

Encourage best-practices by considering nullable properties in the transform function.

## How to Review

Consider whether the extra complexity is warranted.

## Checklist

- 🚫 Unit tests updated
- [x] README updated
- 🚫 `examples/` directory updated (if applicable)

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ By default, openapiTS will generate `updated_at?: string;` because it’s not su
 const types = openapiTS(mySchema, {
   transform(schemaObject, metadata): string {
     if ("format" in schemaObject && schemaObject.format === "date-time") {
-      return "Date";
+      return schemaObject.nullable ? "Date | null" : "Date";
     }
   },
 });


drwpow (approved):
Thank yoU!

drwpow:
@all-contributors please add @duncanbeevers for docs

thebuilder:
Kinda related question - How do you actually convert the `string` to a `Date` object, when you get it from the API with the `fetch` method? 🤔

thebuilder:
Yeah, that was i thought. Kinda a problem to have the type mismatch, unless you have a secure way of converting the string to a date.

duncanbeevers:
@drwpow @thebuilder I think such a layer belongs in a library, though this one may not be the place since it consumes the `openapi-typescript`-generated types, rather than the schema itself.

I have been enthusiastic about the approach taken by `express-openapi-validator` with its `serDes` option.
https://github.com/cdimascio/express-openapi-validator/wiki/Documentation#%EF%B8%8F-serdes-optional

Unified, bi-directional rules that merge the run-time and type-level representations of data has been _pretty great_.

drwpow:
> Yeah, that was i thought. Kinda a problem to have the type mismatch

> @drwpow @thebuilder I think such a layer belongs in a library

One problem I’ve been dealing with lately is representing JSONB columns in an OpenAPI spec. In the simplest sense, the JSON endpoint returns them as a `string` (for Reasons™) because they are JSON-serializable metadata. But is `string` really the best type? After all, in our case, they actually do have a defined structure, and can be better represented as an `object` SchemaObject with defined properties. And whether that parsing happens in the endpoint, or the client, or some middleware layer, I’d argue it’s better to have that be typed out as an `object` because whenever that’s consumed, that’s the only way it’s ever interacted with—as an `object` and never as a `string` (barring the `requestBody` to update it, which is more of an implementation detail).

> Unified, bi-directional rules that merge the run-time and type-level representations of data has been _pretty great._

💯 All that to say, I’m definitely on board with this philosophy in general, and there are valid reasons beyond simple date/`format`s when it is helpful to give more definition/functionality to API data. Even if the “where that happens” answer is blurry, and differs wildly based on your stack / technology needs / other reasons.

Fortunately, openapi-typescript is only one layer and the `transform()` option is “take it or leave it—” there if you need it, but also in no way are we suggesting this is best practice, or this makes the most sense for your specific usecase.

Issue #1051, opened 2023-03-13
Node Error: t.includes is not a function


willsamu:
**Description**

Parsing OpenAPI Schema exported from Stoplight fails for a valid schema. The maybeTypes uses `include` which only works for array or string inputs. Failed if the input is number (e.g. `maybeTypes = [ 2, 4, 5 ]` or `maybeTypes = [ 'Record<string, never>', 1, 2, 3, 4, 5, 6, 7, 8, 9 ]`).

```
$ npx openapi-typescript schema.yaml

file:///Users/myuser/.npm/_npx/21d31b7563e6adbc/node_modules/openapi-typescript/dist/transform/schema-object.js:55
        if (maybeTypes.some((t) => t.includes("{")))
                                                              ^

TypeError: t.includes is not a function
    at file:///Users/myuser/.npm/_npx/21d31b7563e6adbc/node_modules/openapi-typescript/dist/transform/schema-object.js:55:63
    at Array.some (<anonymous>)
    at defaultSchemaObjectTransform (file:///Users/myuser/.npm/_npx/21d31b7563e6adbc/node_modules/openapi-typescript/dist/transform/schema-object.js:55:49)
    at transformSchemaObject (file:///Users/myuser/.npm/_npx/21d31b7563e6adbc/node_modules/openapi-typescript/dist/transform/schema-object.js:3:20)
    at defaultSchemaObjectTransform (file:///Users/myuser/.npm/_npx/21d31b7563e6adbc/node_modules/openapi-typescript/dist/transform/schema-object.js:116:45)
    at transformSchemaObject (file:///Users/myuser/.npm/_npx/21d31b7563e6adbc/node_modules/openapi-typescript/dist/transform/schema-object.js:3:20)
    at transformMediaTypeObject (file:///Users/myuser/.npm/_npx/21d31b7563e6adbc/node_modules/openapi-typescript/dist/transform/media-type-object.js:5:12)
    at transformRequestBodyObject (file:///Users/myuser/.npm/_npx/21d31b7563e6adbc/node_modules/openapi-typescript/dist/transform/request-body-object.js:26:31)
    at transformOperationObject (file:///Users/myuser/.npm/_npx/21d31b7563e6adbc/node_modules/openapi-typescript/dist/transform/operation-object.js:79:37)
    at transformPathItemObject (file:///Users/myuser/.npm/_npx/21d31b7563e6adbc/node_modules/openapi-typescript/dist/transform/path-item-object.js:19:35)
[...]

PR #1056, opened 2023-03-23
Fix `oneOf` number const

qnp:
## Changes

Fixing https://github.com/drwpow/openapi-typescript/issues/1051 alongside the fact that `0` as a const value was evicted.

## How to Review

I wrote tests, they all pass now.

FYI, I wanted to make sure everything is working as expected, so I added tests for the `tsOneOf` util. IMHO, all the utils should be tested (but I did not have time to write these tests).

## Checklist

- [x] Unit tests updated
- [ ] README updated => *Not relevant for the bugfix*

### DIFF
--- a/src/transform/schema-object.ts
+++ b/src/transform/schema-object.ts
@@ -62,7 +62,7 @@ export function defaultSchemaObjectTransform(
   }
 
   // const (valid for any type)
-  if (schemaObject.const) {
+  if (schemaObject.const !== null && schemaObject.const !== undefined) {
     let schemaConst = schemaObject.const as any;
     if ("type" in schemaObject) {
       if (schemaObject.type === "string") {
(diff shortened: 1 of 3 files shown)


src/transform/schema-object.ts
```diff
@@ -62,7 +62,7 @@ export function defaultSchemaObjectTransform(
   }
 
   // const (valid for any type)
-  if (schemaObject.const) {
+  if (schemaObject.const !== null && schemaObject.const !== undefined) {
```
drwpow, line 65: 👍

test/utils.test.ts
```diff
@@ -0,0 +1,21 @@
+import { tsUnionOf } from "../src/utils";
+
+describe("utils", () => {
```
drwpow, line 3: These tests aren’t really necessary as they are all tested by proxy with other tests, and these are such simple functions with such simple inputs/outputs they have clear and obvious failures.

However, the coverage is appreciated and it doesn’t hurt to have. Thank you.

drwpow (approved):
Great fix; thank you!

drwpow:
@all-contributors please add @qnp for code, bug, test

Issue #1070, opened 2023-04-07
Generated type of `additionalProperties` values should not include undefined


dancrumb:
**Description**

When trying to describe a simple dictionary object in Swagger, equivalent to a TS `Record<string,string>`, the generated type is, instead `Record<string, string | undefined>`

See https://swagger.io/docs/specification/data-models/dictionaries/

I dug into the code and noticed the comment on this line:
https://github.com/drwpow/openapi-typescript/blob/main/src/transform/schema-object.ts#L194

> // note: `| undefined` is required to mesh with possibly-undefined keys

I haven't found any kind of spec that outlines clearly how `additionalProperties` should be interpreted with regards to whether a value is required. I concede that this may be a matter of opinion.

However, I would contend that supporting unknown additional properties would imply that such properties must have a value; otherwise, why would that property even exist? Properties with undefined values are omitted from JSON anyway, so it's pretty hard to send them over the wire.

Therefore, I think this implementation may be flawed and should be changed.

If not, would it be possible to add a parameter to toggle this behavior to make `additionalProperties` with a defined type include or exclude `undefined`?

**OpenAPI**

```yaml
openapi: 3.0.0
info:
  version: 0.5.0
  title: Sutro Central
  description: "These are the endpoint for the Sutro Central backend"
components:
  schemas:
    Demo:
      type: object
      required:
        - dataSchema
      properties:
        dataSchema:
          type: object
          minProperties: 1
          additionalProperties:
            type: string
            minLength: 1

paths:
  /dummy:
    get:
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Demo"

```

**Generated TS**

```ts
/**
 * This file was auto-generated by openapi-typescript.
 * Do not make direct changes to the file.
 */


export interface paths {
  "/dummy": {
    get: {
      responses: {
        /** @description OK */
        200: {
          content: {
            "application/json": components["schemas"]["Demo"];
          };
        };
      };
    };
  };
}

export type webhooks = Record<string, never>;

export interface components {
  schemas: {
    Demo: {
      dataSchema: {
        [key: string]: string | undefined; // this is what I'd like to see changed
      };
    };
  };
  responses: never;
  parameters: never;
  requestBodies: never;
  headers: never;
  pathItems: never;
}

export type external = Record<string, never>;

export type operations = Record<string, never>;
```

**Expected TS**

```ts
/**
 * This file was auto-generated by openapi-typescript.
[...]

PR #1071, opened 2023-04-07
fix: 🐛 change how additionalProperties are typed

dancrumb:
## Changes

This removes the `| undefined` addition to additional properties in an object

## How to Review

It's a very simple PR, so review is more: "Is this the behavior that is desired?".

See issue #1070

## Checklist

- [x] Unit tests updated
- [ ] ~README updated~ N/A
- [ ] ~`examples/` directory updated (if applicable)~


drwpow:
Thank you for adding this PR! But I think this behavior is desired for now. Explanation left on #1070.

PR #1073, opened 2023-04-08
Fix sample code in README.md

shoffmeister:
`openapi-fetch` at 0.0.6 uses `baseUrl` as the property name.


mitchell-merry (approved):
Thank you!

mitchell-merry:
@all-contributors please add @shoffmeister for docs

Issue #1011, opened 2022-12-01
Incorrect tuples generation with v6.x


Jmenache:
**Description**

Hello. Thank you very much for your great work on this library 👍 !!

From v6.0 tuples stopped being properly generated.
I get the following :

```ts
export interface paths {
  "/tupleTest": {
    post: {
      requestBody: {
        content: {
          "application/json; charset=UTF-8": {
            tupleTest: [[[object Object], [object Object]], [[object Object], [object Object]]];
          };
        };
      };
    };
  };
}
```

| Name                 | Version                      |
|:---------------------|:-----------------------------|
| `openapi-typescript` | `6.1.0`                      |
| Node.js              | `16.13.0`                      |
| OS + version         | `Windows 10` |

**Reproduction**

```json
{
  "openapi": "3.1.0",
  "paths": {
    "/tupleTest": {
      "post": {
        "requestBody": {
          "required": true,
          "content": {
            "application/json; charset=UTF-8": {
              "schema": {
                "type": "object",
                "properties": {
                  "tupleTest": {
                    "type": "array",
                    "minItems": 2,
                    "maxItems": 2,
                    "items": [
                      {"type": "string"},
                      {"type": "number"}
                    ]
                  }
                },
                "required": ["tupleTest"],
                "additionalProperties": false
              }
            }
          }
        }
      }
    }
  }
}

```

**Expected result**

With v5.4.0 I get :
```ts
export interface paths {
  '/tupleTest': {
    post: {
      requestBody: {
        content: {
          'application/json; charset=UTF-8': {
            tupleTest: [string, number];
          };
        };
      };
    };
  };
}
```

**Checklist**

- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/CONTRIBUTING.md))

mitchell-merry:
This appears to be working in my tests (6.1.1). Could you confirm if this is still happening? i.e. using your example the output is:

```ts
/**
 * This file was auto-generated by openapi-typescript.
 * Do not make direct changes to the file.
 */

export interface paths {
  "/tupleTest": {
    post: {
      requestBody: {
        content: {
          "application/json; charset=UTF-8": {
            tupleTest: [string, number];
          };
        };
      };
    };
  };
}

export interface components {}

export interface operations {}

export interface external {}
```

liangskyli:
@mitchell-merry : I also have the same problem here
[...]

PR #1082, opened 2023-04-23
fix: tuples type generation support (#1011)

liangskyli:
## Changes
fix bug: #1011

## How to Review

Tests

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/test/schema-object.test.ts
+++ b/test/schema-object.test.ts
@@ -109,6 +109,12 @@ describe("Schema Object", () => {
         expect(generated).toBe("(string)[]");
       });
 
+      test("tuple array", () => {
+        const schema: SchemaObject = { type: "array", items: [{ type: "string" },{"type": "number"}],minItems:2,maxItems:2 };
+        const generated = transformSchemaObject(schema, options);
+        expect(generated).toBe("([string,number])[]");
+      });
+
       test("ref", () => {
         const schema: SchemaObject = { type: "array", items: { $ref: 'components["schemas"]["ArrayItem"]' } };
         const generated = transformSchemaObject(schema, options);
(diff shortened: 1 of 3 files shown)


test/schema-object.test.ts
```diff
@@ -109,6 +109,12 @@ describe("Schema Object", () => {
         expect(generated).toBe("(string)[]");
       });
 
+      test("tuple array", () => {
+        const schema: SchemaObject = { type: "array", items: [{ type: "string" },{"type": "number"}],minItems:2,maxItems:2 };
```
drwpow, line 113:
```suggestion
        const schema: SchemaObject = { type: "array", prefixItems: [{ type: "string" },{"type": "number"}],minItems:2,maxItems:2 };
```

drwpow (changes requested):
Thank you very much for the PR! I’m in favor of merging this, but would like this to require usage of `prefixItems` [according to the spec](https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation) rather than `items`

liangskyli:
@drwpow : You can add support for prefixItems, but I think items also need support！
see: https://json-schema.org/draft/2020-12/release-notes.html
image
because: Draft 2019-09 use items, so we alse need support items!

liangskyli:
@drwpow :You can also look at this！

https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation
image
In Draft 4 - 2019-09, tuple validation was handled by an alternate form of the items keyword. When items was an array of schemas instead of a single schema, it behaved the way prefixItems behaves.

drwpow (approved):
Thank you for correcting me! I had only used `prefixItems` before for the tuple type, not items. Glad to see `items` supports it too. This all looks good to merge, then!

PR #1085, opened 2023-04-26
fix: tuples type generation bug, and unit test bug.

liangskyli:
## Changes

Sorry, when solving the problem of generating tuples type before, the unit test assertion was written incorrectly, but now it is correct

## How to Review

test

tuples type schema
{ type: "array", items: [{ type: "string" },{"type": "number"}],minItems:2,maxItems:2 }
The corresponding data should be:
[string, number]

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/src/transform/schema-object.ts
+++ b/src/transform/schema-object.ts
@@ -127,9 +127,11 @@ export function defaultSchemaObjectTransform(
     if (schemaObject.type === "array") {
       indentLv++;
       let itemType = "unknown";
+      let isTupleType = false;
       if (schemaObject.items) {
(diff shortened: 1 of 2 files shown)


liangskyli:
@drwpow : I am very sorry, I found a bug in the tuples type and found that you updated a version today

liangskyli:
#1011

drwpow (approved):

drwpow:
No problem at all. Thank you! Will push out a patch with just this.

drwpow:
Released as 6.2.3!

Issue #940, opened 2022-09-09
Make parent parameter object optional if all fields are optional


dolencd:
When a specification defines some parameters, such as: 
```
{
  "name": "id",
  "in": "path",
  "required": true,
  "type": "string"
},
{
  "name": "accept-language",
  "in": "header",
  "required": false,
  "type": "string"
}
```

The current output contains: 
```
parameters: {
    path: {
      id: string;
    };
    header: {
      "accept-language"?: string;
    };
};
```

Maybe this would be a good option rather than the default behaviour, but I'd like the `header` key in this example to be optional. This makes sense since all the fields included in it are optional and you may want to include none of them.

drwpow:
This makes sense, and I’d argue this is expected behavior, because it means most requests likely happen without specifying those headers at all.

I don’t think this scenario would apply to all situations where this happens, but here in headers I think it’s a good improvement.

Would love a PR fixing this if someone has time.

drwpow:
Update: this has been added in #968 and will ship with the next v6 release.

PR #1092, opened 2023-05-07
make query optional if all params are optional

ysmood:
If all query parameters are optional we should use `query?: { ... }` instead of `query: { ... }`. Or when we use libs like `openapi-fetch` we have to always write an empty `query: {}` on the request like:

```ts
const res = await client.get('/do/{id}', {
  params: {
    query: {},
    path: {
        id: 123,
    },
  },
})
```

We expect to use code like this:

```ts
const res = await client.get('/do/{id}', {
  params: {
    path: {
        id: 123,
    },
  },
})
```

### DIFF
--- a/examples/digital-ocean-api.ts
+++ b/examples/digital-ocean-api.ts
@@ -752,7 +752,7 @@ export interface external {
    */
   "resources/1-clicks/oneClicks_list.yml": {
     parameters: {
-      query: {
+      query?: {
         type?: external["resources/1-clicks/parameters.yml"]["oneClicks_type"];
       };
     };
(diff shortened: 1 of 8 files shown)


drwpow:
Thanks for adding this; this already happens for the overall `parameters` object (#940) but it seems we didn’t mirror that behavior for `query`. I’m in favor of this just for better consistency.

Also, openapi-fetch is still immature, and types will improve over time (such as the behavior you just described—that will go away eventually). So I don’t want to make changes to this library _just yet_ just to support openapi-fetch. But again, there’s reason for this to pass independently of that. And as openapi-fetch matures, it is a good public implementation of how openapi-typescript can be used.

ysmood:
Sure, I updated the tests and examples, it's update to you to modify or closed this PR.

I think it's not just for `openapi-fetch`, we also need this feature for our custom lib to work better.

drwpow (approved):

Issue #1093, opened 2023-05-08
use the same openapi, json file generate ok, but yaml file generate error


liangskyli:
**Description**

the same openapi:
when use openapi.yaml have a generate bug.
when use openapi.json no generate bug.

use openapi.yaml generate error data:
```
/**
 * This file was auto-generated by openapi-typescript.
 * Do not make direct changes to the file.
 */


export interface paths {
  "/root/v4/getQueryParams1-v4": {
    get: {
      parameters: {
        query: {
          /** @description activityBases */
          activityBases: (external["components[%22schemas%22][%22ActivityBase%22]"])[];
        };
      };
      responses: {
        /** @description Success */
        200: {
          content: {
            "text/plain": string;
          };
        };
      };
    };
  };
}

export type webhooks = Record<string, never>;

export interface components {
  schemas: {
    getQueryParams1Request: {
      /** @description activityBases */
      activityBases: (external["components[%22schemas%22][%22ActivityBase%22]"])[];
    };
    ActivityBase: {
      name: string;
    } & {
      id: number;
    };
  };
  responses: never;
  parameters: never;
  requestBodies: never;
  headers: never;
  pathItems: never;
}

export type external = Record<string, never>;

export type operations = Record<string, never>;
```

| Name                 | Version                      |
|:---------------------|:-----------------------------|
| `openapi-typescript` | `6.2.4`                      |
| Node.js              | `16.14.0`                      |
| OS + version         | `macOS 13, Windows 10, etc.` |

**Reproduction**

_How can this be reproduced / when did the error occur?_

openapi.json
```
{
  "openapi": "3.1.0",
  "info": {
    "title": "custom title",
    "license": {
      "name": "This file is auto generated, do not edit!"
    },
    "version": "0.4.1"
  },
  "paths": {
    "/root/v4/getQueryParams1-v4": {
      "get": {
        "tags": ["Test4"],
        "responses": {
          "200": {
            "description": "Success",
            "content": {
              "text/plain": {
                "schema": {
                  "type": "string"
                }
              }
            }
          }
        },
        "parameters": [
          {
            "name": "activityBases",
            "in": "query",
            "required": true,
            "schema": {
              "type": "array",
              "items": {
                "$ref": "#/components/schemas/ActivityBase"
              }
            },
            "description": "activityBases"
          }
        ]
      }
    }
  },
  "components": {
    "schemas": {
      "getQueryParams1Request": {
        "type": "object",
        "properties": {
[...]

PR #1097, opened 2023-05-08
fix: load yaml file transform ref bug

liangskyli:
## Changes

fix: #1093 

## How to Review

## Checklist

- [x] Unit tests updated
- [ ] README updated


liangskyli:
I found a problem with the code being processed
How to determine $ref has been transformed?

packages/openapi-typescript/src/load.ts
```diff
@@ -279,6 +279,10 @@ export default async function load(
 
         const ref = parseRef(node.$ref);
 
+        // when ref.path is [], $ref has been transform, skip transform
+        if (ref.path.length === 0 && (ref.filename.startsWith("components[") || ref.filename.startsWith("external["))) {
```
drwpow, line 283: Hm this may not be true in all cases; `$ref`s could also reference the `operations` interface as well.

This feels like masking a potential bug rather than pinpointing exactly _why_ the `$ref` isn’t being transformed correctly.

Is this an invalid schema?

Or is this a specific bug with specific `$ref`s?

Rather than change this, we should find out _exactly_ what kind of `$ref` isn’t being transformed properly.

liangskyli: The reason for this bug is: 
use items: &ref_0 and items: *ref_0, Causes the ref reference in the currentSchema to point to the same address, and the currentSchema data has changed, resulting in ref transform many times！
I use JSON.parse and JSON.stringify to handle schema data before walk currentSchema to refactor code.

liangskyli:
ci error, because i merge master code, now master code ci error too.

drwpow:
Thanks so much for opening this PR. This updated version is an improvement over the initial PR. But the solution here is too broad, and would have performance impacts for all users when this is a problem unique to people using `js-yaml` to generate YAML files.

The problem lies in the `parseRef()` utility—its one and only job is to transform `#/components/schemas/Foo` into `components["schemas"]["Foo"]`. However, in this bug, it’s actually receiving the string `"components[\"schemas\"][\"Foo\"]"` from `js-yaml` and it just doesn’t know how to parse it. So rather than `JSON.parse(JSON.stringify())` the entire schema—which could not only be slow, but may introduce other bugs—we should just teach `parseRef()` to handle this alternate $ref form.

If it’s all right, I’d prefer handling the fix and writing a test, to ensure it’s working as expected. I’d also like to make a performance improvement here as well.

I’m going to close this PR as it’s I’d prefer another solution to this bug, but you still get full marks for your contribution! 🙂

PR #1111, opened 2023-05-13
feat: introduce excludeDeprecated option

benevbright:
## Changes

Often time we want to exclude deprecated fields completely from generated types so we can spot deprecated fields asap. 
Unfortunately, TS doesn't have a such option: https://github.com/microsoft/TypeScript/issues/49433
The code editor shows deprecated fields with scratched out, which is nice but not enough.

For example, you created a new version of SDK, then you wish TS would complain about use of deprecated fields asap but it doesn't. 

## How to Review

It should be straight forward to review but I'm not sure if I should add this option on every single `getEntries` usages. 

## Checklist

- [x] Unit tests updated
- [x] README updated


drwpow (approved):
This is a cool idea, thank you! I’m in favor of merging this as this seems useful and spiritually-aligned with the `deprecated` property. No feedback on the PR—everything looks well-considered and reasonable. Great tests, flag addition, everything 🎉

benevbright:
@drwpow super! thanks! 👍

I just added README as well. (could you approve again?)

PR #1113, opened 2023-05-15
Update examples, update validator recommendation

drwpow:
## Changes

Updates examples.

Removes swagger-cli as a subtle suggestion in favor of the much better Redocly CLI.

Also makes a few other minor docs improvements.


docs/src/content/docs/about.md
```diff
@@ -23,6 +23,7 @@ description: Additional info about this project
 ## Project goals
 
 1. Support converting any valid OpenAPI schema to TypeScript types, no matter how complicated.
+1. Generate **runtime-free types** for maximum performance.
```
drwpow, line 26: 🤔 Wondering if there’s a better way to say this. Only lately realized that there is an entire school of thought that believes codegen is a natural pair to OpenAPI and I strongly disagree. After all, why impose heavy runtime SDKs on things that don’t need it? TypeScript has the superpower of **static inference** that can deliver all the resiliency of static typing but without any runtime impact. In a sense, codegen feels antithetical to how TypeScript works itself.

Anyway, there are probably better terms to express this sentiment, but just putting this placeholder idea here for now.

PR #1130, opened 2023-05-23
fix: support nested path parameters in --path-params-as-types

barakalon:
## Changes

The existing code for `--path-params-as-types` assumes path parameters are defined on the PathItemObject:
```javascript
{
  openapi: "3.1",
  info: { title: "Test", version: "1.0" },
  paths: {
    "/user/{user_id}": {
      parameters: [{ name: "user_id", in: "path" }],
    },
  },
}
```

But its valid to define them on the OperationObject:
```javascript
{
  openapi: "3.1",
  info: { title: "Test", version: "1.0" },
  paths: {
    "/user/{user_id}": {
      get: {
        parameters: [{ name: "user_id", in: "path" }],
      },
      put: {
        parameters: [{ name: "user_id", in: "path" }],
      }
    },
  },
}
```

[FastAPI](https://github.com/tiangolo/fastapi) does it this way.

## How to Review

This is both my first openapi-typescript PR _and_ my first typescript PR. Don't hold back.

## Checklist

- [x] Unit tests updated
- [x] README updated

### DIFF
--- a/.changeset/wise-badgers-kick.md
+++ b/.changeset/wise-badgers-kick.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+support nested path parameters in --path-params-as-types
--- a/packages/openapi-typescript/src/transform/paths-object.ts
+++ b/packages/openapi-typescript/src/transform/paths-object.ts
@@ -1,7 +1,21 @@
-import type { GlobalContext, PathsObject } from "../types.js";
+import type { GlobalContext, PathsObject, PathItemObject, ParameterObject, ReferenceObject, OperationObject } from "../types.js";
 import { escStr, getEntries, indent } from "../utils.js";
 import transformParameterObject from "./parameter-object.js";
 import transformPathItemObject from "./path-item-object.js";
+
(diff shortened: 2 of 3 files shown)


drwpow (approved):
Great tests, and great code, thank you!

mkosir:
This change made majority of `query` keys optional, which is unwanted behaviour since I can't access query parameters anymore.

drwpow:
> This change made majority of `query` keys optional, which is unwanted behaviour since I can't access query parameters anymore.

`query` keys being optional _should_ only happen now when they are all optional. It shouldn’t happen if even one is required. This is actually more consistent with other types generated by this library, so I’m in favor of it.

However, if you have even _one_ parameter that is required, then `query` should be required as well. If you find that happening, then please file a new issue (and no need for a reproduction; if that’s happening it should be easy to reproduce).

If you need to access that `query` object in types, you can always use the `NonNullable<>` built-in TS generic.

mkosir:
Not sure if I'm in favour of this approach, but thanks for clarifying it.
Yes I use `Required<>`/`NonNullable <>`, find it a bit cumbersome to access parameter types.

PR #1149, opened 2023-05-26
fix(openapi-typescript): Stringify primitive constant values

### LINKED ISSUE #1118: Constant string types result in invalid TypeScript code
**Description**

A property that must have a constant string value

```
              "name": {
                "const": "local-userpass",
                "description": "The name of the authentication provider. For Email/Password authentication, this value is always `local-userpass`."
              },
```
yields invalid TypeScript code (the value has to be quoted):

```
      /**
       * @description The name of the authentication provider. For Email/Password authentication, this value is always `local-userpass`. 
       * @constant
       */
      name?: local-userpass;

```

| `openapi-typescript` | `6.2.4`                      |

**Reproduction**

This can be reproduced by processing the OpenAPI specification of the [MongoDB Atlas App Services Admin API](https://www.mongodb.com/docs/atlas/app-services/admin/api/v3/).

duncanbeevers:
Closes #1118

## Changes

https://github.com/drwpow/openapi-typescript/issues/1118

The `escStr` function _claims_ to always return a string, but in fact would return whatever was passed into it, preserving the original type.
This PR aligns the behavior of the function with type signature, and keeps the original behavior _mostly_ the same.
Because these values were being turned into strings (through implicit `toString()` during backtick string expansion), the final generated output was mostly correct.
Additionally, because there was special handling for `const` values with a specified type, _some_ string consts were being quoted.

In this PR, I removed the special handling of `const` types, and instead pass every const value through `escStr`.
`escStr` then passes non-string arguments through `JSON.stringify`.

## How to Review

Ensure JSON stringification of primitive types is the correct treatment for OpenAPI `const` members, even with no `type` specified.

I added tests for `string` and `number` consts with both explicit and implicit `type` arguments.

* [ ] Where in the OpenAPI spec is `const` described? I couldn't find it in the 3.1 spec.
* [ ] How should other primitive constant types be treated? (eg, `true`, `false`)

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/packages/openapi-typescript/src/transform/schema-object.ts
+++ b/packages/openapi-typescript/src/transform/schema-object.ts
@@ -41,13 +41,7 @@ export function defaultSchemaObjectTransform(schemaObject: SchemaObject | Refere
 
   // const (valid for any type)
   if (schemaObject.const !== null && schemaObject.const !== undefined) {
-    let schemaConst = schemaObject.const as any;
-    if ("type" in schemaObject) {
-      if (schemaObject.type === "string") {
-        schemaConst = escStr(schemaConst);
-      }
-    }
-    return transformSchemaObject(schemaConst, {
+    return transformSchemaObject(escStr(schemaObject.const) as any, {
       path,
       ctx: { ...ctx, immutableTypes: false, indentLv: indentLv + 1 }, // note: guarantee readonly happens once, here
     });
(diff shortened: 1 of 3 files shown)


mitchell-merry:
> Where in the OpenAPI spec is const described? I couldn't find it in the 3.1 spec.

It's in the JSON schema spec: https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values

> How should other primitive constant types be treated? (eg, `true`, `false`)

I'm pretty sure they should be treated as the boolean values rather than the string "true"/"false". That's how the yaml spec works I believe. https://yaml.org/type/bool.html

drwpow (approved):
This looks great to me! I can’t recall the reason why that particular code block in `transform/schema-object.ts` is so anti-idiomatic. So the fact that there’s no change to tests/output generation, and this PR cleans up that bit of code is probably a good sign overall 🙂

PR #1156, opened 2023-06-01
fix: Avoid index signature error for paths with empty params

horaklukas:
## Changes

It fixes using `--path-as-params-types` option for paths that doesn't contain a param (like `/api/user`). In that case, using the option causes generating path name as `[path: '/api/user']` and produces compiler error `An index signature parameter type cannot be a literal type or generic type. Consider using a mapped object type instead.
`

## How to Review

It fixes [this fix](https://github.com/drwpow/openapi-typescript/pull/1130). The `pathParams` variable  is already non-empty variable, but can have empty size.

## Checklist

- [x] Unit tests updated
- [x] README updated - not necessary [bugfix]
- [x] `examples/` directory updated (only applicable for openapi-typescript) - not necessary [bugfix]

### DIFF
--- a/.changeset/selfish-ears-hope.md
+++ b/.changeset/selfish-ears-hope.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Avoid index signature TS error for paths with empty params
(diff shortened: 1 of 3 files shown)


drwpow (approved):
Looks great! Great tests, too.

Issue #1165, opened 2023-06-13
Error when response is plain text / csv


Visual-Dawg:
**Description**

The parsing of the response fails when the response is not `json`.
Subsequently responses of `text/plain` and `text/csv` throw an error, which is not returned, but remains uncaught.

**Reproduction**

Query an api endpoint which return `text/*`, like `https://api64.ipify.org`.

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-fetch/blob/main/CONTRIBUTING.md))

drwpow:
This is improved in `0.3.0`. Now, the `.json()` response will fall back to `.text()`. Or you can override the default behavior with the [parseAs option](https://openapi-ts.pages.dev/openapi-fetch/api/#fetch-options)

CEOehis:
@drwpow is there a parse-as flag for the openapi-typescript CLI?

drwpow:
> @drwpow is there a parse-as flag for the openapi-typescript CLI?

There’s not. It sounds like your server isn’t sending the right MIME types. The best solution is fixing that. But as a workaround you could write a short script that saves that file to a temporary local `.json` file and just running `openapi-typescript` locally.

Making the wrong “guess” of whether something is JSON or YAML is expensive and slows down everyone.

Issue #1161, opened 2023-06-09
Expose the logic computing `finalUrl`


ysulyma:
**Description**

I would like to remove hardcoded API routes from my code, and use this library instead. I no longer need these routes for the fetchers, but I still need them for testing code. Proposal is to export the internal logic for computing `finalUrl`.

**Proposal**

```ts
import fetchMock from "fetch-mock-jest";
import createClient from "openapi-fetch";

import { paths } from "./v1"; // (generated from openapi-typescript)

const { route } = createClient<paths>({ baseUrl: "https://myapi.dev/v1/" });

fetchMock.once(
  // "https://myapi.dev/v1/post/my-post?version=2"
  route("/post/{post_id}", {
    path: { post_id: "my-post" },
    query: { version: 2 },
  }),
  { /* ... */ }
);
```

**Checklist**

- [x] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-fetch/blob/main/CONTRIBUTING.md))

drwpow:
I’d be fine with exposing this under a named import, e.g. `import { createFinalUrl } from 'openapi-fetch`. Probably for most peoples’ testing purposes they’d probably still need to maintain their own testing utility functions for their setup (e.g. which mocking tool they use), but I agree it’d be annoying to have to monitor parity with openapi-fetch’s implementation detail of how a URL is created.

ysulyma:
@drwpow if this is exposed on the return of `createClient`, then it can include `baseUrl` in the output. Or would you prefer an API like `createFinalUrl(url, {baseUrl, path, query})`?

drwpow:
Something like the latter, as this is probably more for testing purposes than anything else. And in that light, a more functionally-pure utility like your suggestion would be easier to deal with in general

drwpow:
This should be exposed now in `0.3.0`.

Issue #1159, opened 2023-06-06
Response body should be able to be read after fetching


Raiondesu:
**Description**

Currently, if one attempts to read a response body in any way (`.text()`/`.blob()`/etc.) after fetching, they will be presented with the following error:
```
TypeError: Failed to execute ['blob'/'text'/etc.] on 'Response': body stream already read
```

**Reproduction**

1. Set up any openapi-fetch client:
```ts
import createClient from 'openapi-fetch';
import type { paths } from './schema'; // generated via openapi-typecript

const client = createClient<paths>({ /* any config */ });
```
2. Call `json()` or `blob()` in a `then` block after the request:
```ts
client
  .get(''/* any request */, { /* any requestinit */ })
  .then(r => r.response.blob()) // <-- throws TypeError
```
3. Run and observe the console:
```
TypeError: Failed to execute 'blob' on 'Response': body stream already read
```

**Expected result**

I argue the user should expect the `response` property to contain a fully functional original response, with the ability to parse the body without any errors in any way should the need arise.

**Checklist**

- [x] I’m willing to open a PR, as I have already implemented a fix using [response.clone()](https://developer.mozilla.org/en-US/docs/Web/API/Response/clone)

drwpow:
Question: why would you want to get `.text()` for a typed fetch library for a JSON API? If your schema doesn’t list a JSON response, this library will automatically call `.text()` for you.

Trying to understand the real-world usecase here to provide a good workaround. I’d like to not have to require everyone calling `.json()` manually themselves for 99.9% of usecases; IMO that’s a bad API that leads to noisy boilerplate (fine when using the fetch API, but not for a wrapper meant for JSON calls)

drwpow:
> My usecase (and I assume it's possible I'm not the only one) is to read the `blob` from the response body.

Could you give more specific detail why you’re trying to read a `blob` for an endpoint marked as JSON in your schema? I’m looking for any clue your schema would give me that you would want to do this.

As you’re probably already aware, that error comes directly from the native fetch API. because we call `json()` internally, you can’t then call `blob()` afterward. Same as the normal fetch API. So internally, if we know that an endpoint has no `application/json` content type, this library calls `.text()` instead. We could add `blob()` support as well based off content-type hints, but that’s currently a bit of a can of worms.

[...]

Issue #1150, opened 2023-05-26
Simple join of `baseUrl` and `path` may result in invalid URLs


rdmurphy:
**Description**

Hello! I noticed that when you have a `baseUrl` with a trailing slash (e.g. `https://example.com/api/`) and `paths` with a leading slash (e.g. `/create-post`) it results in an invalid URL/request. If I am reading the code correctly I think the problem surfaces in this line:

https://github.com/drwpow/openapi-typescript/blob/65ff4b1826f1a70bce2acf5cd0ba77ca18b08502/packages/openapi-fetch/src/index.ts#L69

I think it assumes those two strings can always just be joined directly. If this is the case then I think [the README is also suggesting this would work when it may not](https://github.com/drwpow/openapi-typescript/blob/65ff4b1826f1a70bce2acf5cd0ba77ca18b08502/packages/openapi-fetch/README.md?plain=1#L17-L25).

**Reproduction**

Here is a tiny types file for making a request to `https://jsonplaceholder.typicode.com`:

```ts
/**
 * This file was auto-generated by openapi-typescript.
 * Do not make direct changes to the file.
 */


export interface paths {
  "/posts": {
    get: {
      responses: {
        /** @description OK */
        200: {
          content: {
            "application/json": ({
                userId?: number;
                id?: number;
                title?: string;
                body?: string;
              })[];
          };
        };
      };
    };
  };
}
```

And the script I ran:

```ts
// I'm using Deno, but it's the same otherwise
import createClient from "https://esm.sh/openapi-fetch@0.2.1";

import type { paths } from "./api.d.ts";

const client = createClient<paths>({
  baseUrl: "https://jsonplaceholder.typicode.com/", // note the trailing slash
});

const posts = await client.get("/posts", {});
```

**Expected result**

I think it'd be helpful if `openapi-fetch` smoothed this over! Maybe the leading slash of a `path` should be removed pre-joining if the `baseUrl` has a trailing slash. I could also see this using the [`URL()` constructor instead](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) to get a proper join for free.

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-fetch/blob/main/CONTRIBUTING.md))

drwpow:
[...]

PR #1169, opened 2023-06-16
openapi-fetch improvements

drwpow:
## Changes

Fixes:

- https://github.com/drwpow/openapi-typescript/issues/1165
- https://github.com/drwpow/openapi-typescript/issues/1161
- https://github.com/drwpow/openapi-typescript/issues/1159
- https://github.com/drwpow/openapi-typescript/issues/1150
- https://github.com/drwpow/openapi-typescript/issues/1123

### ✨ new `parseAs` option

To address the handling of non-JSON responses (#1123, #1159), you can now parse the response as something other than `.json()`

```ts
const { data } = await get('/my/url', { parseAs: 'arrayBuffer' });
```

`parseAs` takes any of the following options: 

- `json`
- `text`
- `arrayBuffer`
- `blob`
- `stream` (this skips parsing, and returns the Response.body as-is)

The reason for the manual detection is:

- Your OpenAPI schema isn’t readable at runtime, so when the response comes in, this has no clue what your OpenAPI schema said it would be
- Relying on the `Content-Type` header isn’t possible because it could be wrong or ambiguous (e.g. it’s not clear from `Content-Type` alone when someone may want a `.blob()` vs `.arrayBuffer()`)
- You may want more conditional control over response type (e.g. only return `.arrayBuffer()` in certain instances)

## How to Review

Tests should pass

## Checklist

- [x] Unit tests updated
- [x] README updated


packages/openapi-fetch/src/index.ts
```diff
@@ -54,6 +56,31 @@ export type FetchResponse<T> =
   | { data: T extends { responses: any } ? NonNullable<FilterKeys<Success<T["responses"]>, JSONLike>> : unknown; error?: never; response: Response }
   | { data?: never; error: T extends { responses: any } ? NonNullable<FilterKeys<Error<T["responses"]>, JSONLike>> : unknown; response: Response };
 
+/** Call URLSearchParams() on the object, but remove `undefined` and `null` params */
```
drwpow, line 59: Minor QoL improvement: yes, the library did use `new URLSearchParams()` as the default. But handling `undefined` / `null` params is annoying to have to implement one’s own `querySerializer` for.

packages/openapi-fetch/src/index.ts
```diff
@@ -54,6 +56,31 @@ export type FetchResponse<T> =
...
+    for (const [k, v] of Object.entries(q)) {
+      if (v === undefined || v === null) continue;
+      search.set(k, String(v));
+    }
+  }
+  return search.toString();
+}
+
+/** Construct URL string from baseUrl and handle path and query params */
+export function createFinalURL<O>(url: string, options: { baseUrl?: string; params: { query?: Record<string, unknown>; path?: Record<string, unknown> }; querySerializer: QuerySerializer<O> }): string {
```
drwpow, line 72: `createFinalURL()` is now exposed for testing (#1161)

PR #1175, opened 2023-06-17
Change .get('/post') as it’s confusing

drwpow:
## Changes

Docs-only change: In tests and docs, changes `.get('/post')` to `.get('/blogposts')` because the former is confusing to read for the first time.

## Checklist

- [x] Updated `examples` directory

### DIFF
--- a/packages/openapi-fetch/test/v1.yaml
+++ b/packages/openapi-fetch/test/v1.yaml
@@ -12,7 +12,20 @@ paths:
           $ref: '#/components/responses/CreateReply'
         500:
           $ref: '#/components/responses/Error'
-  /post:
+  /blogposts:
+    get:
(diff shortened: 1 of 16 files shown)


packages/openapi-fetch/test/v1.yaml
```diff
@@ -12,48 +12,35 @@ paths:
...
+            items:
+              type: string
       responses:
-        201:
-          $ref: '#/components/responses/CreatePost'
+        200:
+          $ref: '#/components/responses/AllPostsGet'
         500:
           $ref: '#/components/responses/Error'
-  /post/optional/inline:
```
drwpow, line 33: Renamed to `/blogposts-optional-inline` to be more REST-like

PR #1103, opened 2023-05-08
build: switch to tsup

thebuilder:
## Changes

Switch to tsup, and publish both cjs and esm files for `openapi-fetch`.
This fixes #1102 

## How to Review

Validate locally that the build output still works.

## Checklist

- [ ] Unit tests updated
- [ ] README updated
- [ ] `examples/` directory updated (if applicable)

PR #1176, opened 2023-06-20
build(openapi-fetch): bundle commonjs

### LINKED ISSUE #1102: [ERR_REQUIRE_ESM] with `openapi-fetch`
**Description**

The `openapi-fetch` currently exports a `index.js` and (`.index.min.js`) file, which might work fine if run in a `type: 'module'` environment.

However, this is something that will be bundled for the browser environment, which leads to the joy of bundling, and could also be bundled for CommonJS for the client.
Exporting both a CommonJS and ESM file, and having them indexed correctly in the `package.json` should help with this.

A package like [tsup](https://github.com/egoist/tsup) should improve it. I think a simple setup like this should be enough.

```json
{
  "main": "dist/index.js",
  "module": "dist/index.mjs",
  "types": "dist/index.d.ts",
  "files": [
    "dist"
  ],
  "scripts": {
    "build": "tsup src/index.tsx"
  }
}
```

## Error message

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Frontend/node_modules/.pnpm/openapi-fetch@0.1.2/node_modules/openapi-fetch/dist/index.js from /Frontend/dist/prerender/prerender.js not supported.
Instead change the require of index.js in /Frontend/dist/prerender/prerender.js to a dynamic import() which is available in all CommonJS modules.
```

**Checklist**

kecrily:
## Changes

Also to fix #1102, but unlike #1103, this PR reuses `esbuild` instead of adding a new dependency.

Just change the `--format` flag to bundle cjs.

Ref https://esbuild.github.io/api/#format-commonjs

## How to Review

Is this type of bundling acceptable?

## Checklist

N/A

- [ ] ~~Unit tests updated~~
- [ ] ~~README updated~~
- [ ] ~~`examples/` directory updated (only applicable for openapi-typescript)~~

## Other

If your project/package is being blocked by the lack of cjs distribution for openapi-fetch, you can temporarily try using [`@kecrily/openapi-fetch`](https://www.npmjs.com/package/@kecrily/openapi-fetch) based on this PR before this PR is merged.

### DIFF
--- a/packages/openapi-fetch/package.json
+++ b/packages/openapi-fetch/package.json
@@ -9,6 +9,15 @@
   "license": "MIT",
   "type": "module",
   "main": "./dist/index.js",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "require": "./dist/index.cjs",
+      "import": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    }
+  },
   "homepage": "https://openapi-ts.pages.dev",
   "repository": {
     "type": "git",
@@ -32,10 +41,11 @@
     "svelte"
   ],
   "scripts": {
-    "build": "pnpm run build:clean && pnpm run build:ts && pnpm run build:ts-min",
+    "build": "pnpm run build:clean && pnpm run build:ts && pnpm run build:ts-min && pnpm run build:cjs",
     "build:clean": "del dist",
     "build:ts": "tsc -p tsconfig.build.json",
     "build:ts-min": "esbuild --bundle src/index.ts --format=esm --minify --outfile=dist/index.min.js && cp dist/index.d.ts dist/index.min.d.ts",
+    "build:cjs": "esbuild --bundle src/index.ts --format=cjs --outfile=dist/index.cjs",
     "lint": "pnpm run lint:js",
     "lint:js": "prettier --check \"{src,test}/**/*\"",
     "test": "pnpm run test:ts && npm run test:js",
--- a/packages/openapi-fetch/package.json
+++ b/packages/openapi-fetch/package.json
@@ -16,7 +16,8 @@
       "require": "./dist/index.cjs",
       "import": "./dist/index.js",
       "types": "./dist/index.d.ts"
-    }
+    },
+    "*": "./*"
   },
   "homepage": "https://openapi-ts.pages.dev",
   "repository": {


packages/openapi-fetch/package.json
```diff
@@ -9,6 +9,15 @@
...
   "type": "module",
   "main": "./dist/index.js",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "require": "./dist/index.cjs",
+      "import": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    }
```
drwpow, line 19:
```suggestion
    },
    "*": "./*"
```

we’ll want to expose all files in this package like before to avoid breaking changes. by default, package imports make all non-explicit imports private

drwpow:
I’m on board for this addition, so long as we have the CLI tests testing this CommonJS bundle as well.

The major barrier for this existing in v6 wasn’t the actual bundling of it; it was maintenance, and getting bug reports for CommonJS that I don’t have time to support. Any way we can mitigate that by having the test suite ensure it works would help (FWIW, I don’t think the unit tests need to be duplicated; just the CLI tests expanded to test this bundle).

kecrily:
> so long as we have the CLI tests testing this CommonJS bundle as well.

What does the CLI test refer to here? It means to test if esbuild can bundle cjs properly?

> The major barrier for this existing in v6 wasn’t the actual bundling of it; it was maintenance, and getting bug reports for CommonJS that I don’t have time to support.

We use cjs in our projects, and if we find a problem, we will try to fix it and submit a PR if possible.

drwpow:
Oh the CLI test here: [packages/openapi-typescript/test/cli.test.ts](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-typescript/test/cli.test.ts)

Should be as simple as duplicating those tests and setting `cmd` to the CJS build (but obviously keep the existing build tests in place).

> We use cjs in our projects, and if we find a problem, we will try to fix it and submit a PR if possible.

That would be enormously helpful, thank you 🙂

kecrily:
> Should be as simple as duplicating those tests and setting cmd to the CJS build (but obviously keep the existing build tests in place).

This PR is not about `openapi-typescript`. this cjs test seems to be non-existent. Also I don't think `openapi-typescript` should distribute a cjs version, since it is usually only used on the command line. (Of course if you think this is necessary, I'd be happy to modify it together)

drwpow:
🤦 Ah I’m so sorry—I’m still getting used to these projects being merged together and I was hopping back-and-forth between package conversations today. This looks great! Sorry for my confusion.

drwpow (approved):

packages/openapi-fetch/package.json
```diff
@@ -9,6 +9,16 @@
   "license": "MIT",
   "type": "module",
   "main": "./dist/index.js",
+  "module": "./dist/index.js",
```
drwpow, line 12: Also I’m fine with adding this, but this is more of an “invalid” / “legacy” property which was never officially standardized, which was why it was missing in the first place. It’s also not needed because it’s redundant with `"main"`, but it’s also fine to include just in case it helps something.

drwpow:
Released in `v0.4.0` :tada. Thank you!

dino-rodriguez:
@drwpow How do I actually get the cjs output to be created? When I install the module using `yarn` and use it in a script, it can't find the `.cjs` files because they don't seem to exist when I try to execute the script:

```
Error: Cannot find module '/Users/dino/Code/indie/api/client/node_modules/openapi-fetch/dist/index.cjs'
    at createEsmNotFoundErr (node:internal/modules/cjs/loader:1016:15)
    at finalizeEsmResolution (node:internal/modules/cjs/loader:1009:15)
    at resolveExports (node:internal/modules/cjs/loader:529:14)
    at Function.Module._findPath (node:internal/modules/cjs/loader:569:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:981:27)
    at Function.Module._resolveFilename.sharedData.moduleResolveFilenameHook.installedValue [as _resolveFilename] (/Users/dino/.volta/tools/image/packages/ts-node/lib/node_modules/@cspotcode/source-map-support/source-map-support.js:811:30)
    at Function.Module._load (node:internal/modules/cjs/loader:841:27)
    at Module.require (node:internal/modules/cjs/loader:1061:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/Users/dino/Code/indie/api/client/client.ts:1:1) {
  code: 'MODULE_NOT_FOUND',
  path: '/Users/dino/Code/indie/api/client/node_modules/openapi-fetch/package.json'
  ```

Confirmed on the latest version.

drwpow:
Ah sorry—this is a problem with the automated releases; not a problem with this PR at all. Will fix.

drwpow:
Thanks for bearing with me on the automated releases. It’s a PITA to set up, and there are usually bugs like this for the first month or so. But long-term it’s much more sustainable. And improves the quality of the library because maintainers can release as well.

dino-rodriguez:
@drwpow No worries at all - thank you for responding so quickly! Hugely appreciate it

PR #1186, opened 2023-06-26
Fix monorepo release build

### LINKED ISSUE #1185: CJS not in bundle after latest PR
The `.cjs` module is not present for openapi-fetch after fetching the module with `npm` and using it. See my comment on the PR here: https://github.com/drwpow/openapi-typescript/pull/1176#issuecomment-1608424912

drwpow:
## Changes

Fixes #1185. The build triggered by the autorelease is different than the `build` script dictated by each package. While `openapi-typescript` was unaffected, `openapi-fetch` had missing files in the latest release.

## How to Review

Will have to release to check 🤞

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
   },
   "scripts": {
     "build": "run-p -s build:*",
-    "build:clean": "del packages/openapi-typescript/dist && del packages/openapi-fetch/dist",
-    "build:ts": "tsc -b packages/openapi-typescript/tsconfig.build.json packages/openapi-fetch/tsconfig.build.json",
+    "build:openapi-typescript": "cd packages/openapi-typescript && pnpm run build",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -9,8 +9,8 @@
   },
   "scripts": {
     "build": "run-p -s build:*",
-    "build:clean": "del packages/openapi-typescript/dist && del packages/openapi-fetch/dist",
-    "build:ts": "tsc -b packages/openapi-typescript/tsconfig.build.json packages/openapi-fetch/tsconfig.build.json",
```
drwpow, line 13: `tsc -b` is a nice little optimization, but when we’re only building 2 packages, this won’t have much impact at all. Just let each package decide how it builds, even if it’s a second or two slower.

PR #1192, opened 2023-07-01
fix(openapi-fetch): fix headers for FormData

### LINKED ISSUE #1191: `multipart/form-data` does not allow browser to set boundary expression
**Description**

`multipart/form-data` content type requires a boundary expression to serve as a form field delimiter in `body`. 

[If you set `Content-Type` to `multipart/form-data` explicitly, the browser does not set boundary](https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects#sending_files_using_a_formdata_object).

`openapi-fetch` either sets `application/json` or uses whatever the user provided, so it's currently not possible for it to allow the browser to correctly set the boundary.

**Reproduction**

I can create one if really needed

**Checklist**

@nilsso

psychedelicious:
We need to delete the `Content-Type` header if the serialized body is a `FormData`. This allows the browser to set the `Content-Type` and boundary expression correctly.

Resolves #1191

## Changes

Because `FormData` gets special handling by the browser, we need to give it special handling too. 

Check if the serialized body is an instance of `FormData`, and if so, delete the `Content-Type` header. 

Resolves #1191 

## How to Review

I've tested this in my production app but I think you'd need to whip up a simple test project to see for yourself. I'm out of time for that today though, sorry!

One hiccup in the tests - `vitest-fetch-mock` does not add the form field boundary like the browser does. Ideally we could confirm that the `Content-Type` for `FormData` bodies are correct:
```ts
// expect requests with FormData to have correct Content-Type with boundary expression
expect((req.headers as Headers).get("Content-Type")).toMatch(/^multipart\/form-data; boundary=.*/);
```

Instead, we must check if it is null:
```ts
// `vitest-fetch-mock` does not add the boundary to the Content-Type header like browsers do, so we expect the header to be null expect((req.headers as Headers).get("Content-Type")).toBeNull();
```

@nilsso Please give try this out.

## Checklist

- [x] Unit tests updated
- [x] README updated


packages/openapi-fetch/src/index.ts
```diff
@@ -122,9 +122,11 @@ export default function createClient<Paths extends {}>(clientOptions: ClientOpti
       redirect: "follow",
       ...options,
       ...init,
-      headers: baseHeaders,
```
drwpow, line 125: I don’t think this line needs to be deleted; I believe the reference is the same either way so the `baseHeaders.delete` will still work, no?

drwpow:
Thanks for opening this PR! I’m in favor of merging this PR. Just had a minor nitpick on one unneeded change before approving and merging though. Otherwise looks great

packages/openapi-fetch/src/index.ts
psychedelicious: Yep! I've undone that and confirmed it works as intended.

drwpow (approved):

PR #1193, opened 2023-07-01
Update README.md with `transform` example for `Blob`

psychedelicious:
Add example `transform` callback for file uploads, where the `multipart/form-data` has a field that should be typed as `Blob`.

## Changes

Update the README with additional example for what I think is a common use-case.

## How to Review

Maybe this is extraneous? Dunno. I see the `stripe` API in `examples` has some schemas that should have this transformation done.

Should this transformation be the default behaviour?

## Checklist

- ~~[ ] Unit tests updated~~
- [x] README updated

### DIFF
--- a/.changeset/nasty-windows-chew.md
+++ b/.changeset/nasty-windows-chew.md
@@ -1,0 +1,6 @@
+---
+"openapi-typescript": patch
+"openapi-typescript-docs": patch
+---
+
+Add example for `Blob` type transforms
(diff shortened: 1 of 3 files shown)


drwpow:
Thanks for adding! If you could, this would probably be great to add somewhere in the `docs/` folder as well. Maybe in the Advanced docs (not too particular on the placement; it may be shuffled around, but just having it _somewhere_ makes it more searchable)

drwpow (approved):

psychedelicious:
Somehow I missed that `docs/` and the docs site exist - happy to update those.

drwpow (approved):

drwpow:
> Somehow I missed that `docs/` and the docs site exist - happy to update those.

No worries. The docs are relatively new and still need some  refinement (any suggestions are welcome!). I didn’t want to completely delete the README until the docs were further along; sorry for the confusion there.

PR #1195, opened 2023-07-02
docs(openapi-ts): add docs for using Node.js API in web projects

psychedelicious:
## Changes

Update docs with examples for using the Node.js API in a project with a browser build target. 

## How to Review

I'm not sure if this deserved to be in the docs as it's only tangential to `openapi-typescript`, and I imagine most devs know how to set this up already. 

I, however, didn't know, and it took me an hour of going down stack-overflow garden paths before discovering `ts-node --esm`, so I figured I'd propose the changes.

## Checklist

- ~~[ ] Unit tests updated~~
- [x] README updated

### DIFF
--- a/.changeset/quick-parrots-worry.md
+++ b/.changeset/quick-parrots-worry.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript-docs": patch
+---
+
+Add docs for using Node.js API in web projects
--- a/docs/src/content/docs/node.md
+++ b/docs/src/content/docs/node.md
@@ -76,3 +76,72 @@ That would result in the following change:
 Any [Schema Object](https://spec.openapis.org/oas/latest.html#schema-object) present in your schema will be run through this formatter (even remote ones!). Also be sure to check the `metadata` parameter for additional context that may be helpful.
 
 There are many other uses for this besides checking `format`. Because this must return a **string** you can produce any arbitrary TypeScript code you’d like (even your own custom types).
+
+## Node.js API with browser build target
+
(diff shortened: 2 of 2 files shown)


drwpow (approved):

drwpow:
I’ll admit I actually didn’t read these too carefully before approving—I assumed they were related to #1193 which I’m fine with some redundancy. But I’m not thinking they add pertinent info to the Node.js API.

To be more specific:

- The term “browser build target” doesn’t mean something universally; this looks more like confusion over CJS required for a webpack config vs running Node in ESM (which is already “browser build target”-ready).
- `ts-node` is **NOT** required to run this script. Only [modern Node in ESM mode](https://nodejs.org/api/esm.html). In fact, I generally would not recommend people run `ts-node` for… anything, really. If people need a TS runtime, `vite-node` is probably better overall. But again, since this isn’t needed, I’d rather recommend _nothing_
- `{ "type": "module" }` _is_ actually a good callout, however, this is not “browser build target”; this is just a necessary requirement of running this library.

**Proposal**
- The critical `{ "type": "module" }` (ESM mode) should be specified **at the top of the page** in necessary setup. Probably as a **warning callout** (I think I’ve just been using blockquotes with the ⚠️ emoji).
- There should be a separate section for CJS projects just saying it’s not supported. Or if it is, I’d probably recommend `vite-node` as a last resort.

I’d be happy to accept a PR with those changes, if you’d like!

Also relevant to #1217

psychedelicious:
@drwpow 
Thanks for that explanation. I'm not sure why, but in the app in which I was using `openapi-typescript`, I had to do all the things I added to the docs via this PR. I suppose there is some unique configuration in our app.

I do intend to fix the docs per your advice, but am in the middle of a big launch. I won't have time until next weekend.

Issue #1197, opened 2023-07-04
feat: document polymorphic [`null`, `string`] type non spec-compliant output


toomuchdesign:
**Description**

I've recently run into a probably known issue related to polymorphic object definitions returning an unexpected empty string value in the resulting union type:

```json
{
  "properties": {
    "petType": {
      "type": [
        "null",
        "string"
      ],
      "enum": [
        "option1",
        "option2",
        null
      ]
    }
  }
}
```
...resulting into:

```ts
petType?: "option1" | "option2" | "" | null; // "" should not be there
```
I've found this behaviour documented as expected in the tests:
https://github.com/drwpow/openapi-typescript/blob/c732542df136eee2f41d242321758a0901b45248/packages/openapi-typescript/test/schema-object.test.ts#L283

**Proposal**

I'm writing to ask:
- Is this behaviour intended?
- Why tests mention a "no fix"? is it so because of a technical limitation or any other reason?
- How do you suggest to deal with this unexpected output (patching nested structures of objects/arrays in TS is quite hard)
- In case fixing this is not possible, I'd be happy to document this somewhere.

Happy to open a PR in case maintainers decided to consider fixing this, instead.

Thanks for the outstanding work here!

**Checklist**

- [X] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/CONTRIBUTING.md))

tyronen9:
> As a temporary workaround, again, `nullable: true` is probably preferred in general for 3.x schemas.

[It's not](https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0):

> In line with JSON Schema, the `type` keyword can now define multiple types for a schema with an array. This is useful new functionality, but has also made nullable redundant. In order to achieve the goal of letting JSON Schema tools understand OpenAPI, it was decided to remove nullable entirely instead of deprecate it. 
```
# OpenAPI v3.0
type: string
nullable: true

# OpenAPI v3.1
type:
- "string"
- "null" 
```

However, if using openapi-typescript, this changes generated code from:

```
fieldname: string | null
```

to 

```
fieldname: string | Record<string, never>
```

We may want a separate issue for this, but why does `null` map to `Record<XX, never>` instead of simply `null` in the first place? It means I have to change type definitions all over my codebase to migrate to 3.1.

toomuchdesign:
[...]

PR #1200, opened 2023-07-05
fix: empty strings in enums when type [string, null]

toomuchdesign:
## Changes

Related to: #1197

JSON schemas like:
```json
{
  "properties": {
    "petType": {
      "type": [
        "null",
        "string"
      ],
      "enum": [
        "foo", "bar"
      ]
    }
  }
}
```
..results into:
```ts
{
  petType: "foo" | "bar" | ""
}
```
...which seems to add an unexpected empty string to the union. See the example in [this JSON schema validator](https://www.jsonschemavalidator.net/s/4oHwGoE7).

With this PR the empty `""` is removed from the union:
```ts
{
  petType: "foo" | "bar"
}
```

Happy to review the PR based on your suggestions!

### `null` not listed as enum

In case type contains `null` but then `null` is not listed as enum:

```json
{
  "properties": {
    "petType": {
      "type": [
        "null",
        "string"
      ],
      "enum": [
        "foo", "bar"
      ]
    }
  }
}
```
..I understand the result should not contain `null`. See [this online validator test case](https://www.jsonschemavalidator.net/s/v1I0MwYP).

## How to Review

Take a look at the new test cases added. The fix didn't introduce any regressive test error.

## Checklist

- [X] Unit tests updated
- [ ] README updated

### DIFF
--- a/packages/openapi-typescript/src/transform/schema-object.ts
+++ b/packages/openapi-typescript/src/transform/schema-object.ts
@@ -51,13 +51,13 @@ export function defaultSchemaObjectTransform(schemaObject: SchemaObject | Refere
   if (schemaObject.enum) {
     let items = schemaObject.enum as any[];
     if ("type" in schemaObject) {
-      if (schemaObject.type === "string" || (Array.isArray(schemaObject.type) && schemaObject.type.includes("string"))) items = items.map((t) => escStr(t || "")); // empty/missing values are empty strings
+      if (schemaObject.type === "string" || (Array.isArray(schemaObject.type) && schemaObject.type.includes("string"))) items = items.map((t) => escStr(t)); // empty/missing values are empty strings
     }
     // if no type, assume "string"
     else {
       items = items.map((t) => escStr(t || ""));
     }
-    return tsUnionOf(...items, ...(schemaObject.nullable || ("type" in schemaObject && Array.isArray(schemaObject.type) && schemaObject.type.includes("null")) ? ["null"] : []));
+    return tsUnionOf(...items, ...(schemaObject.nullable ? ["null"] : []));
   }
 
   // oneOf (no discriminator)
--- a/packages/openapi-typescript/src/transform/schema-object.ts
+++ b/packages/openapi-typescript/src/transform/schema-object.ts
@@ -51,7 +51,7 @@ export function defaultSchemaObjectTransform(schemaObject: SchemaObject | Refere
   if (schemaObject.enum) {
     let items = schemaObject.enum as any[];
     if ("type" in schemaObject) {
-      if (schemaObject.type === "string" || (Array.isArray(schemaObject.type) && schemaObject.type.includes("string"))) items = items.map((t) => escStr(t)); // empty/missing values are empty strings
+      if (schemaObject.type === "string" || (Array.isArray(schemaObject.type) && schemaObject.type.includes("string"))) items = items.map((t) => escStr(t));
     }
     // if no type, assume "string"
     else {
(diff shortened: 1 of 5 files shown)


packages/openapi-typescript/src/transform/schema-object.ts
```diff
@@ -51,13 +51,13 @@ export function defaultSchemaObjectTransform(schemaObject: SchemaObject | Refere
   if (schemaObject.enum) {
     let items = schemaObject.enum as any[];
     if ("type" in schemaObject) {
-      if (schemaObject.type === "string" || (Array.isArray(schemaObject.type) && schemaObject.type.includes("string"))) items = items.map((t) => escStr(t || "")); // empty/missing values are empty strings
+      if (schemaObject.type === "string" || (Array.isArray(schemaObject.type) && schemaObject.type.includes("string"))) items = items.map((t) => escStr(t)); // empty/missing values are empty strings
```
drwpow, line 54: 👍 Nice cleanup here. I think this was a byproduct of being written before `tsUnionOf()` deduplicated things.

packages/openapi-typescript/test/schema-object.test.ts
```diff
@@ -278,12 +278,19 @@ describe("Schema Object", () => {
...
+        const generated = transformSchemaObject({ type: ["string", "null"], enum: ["false positive", "won't fix", "used in tests"] }, options);
+        expect(generated).toBe(`"false positive" | "won't fix" | "used in tests"`);
+      });
+
+      test("enum + polymorphism + nullable 2", () => {
+        const generated = transformSchemaObject({ type: ["string", "null"], enum: ["", "false positive", "won't fix", "used in tests"] }, options);
+        expect(generated).toBe(`"" | "false positive" | "won't fix" | "used in tests"`);
+      });
+
+      test("enum + polymorphism + nullable 3", () => {
```
drwpow, line 291: Awesome tests

drwpow (approved):

drwpow:
Could you add a changeset so the changelog is updated (details in comment)? 🙏

toomuchdesign:
Thank you @drwpow!

A few more questions:
- shall I change the test case so that they don't use the confusing `"false positive" | "won't fix" | "used in tests"` values?
- is `// empty/missing values are empty strings` comment still relevant?
- shall I generate a changeset entry in this same pr?

Cheers!

toomuchdesign:
Thank you @drwpow!
Cleaned up my PR, rebased and and added the relevant changelog.

Cheers!

packages/openapi-typescript/test/schema-object.test.ts
```diff
@@ -278,9 +278,19 @@ describe("Schema Object", () => {
         expect(generated).toBe("OneOf<[string, boolean, number, null]>");
       });
 
-      test("enum + polymorphism + nullable", () => {
-        const generated = transformSchemaObject({ type: ["string", "null"], enum: ["", "false positive", "won't fix", "used in tests"] }, options);
-        expect(generated).toBe(`"" | "false positive" | "won't fix" | "used in tests" | null`);
```
drwpow, line 283: Yeah thanks for changing these. I don’t think the old strings were meaningful / helpful.

drwpow:
The test failures may be from the snapshots failing. In the `packages/openapi-typescript` directory, try running `pnpm run build && pnpm run update:examples` to regenerate those.

The diff that shows is also helpful to see what it does on large real-world schemas, and they’re helpful in addition to the ever-essential unit tests.

toomuchdesign:
All updated 👍 I've seen a few `""` being removed :)

drwpow:
> All updated 👍 I've seen a few `""` being removed :)

Yup those look like good diffs to me 🙂

PR #1202, opened 2023-07-06
Ignore docs from changesets

drwpow:
## Changes

Changelogs don’t need to be generated for the documentation (version doesn’t need to be bumped)

## How to Review

N/A

## Checklist
N/A

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,10 +1,10 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
   "changelog": ["@changesets/changelog-github", { "repo": "drwpow/openapi-typescript" }],
+  "ignore": ["openapi-typescript-docs"],
   "commit": false,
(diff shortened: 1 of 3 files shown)


.changeset/config.json
```diff
@@ -1,10 +1,10 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@2.3.0/schema.json",
   "changelog": ["@changesets/changelog-github", { "repo": "drwpow/openapi-typescript" }],
+  "ignore": ["openapi-typescript-docs"],
   "commit": false,
   "fixed": [],
   "linked": [],
   "access": "public",
-  "baseBranch": "main",
-  "updateInternalDependencies": "patch"
```
drwpow, line 9: Can be removed, as this currently blocks releases. `openapi-fetch` doesn’t rely on `openapi-typescript` except as a devDep for testing. Release-wise it’s safe to be out-of-date

PR #1203, opened 2023-07-06
Fix mutating $refs in Node.js API

### LINKED ISSUE #1178: openapiTS mutates the schema object
**Description**

`openapiTS` mutates the schema object. (`$ref`'s mainly, it appears)

**Reproduction**

* Serialize schema with `JSON.stringify()`
* Pass schema with components and refs.
* Serialize schema with `JSON.stringify()` again
* Compare serializations to confirm transformation

```ts
openapiTS = await import('openapi-typescript').then((module) => module.default)

schema = {
  openapi: '3.0.2',
  info: {
    title: 'My Schema',
    description: 'This is a very fancy service.',
    version: '1.0',
    contact: {}
  },
  tags: [],
  paths: {
    '/{id}': {
      get: {
        parameters: [
          { name: 'id', in: 'path', schema: { $ref: '#/components/schemas/ObjectId' }, required: true }
        ],
        responses: {
          '204': {
            description: 'OK'
          }
        }
      }
    },
  },
  components: {
    schemas: {},
    links: {},
    callbacks: {}
  }
}

before = JSON.stringify(schema);
await openapiTS(schema)
after = JSON.stringify(schema);

same = before === after // false
```

**Expected result**

* API / type signatures accept readonly schemas
* Don't mutate the schema.

**Checklist**

drwpow:
## Changes

Fixes #1178.

## How to Review

N/A

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/packages/openapi-typescript/src/load.ts
+++ b/packages/openapi-typescript/src/load.ts
@@ -173,7 +173,7 @@ export default async function load(schema: URL | Subschema | Readable, options: 
   else if (typeof schema === "object") {
     options.schemas[schemaID] = {
       hint: "OpenAPI3",
-      schema: schema as any,
+      schema: JSON.parse(JSON.stringify(schema)), // create deep clone of inline schema (don’t mutate)
     };
   }
   // 1d. failsafe
--- a/.changeset/fresh-dancers-joke.md
+++ b/.changeset/fresh-dancers-joke.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Fix mutating $refs in Node.js API
--- a/packages/openapi-typescript/test/index.test.ts
+++ b/packages/openapi-typescript/test/index.test.ts
@@ -189,10 +189,7 @@ export type operations = Record<string, never>;
           "/post/{id}": {
             get: {
               operationId: "getPost",
-              parameters: [
-                { name: "format", in: "query", schema: { type: "string" } },
-                { $ref: "#/components/parameters/post_id" },
-              ],
+              parameters: [{ name: "format", in: "query", schema: { type: "string" } }, { $ref: "#/components/parameters/post_id" }],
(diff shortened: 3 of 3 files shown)


packages/openapi-typescript/src/load.ts
```diff
@@ -173,7 +173,7 @@ export default async function load(schema: URL | Subschema | Readable, options:
   else if (typeof schema === "object") {
     options.schemas[schemaID] = {
       hint: "OpenAPI3",
-      schema: schema as any,
+      schema: "structuredClone" in globalThis ? structuredClone(schema as any) : schema, // TODO: `"structuredClone" in globalThis ?` only temporarily needed until Node 16 is EOL on 2023-09-11
```
drwpow, line 176: I figured adding lodash, or some other heavy-weight clone tool is unnecessary just for Node 16, since that will be EOL in [a matter of weeks](https://nodejs.org/en/blog/announcements/nodejs16-eol). Until then, this just means Node 16 mutates the original reference and doesn’t receive this bugfix.

packages/openapi-typescript/src/load.ts
```diff
@@ -173,7 +173,7 @@ export default async function load(schema: URL | Subschema | Readable, options:
   else if (typeof schema === "object") {
     options.schemas[schemaID] = {
       hint: "OpenAPI3",
-      schema: schema as any,
+      schema: JSON.parse(JSON.stringify(schema)), // create deep clone of inline schema (don’t mutate)
```
drwpow, line 176: So funny enough, after some lazy performance testing in V8, `JSON.parse(JSON.stringify(…))` was faster than `structuredClone()`. Which I guess makes sense with JSON-serializable data that we’re working with.

This is probably not that much of a hack as I originally thought it was. And this only runs in the Node.js API (only when needed)

PR #1212, opened 2023-07-07
Fix remote schema maps

### LINKED ISSUE #1131: Generation of schema that references outside schema
**Description**

Generation of a schema, that references other schemas in the $ref results in a invalid schema.ts, with "unkown" types and wrong property referencing `Property 'Problem' does not exist on type 'unknown'`

| `openapi-typescript` | `6.2.4`                      |
| Node.js              | `20.1.0`                      |

**Reproduction**

Run following code as a .yml in the generator:
```yml
openapi: 3.0.0
servers:
  - url: https://dummy.to.server/v1
info:
  description: Test YAML
  version: 0.1.0
  title: Test
  termsOfService: ''
security:
  - bearerAuth: [ ]
tags:
  - name: Test Object
    description: Basic operations on a test object
paths:
  /test-object/{uuid}:
    get:
      summary: Returns a test object
      tags:
        - Test Object
      description: Returns a single test object if exists
      parameters:
        - name: uuid
          in: path
          required: true
          schema:
            type: string
      operationId: getLatestTOByUUID
      responses:
        '200':
          description: Test Object
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/TestObject'
        '404':
          description: Test Object not found
          content:
            application/problem+json:
              schema:
                $ref: 'https://opensource.zalando.com/restful-api-guidelines/models/problem-1.0.0.yaml#/Problem'
        default:
          description: Unexpected error
          content:
            application/problem+json:
              schema:
                $ref: 'https://opensource.zalando.com/restful-api-guidelines/models/problem-1.0.0.yaml#/Problem'
components:
  schemas:
    TestObject:
      type: object
      properties:
        uuid:
          type: string
        id:
          type: integer
  securitySchemes:
    bearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
```

Generate a schema based on the prior .yml file.

**Expected result**

_(in case it’s not obvious)_

**Checklist**

drwpow:
## Changes

Fixes #1131. There are likely some other bugs with deeply-nested remote schemas, but this fixes at least one.

The reason was in our `getHint()` logic that traverses the nodePath, it wasn’t keeping track of remote filenames, which is necessary for certain inferences (the hints are necessary because partial schemas could be _anything_ and still be valid—the only way to know is by tracing from the root). There’s probably a better solution than this, but this will do for now.

## How to Review

Tests should pass.

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/packages/openapi-typescript/examples/digital-ocean-api.ts
+++ b/packages/openapi-typescript/examples/digital-ocean-api.ts
@@ -3965,7 +3965,160 @@ export interface external {
       default: external["shared/responses/unexpected_error.yml"];
     };
   }
-  "resources/certificates/models/certificate_create.yml": unknown
+  "resources/certificates/models/certificate_create.yml": {
+    certificate_create_base: {
+      /**
+       * @description A unique human-readable name referring to a certificate. 
+       * @example web-cert-01
+       */
+      name: string;
+      /**
+       * @description A string representing the type of the certificate. The value will be `custom` for a user-uploaded certificate or `lets_encrypt` for one automatically generated with Let's Encrypt. 
+       * @example lets_encrypt 
+       * @enum {string}
+       */
+      type?: "custom" | "lets_encrypt";
+    };
+    /** Custom Certificate Request */    certificate_request_custom: certificate_create_base[] & {
+      /**
+       * @description The contents of a PEM-formatted private-key corresponding to the SSL certificate. 
+       * @example -----BEGIN PRIVATE KEY-----
+       * MIIEvgIBADANBgkqhkiG9w0BAQEFAASCBKgwggSkAgEAAoIBAQDBIZMz8pnK6V52
+       * SVf+CYssOfCQHAx5f0Ou5rYbq3xNh8VHAIYJCQ1QxQIxKSP6+uODSYrb2KWyurP1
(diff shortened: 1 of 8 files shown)


packages/openapi-typescript/examples/digital-ocean-api.ts
```diff
@@ -3965,7 +3965,160 @@ export interface external {
       default: external["shared/responses/unexpected_error.yml"];
     };
   }
-  "resources/certificates/models/certificate_create.yml": unknown
```
drwpow, line 3968: 🤔 this is probably a good sign

packages/openapi-typescript/src/load.ts
```diff
@@ -306,105 +309,116 @@ function relativePath(src: URL, dest: URL): string {
...
+function getHintFromSchemaObject(path: (string | number)[], external: boolean): Subschema["hint"] {
   switch (path[0]) {
     case "allOf":
     case "anyOf":
     case "oneOf":
-      return getHintFromSchemaObject(path.slice(2)); // skip array index at [1]
+      return getHintFromSchemaObject(path.slice(2), external); // skip array index at [1]
+  }
+  // if this is external, and the path is [filename, key], then the external schema is probably a SchemaMap
+  if (path.length === 2 && external) {
```
drwpow, line 419: Very wacky logic, but it seems to check out 🤷

PR #1219, opened 2023-07-10
Update README.md

metreniuk:
## Changes

Separate local and remote schema examples so the github copy button works for both properly. Now if you try to copy it you get the local **and** the remote commands

## How to Review

Preview the README.md

## Checklist

- [x] README updated


drwpow (approved):
Fine with me

PR #1221, opened 2023-07-10
Fix 3.1 nullable types

### LINKED ISSUE #1215: Support null in OpenAPI 3.1 union types
**Description**

OpenAPI 3.1 [dropped support for nullable](https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0). Instead, callers are expected to use array types:

```
# OpenAPI v3.0
type: string
nullable: true

# OpenAPI v3.1
type:
- "string"
- "null" 
```

**Reproduction**

For a type like the one above, the generated type is `OneOf<[string, Record<string, never>]>`. 

As far as I can tell, `Record<string, never>` is returned when the tool cannot tell what type it should be generating for.

**Expected result**

The generated type should be `OneOf<[string, null]>`

**Checklist**

- [X] My OpenAPI schema passes the [Redocly validator](https://redocly.com/docs/cli/commands/lint/) (`npx @redocly/cli@latest lint`)
- [X] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/CONTRIBUTING.md))

drwpow:
## Changes

Fixes #1215, #1197. Adds better tests for 3.1 nullable types and improves handling. Also removes the `OneOf<>` helper for simple types.

## How to Review

Tests should pass.

## Checklist

- [x] Unit tests updated
- [x] README updated


packages/openapi-typescript/src/transform/schema-object.ts
```diff
@@ -126,8 +121,14 @@ export function defaultSchemaObjectTransform(schemaObject: SchemaObject | Refere
         itemType = tsArrayOf(itemType);
       }
       itemType = ctx.immutableTypes || schemaObject.readOnly ? tsReadonly(itemType) : itemType;
+
       return schemaObject.nullable ? tsUnionOf(itemType, "null") : itemType;
     }
+
+    // polymorphic, or 3.1 nullable
+    if (Array.isArray(schemaObject.type)) {
```
drwpow, line 129: moving this is arbitrary; just a little easier to reason with.

packages/openapi-typescript/src/transform/schema-object.ts
```diff
@@ -126,8 +121,14 @@ export function defaultSchemaObjectTransform(schemaObject: SchemaObject | Refere
...
+      return tsUnionOf(...schemaObject.type.map((t) => transformSchemaObject({ ...schemaObject, type: t }, { path, ctx })));
```
drwpow, line 130: Big change: `tsUnionOf` rather than `tsOneOf` here is more correct, and now handles 3.1 nullable types correctly. There shouldn’t be any performance impact from this.

packages/openapi-typescript/test/schema-object.test.ts
```diff
@@ -49,12 +49,6 @@ describe("Schema Object", () => {
   status?: "complete" | "incomplete";
 }`);
       });
-
-      test("nullable", () => {
```
drwpow, line 53: Removed the disparate nullable tests and grouped them together (which helped identify several missing test cases)

PR #1231, opened 2023-07-20
Do not append trailing spaces to JSDoc tags

tkrotoff:
Before:

```
/**
 * A title <== Trailing space here
 * @description A description
 */
```

After:

```
/**
 * A title <== No trailing space
 * @description A description
 */
```


packages/openapi-typescript/src/utils.ts
```diff
@@ -55,13 +55,13 @@ export function getSchemaObjectComment(v: CommentObject, indentLv?: number): str
   const output: string[] = [];
 
   // * Not JSDOC tags: [title, format]
-  if (v.title) output.push(`${v.title} `);
-  if (v.summary) output.push(`${v.summary} `);
-  if (v.format) output.push(`Format: ${v.format} `);
+  if (v.title) output.push(v.title);
+  if (v.summary) output.push(v.summary);
+  if (v.format) output.push(`Format: ${v.format}`);
```
tkrotoff, line 60: The `comment` function already does the job of adding a trailing space in case of a single-line comment

drwpow (approved):
This is such a small detail but has a really big impact! Thank you for fixing this.

packages/openapi-typescript/package.json
```diff
@@ -41,10 +41,10 @@
     "build": "del dist && tsc -p tsconfig.build.json",
     "dev": "tsc -p tsconfig.build.json --watch",
     "download:schemas": "vite-node ./scripts/download-schemas.ts",
-    "format": "pnpm run prettier -w .",
+    "format": "prettier --write 'src/**/*'",
     "lint": "run-p -s lint:*",
-    "lint:js": "eslint \"src/**/*.{js,ts}\"",
-    "lint:prettier": "prettier --check \"src/**/*\"",
+    "lint:js": "eslint 'src/**/*.{js,ts}'",
```
drwpow, line 46:
```suggestion
    "lint:js": "eslint \"src/**/*.{js,ts}\"",
```

We actually want to keep the escaped double quotes. Though *nix and mac users can run this just fine, IIRC Windows contributors run into problems with single quotes.

packages/openapi-typescript/src/utils.ts
```diff
@@ -55,13 +55,13 @@ export function getSchemaObjectComment(v: CommentObject, indentLv?: number): str
   const output: string[] = [];
 
   // * Not JSDOC tags: [title, format]
-  if (v.title) output.push(`${v.title} `);
-  if (v.summary) output.push(`${v.summary} `);
```
drwpow, line 59: I think some of these string escapes mostly came from fields that could contain non-strings, but I think that was really only `example`. I don’t see the harm in removing the string escapes.

packages/openapi-typescript/test/utils.test.ts
```diff
@@ -112,4 +112,60 @@ describe("utils", () => {
       expect(escObjKey("_ref_")).toStrictEqual("_ref_");
     });
   });
+
+  describe("comment", () => {
```
drwpow, line 116: Great tests 👏

packages/openapi-typescript/package.json
drwpow: Since this is such a minor thing, I’ll revert this after merge, but won’t touch anything else. Thanks!

tkrotoff: Sorry for that I always work on a UNIX OS and though the scripts without escaped quotes look nicer.

drwpow: oh no worries at all. I did the same thing until I had Windows users complain 😛 (and I agree with you)

drwpow: oh no worries at all. I did the same thing until I had Windows users complain 😛 (and I agree with you)

PR #1236, opened 2023-07-21
Improve handling of oneOf and enum

drwpow:
## Changes

Before, if `oneOf` or `enum` was present, we ignored the core type (for objects, at least). In this PR, we now create an intersection of the core type + `oneOf`/`enum`.

The reason for the `enum` handling was although unconventional, I did find usage of it in the GitHub API schemas. And I couldn’t find any explicit signal that this is invalid. So just for object types, we’ll treat `enum == oneOf`.

## How to Review

Tests should pass.

## Checklist

- [x] Unit tests updated


packages/openapi-typescript/examples/github-api.ts
```diff
@@ -13321,7 +13323,7 @@ export interface components {
      * Organization ruleset conditions
      * @description Conditions for an organization ruleset
      */
-    "org-ruleset-conditions": (components["schemas"]["repository-ruleset-conditions"] & components["schemas"]["repository-ruleset-conditions-repository-name-target"]) | (components["schemas"]["repository-ruleset-conditions"] & components["schemas"]["repository-ruleset-conditions-repository-id-target"]);
+    "org-ruleset-conditions": Record<string, never> & ((components["schemas"]["repository-ruleset-conditions"] & components["schemas"]["repository-ruleset-conditions-repository-name-target"]) | (components["schemas"]["repository-ruleset-conditions"] & components["schemas"]["repository-ruleset-conditions-repository-id-target"]));
```
drwpow, line 13326: Fortunately the `Record<string, never>` doesn’t interfere with the type at all; it’s just an empty object that gets ignored

packages/openapi-typescript/examples/github-api-next.ts
```diff
@@ -90333,15 +90335,97 @@ export interface operations {
     };
     requestBody: {
       content: {
-        "application/json": OneOf<[{
+        "application/json": ({
```
drwpow, line 90338: This is a big improvement, and a scenario where maybe we should start preferring practice over theory

packages/openapi-typescript/src/types.ts
```diff
@@ -447,18 +447,21 @@ export type SchemaObject = {
 
 export interface StringSubtype {
   type: "string";
+  enum?: (string | ReferenceObject)[];
```
drwpow, line 450: I’m not completely happy with the `enum` type changes; we may need to change these in the future. But it just made this change easier.

PR #1239, opened 2023-07-25
Fix params.header missing in request

### LINKED ISSUE #1230: Parmeters set with in: header are not added to headers
**Description**

When you have a parameter in your openAPI spec with in: header, it is not added to the headers in the request

**Reproduction**

openapi spec like this:
```
(...)
paths:
  "/api/v1/register":
    post:
      summary: Add an attendee registration
      parameters:
      - name: SomeHeader
        in: header
        schema:
          type: string
      responses:
        '202':
          description: only required fields included
(...)
```
Generates a schema file like this
```
(...)
  "/api/v1/register": {
    /** Add an attendee registration */
    post: {
      parameters: {
        header?: {
          SomeHeader?: string;
        };
      };
(...)
```
which I use like this:
```
const { data, error, response } = await post('/api/v1/register', {
    params: { header: { SomeHeader: value } },
    body,
  })
```
If I look into the request in Firefox, the header parameter is not send with it. I also took a quick look at the code and I don't think this is in there either.
**Expected result**
Header to be added to the headers. 

**Checklist**

- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-fetch/blob/main/CONTRIBUTING.md))

drwpow:
## Changes

Fixes #1230. Adds testing & support for `params.header`.

## How to Review

Tests should pass.

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/packages/openapi-fetch/src/index.test.ts
+++ b/packages/openapi-fetch/src/index.test.ts
@@ -85,106 +85,204 @@ describe("client", () => {
       }
     });
 
-    it("requires path params", async () => {
-      const client = createClient<paths>({ baseUrl: "https://myapi.com/v1" });
-      mockFetch({ status: 200, body: JSON.stringify({ message: "OK" }) });
+    describe("params", () => {
+      it("path", async () => {
+        const client = createClient<paths>({ baseUrl: "https://myapi.com/v1" });
+        mockFetch({ status: 200, body: JSON.stringify({ message: "OK" }) });
 
-      // expect error on missing 'params'
-      // @ts-expect-error
-      await client.get("/blogposts/{post_id}", {});
-
-      // expect error on empty params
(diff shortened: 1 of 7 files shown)


packages/openapi-fetch/src/index.test.ts
```diff
@@ -85,106 +85,204 @@ describe("client", () => {
       }
     });
 
-    it("requires path params", async () => {
-      const client = createClient<paths>({ baseUrl: "https://myapi.com/v1" });
-      mockFetch({ status: 200, body: JSON.stringify({ message: "OK" }) });
+    describe("params", () => {
```
drwpow, line 88: Minor rearranging of test suite to be more organized

PR #1242, opened 2023-07-25
Improve impossible body typing

### LINKED ISSUE #1124: Impossible 'body' field required in GET
**Description**

I'm seeing an issue where generated typescript types result in some impossible types:

```
Type error: Argument of type '{ params: { path: { id: string; }; }; }' is not assignable to parameter of type 'FetchOptions<{ parameters: { path: { id: string; }; }; responses: ...snipped
  Property 'body' is missing in type '{ params: { path: { id: string; }; }; }' but required in type '{ body: never; }'.

  10 |
  11 | async function exampleGet() {
> 12 |   const { data, error } = await get(`/v1/endpoint`, {
     |                                                                             ^
  13 |     params: { path: { id: "todo" } },
  14 |   });
  15 | }
```

I am able to work around it by adding a field `body: "" as never`, which seems to be forcing TS into the right path.  `body: undefined` or `body: null` does not work either.

**Reproduction**

I used the latest versions of openapi-fetch, openapi-typescript, and an internal OpenAPI spec with only GETs in it.

Something relevant in the generated code seems to be:
```
export interface components {
...
  responses: never;
  parameters: never;
  requestBodies: never;  // <--
  headers: never;
  pathItems: never;
}
```

**Expected result**

I can work out the types in this scenario more easily.

**Checklist**

- [X] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-fetch/blob/main/CONTRIBUTING.md))

drwpow:
## Changes

Improves #1124 by adding better inference to “impossible” body types. But I think the real fix will come in #1280 which fixes the underlying typing for responses.

## How to Review

CI should pass.

## Checklist

N/A

### DIFF
--- a/.changeset/cold-pets-sit.md
+++ b/.changeset/cold-pets-sit.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Fix impossible body typing
--- a/packages/openapi-fetch/src/index.test.ts
+++ b/packages/openapi-fetch/src/index.test.ts
@@ -235,7 +235,7 @@ describe("client", () => {
 
         // expect error on missing `body`
         // @ts-expect-error
-        await client.GET("/blogposts", {});
+        await client.PUT("/blogposts", {});
 
         // expect error on missing fields
         // @ts-expect-error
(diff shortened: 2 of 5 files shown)


djMax:
I can confirm that making this same change on my "variant" for apollo-datasource-rest fixes the issue.

PR #1246, opened 2023-07-25
Fix remote path item object $refs

### LINKED ISSUE #1107: Support $ref at path item object level
**Description**

Based on the openapi spec `$ref` is supported in the path-item-object config (see [spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#path-item-object)). When I tried generating the types for the following schema it does not generate the link correctly and the result object is empty.

``` yaml
openapi: "3.0"
info:
  title: test
  version: "1.0"
paths:
  /test:
    $ref: './_paths-test.yaml'
...
``` 

**Proposal**

Support `$ref` at the path item object level instead of the [operation object](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#operation-object) where it is not supported (we can keep it for backward compatibility purposes).

**Checklist**

- [ ] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/CONTRIBUTING.md))

drwpow:
## Changes

Fixes #1107. Also squashes quite a few remote schema bugs in the process, too 🎉 

## How to Review

Tests should pass.

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/packages/openapi-typescript/examples/digital-ocean-api.ts
+++ b/packages/openapi-typescript/examples/digital-ocean-api.ts
@@ -3984,7 +3984,7 @@ export interface external {
        */
       type?: "custom" | "lets_encrypt";
     };
-    /** Custom Certificate Request */    certificate_request_custom: certificate_create_base[] & {
+    /** Custom Certificate Request */    certificate_request_custom: external["resources/certificates/models/certificate_create.yml"]["certificate_create_base"] & {
       /**
        * @description The contents of a PEM-formatted private-key corresponding to the SSL certificate.
        * @example -----BEGIN PRIVATE KEY-----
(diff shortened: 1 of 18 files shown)


packages/openapi-typescript/examples/digital-ocean-api.ts
```diff
@@ -3984,7 +3984,7 @@ export interface external {
        */
       type?: "custom" | "lets_encrypt";
     };
-    /** Custom Certificate Request */    certificate_request_custom: certificate_create_base[] & {
+    /** Custom Certificate Request */    certificate_request_custom: external["resources/certificates/models/certificate_create.yml"]["certificate_create_base"] & {
```
drwpow, line 3987: This was generating incorrectly before due to an unrelated bug. Fix was easy; we just didn’t have a test for it before.

PR #1248, opened 2023-07-25
Fix Record<string, never> appearing in discriminator union

drwpow:
## Changes

If a discriminator has `type: "object"` in it, it would completely change the output.

Unfortunately, fixing this did cause some ⚠️ minor breaking changes in generated types. But I’m hopeful the change is an _improvement_ and not a regression, as the new types should yield better TS inference.

## How to Review

Tests should pass.

## Checklist

- [x] Unit tests updated
- [ ] README updated


.eslintrc.cjs
```diff
@@ -4,8 +4,8 @@ module.exports = {
   parserOptions: {
     project: ["./tsconfig.json"],
   },
-  extends: ["eslint:recommended", "plugin:@typescript-eslint/strict"],
-  plugins: ["@typescript-eslint", "prettier"],
+  extends: ["eslint:recommended", "plugin:@typescript-eslint/strict", "plugin:vitest/recommended"],
```
drwpow, line 7: Minor linting improvements to test files

packages/openapi-typescript/examples/github-api.ts
```diff
@@ -34143,7 +34143,7 @@ export interface components {
           public?: boolean;
           /** Format: uri-template */
           pulls_url: string;
-          pushed_at: number | string;
+          pushed_at: number | string | null;
```
drwpow, line 34146: 🐛 This is a bugfix

packages/openapi-typescript/examples/github-api.ts
```diff
@@ -73010,7 +73010,7 @@ export interface components {
             files?: string[];
             bin?: Record<string, never>;
             man?: Record<string, never>;
-            directories?: string | Record<string, never>;
+            directories?: string | Record<string, never> | null;
```
drwpow, line 73013: This seems wrong, but is generated from:

```yaml
oneOf:
  - type: string
  - type: object
nullable: true
```

This is correct, at least as long as we generate `Record<string, never>` for empty objects (with no additionaalProperties)

PR #1251, opened 2023-07-26
Fix 2XX, 4XX, 5XX responses

drwpow:
## Changes

Adds support for `2XX`, `4XX`, and `5XX` responses in schema, which is [supported by the spec](https://spec.openapis.org/oas/v3.1.0#x4-8-16-2-patterned-fields).

Funny enough, I have never used this or seen this; just happened to skim over that definition and saw it was allowed. In `openapi-typescript` we just skip right over this and don’t need to handle it. But `openapi-fetch` needed to add this for proper support.

## How to Review

Tests should pass.

## Checklist

- [x] Unit tests updated
- [ ] README updated

### DIFF
--- a/packages/openapi-typescript/test/operation-object.test.ts
+++ b/packages/openapi-typescript/test/operation-object.test.ts
@@ -1,0 +1,72 @@
+import type { GlobalContext, OperationObject } from "../src/types.js";
+import transformOperationObject from "../src/transform/operation-object.js";
+
+const ctx: GlobalContext = {
+  additionalProperties: false,
+  alphabetize: false,
+  defaultNonNullable: false,
+  discriminators: {},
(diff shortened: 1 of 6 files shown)


packages/openapi-typescript/test/operation-object.test.ts
```diff
@@ -0,0 +1,72 @@
...
+            "application/json": { schema: { $ref: 'components["schemas"]["Error"]' } },
+          },
+        },
+      },
+    };
+    const generated = transformOperationObject(schema, { ctx, path: "#/paths/~get-item" });
+    expect(generated).toBe(`{
+  responses: {
+    /** @description OK */
+    "2XX": {
```
drwpow, line 52: 😅 Glad this worked! Added a test just cuz I got scared

PR #1260, opened 2023-07-28
feat: export schemas types at root level

jean-smaug:
## Changes

Parse the schemas node in order to exports them as types at root level of the file.

## How to Review

You can run examples with the `--root-types` flag 😄 

It will add the following exports at root level (example with Stripe)

Capture d’écran 2023-07-31 à 10 08 08

## Checklist

- [x] Unit tests updated


packages/openapi-typescript/src/index.ts
```diff
@@ -253,7 +256,7 @@ async function openapiTS(schema: string | URL | OpenAPI3 | Readable, options: Op
     output.splice(1, 0, "/** WithRequired type helpers */", "type WithRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };", "");
   }
 
-  return output.join("\n");
+  return output.join("\n").concat(schemasExportedTypes);
```
drwpow, line 259: nit: 

```suggestion
// --root-types
if (options.rootTypes) {
  const schemasExportedTypes = transformComponentsObjectToTypes((allSchemas["."].schema as OpenAPI3).components;
  output.push(schemasExportedTypes);
}

  return output.join("\n");
```

This has the same execution as what you wrote, but it contains all the logic in one place rather than scattering it around in multiple places.

jean-smaug: I wish I could do this, but `allSchemas` is deleted line 112 of this file 😅 

I imagine you did this on purpose, is there any counter-indication to remove this line ?

drwpow: Sorry; not sure I understand?

drwpow (changes requested):
Still in favor of this merging! But just marking as “changes requested” until it’s ready for review again.

jean-smaug:
> Still in favor of this merging! But just marking as “changes requested” until it’s ready for review again.

No problem 👍 
I'm working on it today

drwpow:
@alex-aveva it’s because the original PR was based off of `6.5.3` so that’d be the reported version.

I’m not sure of the status of this PR; some tests were added but we still need a few more tests (such as testing conflicts are handled, testing generating rootTypes from external sources, and testing that shapes beyond `#/components/schemas` work as expected) in order to merge this.

I did rebase it against the `6.x` branch since now `main` is tracking the upcoming v7 release.

packages/openapi-typescript/package.json
```diff
@@ -6,6 +6,12 @@
     "name": "Drew Powers",
     "email": "drew@pow.rs"
   },
+  "contributors": [
```
drwpow, line 9: We’re not using this field in `package.json`. We’re tracking contributors via the docs site. Let’s remove it.

drwpow:
Just checking in on this—again, still in favor of merging, but we still need:

1. **Conflict resolution** (with tests): not only test that conflicts are handled properly (`components["schemas"]["foo.bar.baz"]` vs `components["schemas"]["foo-bar-baz"]`), but that the resolution mechanism is well-defined and deterministic.
2. **Producing schema objects beyond #/components/schemas**. Also with the tests to assert this is being handled properly.
3. **Tests, tests, tests.** In case it wasn’t clear, this is a very complex feature add, so it will need many, many tests added even beyond the previous 2 points. Think of every way this can break with edge cases.

Unfortunately, these are all requirements for merging, because this is a big feature. And big features place a burden on maintainers to carry longterm. This still has lots of sharp edges that need to be filed down to last in the project longterm.

drwpow:
@alex-aveva I’m not sure about the state of this; at this point I think it’s fair to take over (and open a new PR)! The [list of requirements](https://github.com/drwpow/openapi-typescript/pull/1260#issuecomment-1658694431) still has to be met though 😅

knut-sverdrup-adsk:
If this PR is being picked up again, I'd like to chime in with my 2 cents on this part of the requested changes: 

> 4\. **Resolution to original types.** I’d probably suggest this instead of rebuilding the entire type again:
>     ```ts
>     export type PaymentLinksResourcePhoneNumberCollection = components["schemas"]["payment_links_resource_phone_number_collection"];
>     ```
>     
>     
>         
>           
>         
>     
>           
>         
>     
>         
>       
>     That would not only be more performant; it’d also reduce errors because then there’s not 2 independently-generated duplicates of the same type (which may have a bug in some obscure case that is hard to test for, and if the types are buggy, this library is useless!)

Although aliasing like this would be more performant and arguably safe, from my point of view there are a couple of substantial downsides as well. 

First off, aliasing does not retain docstrings, so any generated descriptions are lost. In my eyes, semantic descriptions are key enablers for ease-of-use, and with the current setup it is necessary to copy schema descriptions from generated file into the file which re-exposes types if you want to retain this information. 

Secondly, it renders tools such as [TypeDoc](https://typedoc.org/guides/overview/) less valuable for documenting types, since the shape is not inferred from aliases. 

The initial screenshot posted in the PR description shows output which is exactly what I want to achieve, and some benefits would be lost by the quoted suggestion. 

@drwpow do these arguments make sense to you? 

@alex-aveva if this PR ends up taking the approach of rebuilding types to give full access to the shape and docstrings directly, I'd be happy to contribute in finalising the PR -- don't hesitate to reach out :)

tgross35:
> First off, aliasing does not retain docstrings, so any generated descriptions are lost. In my eyes, semantic descriptions are key enablers for ease-of-use, and with the current setup it is necessary to copy schema descriptions from generated file into the file which re-exposes types if you want to retain this information.

It seems like maybe docstrings are preserved for the properties but just not the parent types? For example, I have the alias `type LoginSuccess = components["schemas"]["LoginSuccess"]` and hints still work for the fields:

image

But they do not seem to work for the parent types:

image

If this is consistent, maybe it would be OK to still export type aliases (i.e. not recreate the type with all fields, like I do in my example), but reapply the schema's docstring to the type alias.

fredrikssongustav:
@alex-aveva did you find time to pick this up? Would be a banger feature.

drwpow:
> First off, aliasing does not retain docstrings, … Secondly, it renders tools such as [TypeDoc](https://typedoc.org/guides/overview/) less valuable for documenting types … @drwpow do these arguments make sense to you?

Ah that’s a great callout. If it’s feasible, I’d like to still alias it but **copy the docstring from the aliased type** for this very reason. In 6.x this is difficult because there’s no resolver, but 7.x’s resolver makes lookups trivial and fast. So if this lands in 6.x, it could get this fix when being ported into 7.x.

cmcnicholas:
need any help with this? its an absolute must have for a large codebase I have things like:

```ts
const orgs = ref<components['schemas']['WebApiEndpointsOrganisationModelsOrganisationWebResponse']['organisations']>()
```

drwpow:
@cmcnicholas the current status of this is there is a lot of work to do, but if someone else wants to pick it up it’s fair game (so long as they [hit the original list of requirements](https://github.com/drwpow/openapi-typescript/pull/1260#issuecomment-1658694431) and take into account other great feedback here like @knut-sverdrup-adsk, @alex-aveva, and others)!

I’m going to close this PR not as a “won’t accept this in the repo” but just leaving room for someone else to pick it up.

Just to restate my feelings, this isn’t something that’s high on my priority list, but since there seems to be community interest in it I’d accept the addition. But as this PR stands it would leave me a lot of work in bugfixes and maintenance getting this to meet the existing quality of the rest of the codebase, hence [my change request](https://github.com/drwpow/openapi-typescript/pull/1260#issuecomment-1658694431)). If anyone else makes a PR that adds this feature and doesn’t leave any loose ends untied, I’d happily accept it and ship it.

Issue #888, opened 2022-03-23
Request: update docs for recommended programmatic usage post-ESM


zackdotcomputer:
Right now the README explains how to import and use the openapiTS function for programmatic usage, but doesn't give context around how to do so for folks who are on LTS node or ts-node with default settings (i.e. did not have `"type": "module"` in their package.json and are just trying to run an `importTypes.ts` script with `ts-node`).

I think that ESM is just now becoming the default but the tooling around it is still pretty early and confusing between the prompts to add `"type": "module"` fighting with `ts-node` suddenly [forgetting how to read TS files](https://stackoverflow.com/questions/62096269/cant-run-my-node-js-typescript-project-typeerror-err-unknown-file-extension) when you do. It would be nice to not only provide guidance on what should go in one's `importTypes` script but also the recommended way to run it now that the 5.x releases are shipping without a CJS copy.

drwpow:
I’d be in favor of adding this. It might be easier if I wrote the docs, but I’d love to source info from people:

1. What are you using to compile TypeScript (`tsc`? `ts-node`? `babel`? `esbuild`?)
2. Are you using a bundler (`webpack`, `vite`, etc.)
3. Are you transforming your TypeScript with something other than the TS compiler (e.g. Babel)?
4. What platform are you targeting? (Node version, along with whether a CJS or ESM build)
5. What is the final code that gets built/run (e.g. does your code transform to `require('openapi-typescript')` or `import … from 'openapi-typescript'`)?

All of these questions do contribute to different setups for people.

Unfortunately, ESM errors are all a side-product of a runtime completely replacing its module system, and it’s messy and frustrating for all involved (no shade at all to the Node team; I think they’re doing a wonderful job! It’s just an incredibly-hard problem dealing with upgrading the largest open software repository in history)

jonasgroendahl:
would be great

zackdotcomputer:
@drwpow I can answer your questions but with the caveat that my case is a little bit of a special case. For us, we're using this project in a JS script file that runs on-build to generate an API types file. We could `npx` this except we are passing in a custom `formatter` function to perform transforms such as `uuid` to `string`.

So the answers to your questions for our use of this project:
1. I think this is irrelevant - the import here is in a plain `.js` file that we run with `node`. We're using NextJS's Rust compiler for the final build, though.
2. Not when calling this package because this is used in a script.
3. Same answer as 2
4. CJS and Node 14.
[...]

PR #1263, opened 2023-07-31
Export CJS bundle

drwpow:
## Changes

Addresses #888. Exports a CJS bundle for `openapi-typescript`. Does so via [package.json `exports`](https://nodejs.org/api/packages.html#main-entry-point-export) so it requires a fairly-modern version of Node (but all current supported versions do).

Bundles via esbuild to gloss over complexities of CJS vs ESM deep imports for dependencies. Since all the dependencies are tiny, the bundle isn’t too bad

## How to Review

Tests should pass

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/package.json
+++ b/packages/openapi-typescript/package.json
@@ -12,11 +12,13 @@
   },
   "type": "module",
   "main": "./dist/index.js",
-  "files": [
-    "bin",
-    "dist",
-    "src"
-  ],
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
+    },
+    "./*": "./*"
+  },
   "homepage": "https://openapi-ts.pages.dev",
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


packages/openapi-typescript/package.json
```diff
@@ -38,7 +40,9 @@
     "url": "https://github.com/drwpow/openapi-typescript/issues"
   },
   "scripts": {
-    "build": "del dist && tsc -p tsconfig.build.json",
+    "build": "run-s -s build:*",
+    "build:esm": "tsc -p tsconfig.build.json",
+    "build:cjs": "esbuild --bundle --platform=node --target=es2019 --outfile=dist/index.cjs src/index.ts --footer:js=\"module.exports = module.exports.default;\"",
```
drwpow, line 45: With `dist/index.cjs`, the hope is that this will just shadow TypeScript types generated for `dist/index.js`. But we’ll see.

drwpow: Also the `--footer:js` is a hack to get rid of `require(…).default` without affecting the ESM package.

I spent some time trying to bundle with Rollup, but it turned out to be far more complex, and the end result still required similar hacks (just more in number)

packages/openapi-typescript/package.json
```diff
@@ -38,7 +40,9 @@
     "url": "https://github.com/drwpow/openapi-typescript/issues"
   },
   "scripts": {
-    "build": "del dist && tsc -p tsconfig.build.json",
+    "build": "run-s -s build:*",
+    "build:esm": "tsc -p tsconfig.build.json",
+    "build:cjs": "esbuild --bundle --platform=node --target=es2019 --outfile=dist/index.cjs --external:js-yaml --external:undici src/index.ts --footer:js=\"module.exports = module.exports.default;\"",
```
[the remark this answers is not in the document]
drwpow: 🤔 Probably also need to remove `undici`, but that will wait until we [drop Node 16 support on Sep 11, 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol). Though there are probably a few Node 16 bugs already that make upgrading worth it.

djMax:
I think some situations (I can't isolate yet) do not pick up the types with this change.

PR #1269, opened 2023-08-02
Stop trimming whitespace other than linebreaks in string values

### LINKED ISSUE #1224: Enum values with trailing white space are trimmed
**Description**

In OpenAPI, it is possible to define an enum type, with different string constants as its values. These string constants might contain trailing white space, or in the extreme case consist of only white space. 

Example OpenAPI YAML:

```
FooBarCode:
  type: string
  description: "Character code to indicate the FooBar attitude"
  enum:
    - " "
    - "1"
    - "2"
    - "X"
```

Currently, when `openapi-typescript` generates types for such enums, any trailing whitespace is trimmed.

Example resulting Typescript Type with current bug:

```
FooBarCode: "" | "1" | "2" | "X"
```

Expected behaviour for resulting Typescript Type:

```
FooBarCode: " " | "1" | "2" | "X"
```

Currently, the generated type does not match the OpenAPI spec, and as a result type checking cannot be done, and in fact generates false positive errors.

**Reproduction**

This occurs for any enum with string values containing trailing white space.
It seems that the trimming occurs [here in the code](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-typescript/src/transform/schema-object.ts#L54).

**Resolution**

If the maintainers agree that the current behaviour should be considered as a bug, I'm happy to open an MR with unit tests for this scenario and implement a fix.

**Checklist**

pimveldhuisen:
## Changes

This PR attempts to fix #1224 while minimizing other impact.

## How to Review

The unit test gives a clear example of the scenario in which the behaviour is currently not as expected. The code change is required to make the unit test pass.

## Checklist

- [x] Unit tests updated
- [x] `docs/` updated (if necessary) -> Not necessary
- [x] `pnpm run update:examples` run (only applicable for openapi-typescript) -> No changes produced?

### DIFF
--- a/packages/openapi-typescript/src/utils.ts
+++ b/packages/openapi-typescript/src/utils.ts
@@ -273,7 +273,7 @@ export function tsUnionOf(...types: (string | number | boolean)[]): string {
 /** escape string value */
 export function escStr(input: any): string {
   if (typeof input !== "string") return JSON.stringify(input);
-  return `"${input.trim().replace(DOUBLE_QUOTE_RE, '\\"')}"`;
+  return `"${input.replace(LB_RE, "").replace(DOUBLE_QUOTE_RE, '\\"')}"`;
 }
 
 /** surround a JS object key with quotes, if needed */
--- a/.changeset/new-rats-yawn.md
+++ b/.changeset/new-rats-yawn.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Stop trimming whitespace other than linebreaks in string values
--- a/packages/openapi-typescript/test/schema-object.test.ts
+++ b/packages/openapi-typescript/test/schema-object.test.ts
@@ -57,6 +57,12 @@ describe("Schema Object", () => {
   status?: "complete" | "incomplete";
 }`);
       });
+
+      test("enum (whitespace)", () => {
+        const schema: SchemaObject = { type: "string", enum: [" blue", "green ", " ", ""] };
+        const generated = transformSchemaObject(schema, options);
+        expect(generated).toBe('" blue" | "green " | " " | ""');
+      });
     });
 
     describe("number", () => {


packages/openapi-typescript/src/utils.ts
```diff
@@ -273,7 +273,7 @@ export function tsUnionOf(...types: (string | number | boolean)[]): string {
 /** escape string value */
 export function escStr(input: any): string {
   if (typeof input !== "string") return JSON.stringify(input);
-  return `"${input.trim().replace(DOUBLE_QUOTE_RE, '\\"')}"`;
+  return `"${input.replace(LB_RE, "").replace(DOUBLE_QUOTE_RE, '\\"')}"`;
```
drwpow, line 276: 👍 this is a smarter way to handle it

drwpow (approved):

drwpow:
Hm this seems to now be generating this type in the [Stripe API example](https://github.com/drwpow/openapi-typescript/pull/1272/files#diff-bd5a2541e9d46932ab33a52735c34ab1947820350fc168e23b93f33baf34393bL17252):

```diff
- name?: string;
+ name?: string | "";
```

This is on the right track but I think we’ll need to add a couple more tests before shipping

drwpow:
Oh my god this may have actually been a bug that was fixed 🙃 

```yaml
name:
  anyOf:
    - maxLength: 5000
      type: string
    - enum:
        - ''
      type: string
```

That kinda looks right to me! But will poke through the other changes to see if anything else is unexpected

pimveldhuisen:
Seems to me that the change in the Stripe API you mention is actually the result of a [change in the source OpenAPI yaml](https://github.com/drwpow/openapi-typescript/pull/1272/files#diff-2d1fc5317a97b43ab1a99cf25c948e22f937253101a6d3362a1aac754caada54R40917), not related to the code change in this PR.

` string | ""` is a bit redundant since the empty string is itself a string, but the TypeScript type here just inherits the redundancy from the OpenAPI spec. You could argue for or against removing this redundancy when generating the TypeScript Types, but it seems unrelated to this PR (although both coincidentally involve empty strings! 😄).

PR #1281, opened 2023-08-08
Refactor CLI path handling, fixing several bugs

### LINKED ISSUE #1091: Bulk types generation
**Description**

Previously we were running the following command:
`npx openapi-typescript 'generated-schemas/**/*.yml' --output generated-schemas`

It would get all `.yml` files in the folder and bulk convert them into appropriate `.ts` files.

Version `6.2.3` broke it and started throwing the following error:
```
Error: EISDIR: illegal operation on a directory, open `/path/generated-schemas'
    at Object.openSync (node:fs:585:3)
    at Object.writeFileSync (node:fs:2155:35)
    at generateSchema (file:///path/node_modules/openapi-typescript/bin/cli.js:122:8)
    at async file:///path/node_modules/openapi-typescript/bin/cli.js:195:7
    at async Promise.all (index 0)
    at async main (file:///path/node_modules/openapi-typescript/bin/cli.js:185:3) {
  errno: -21,
  syscall: 'open',
  code: 'EISDIR',
  path: '/path/generated-schemas'
}
error Command failed with exit code 1.
```

Our solution was to convert each file separately, but would be great to get the old way working.
```
npx openapi-typescript generated-schemas/file1.yml --output generated-schemas/file1.ts
npx openapi-typescript generated-schemas/file2.yml --output generated-schemas/file2.ts
```

**Reproduction**

Version `6.2.0` works

pvanagtmaal:
## Changes

Fixes #1091

Uses of `URL` have been replaced by their `path` variants. This makes more sense because we're handling actual file paths here. This fixes several bugs, including the reporting one, but also for example exporting to absolute paths.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/bin/cli.js
+++ b/packages/openapi-typescript/bin/cli.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 import fs from "node:fs";
-import path from "path";
+import path from "node:path";
 import { URL } from "node:url";
 import glob from "fast-glob";
 import parser from "yargs-parser";
@@ -114,7 +114,10 @@ async function generateSchema(pathToSpec) {
     let outputFilePath = new URL(flags.output, CWD); // note: may be directory
     const isDir = fs.existsSync(outputFilePath) && fs.lstatSync(outputFilePath).isDirectory();
     if (isDir) {
-      const filename = pathToSpec.replace(EXT_RE, ".ts");
+      if (typeof flags.output === 'string' && !flags.output.endsWith('/')) {
+        outputFilePath = new URL(`${flags.output}/`, CWD)
+      }
+      const filename = path.basename(pathToSpec).replace(EXT_RE, ".ts");
       const originalOutputFilePath = outputFilePath;
       outputFilePath = new URL(filename, originalOutputFilePath);
       if (outputFilePath.protocol !== 'file:') {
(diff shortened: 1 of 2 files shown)


drwpow:
I would prefer to always work with URLs as they’re much more resilient to preventing Windows vs Unix path bugs ([fileURLToPath](https://nodejs.org/api/url.html#urlfileurltopathurl)), easy to modify, and prevent sloppy string mashing. They actually are meant for files, not just remote resources (that’s the whole design of the `file:` protocol). Though we don’t have a ton of Windows tests in this repo, in other applications URLs have saved me hundreds of hours of debugging time from my work on Snowpack & Astro, and other projects like SvelteKit take this approach. I believe this bug could be fixed without undoing the URLs.

packages/openapi-typescript/bin/cli.js
```diff
@@ -140,15 +136,24 @@ async function main() {
     console.info(HELP);
     process.exit(0);
   }
-  const packageJSON = JSON.parse(fs.readFileSync(new URL("../package.json", import.meta.url), "utf8"));
+  const packageJSON = (await import("../package.json", { assert: { type: 'json' }})).default;
```
drwpow, line 139: I really like import assertions! But let’s not use it here as it’s only stabilized in v20 and we still want to support v18 (and unofficially we support v16 for the next month until it’s EOL)

drwpow (approved):
Thank you! Awesome fix

Issue #1158, opened 2023-06-02
Discriminator property not set for 2-level inheritance types


melvinnijholt:
**Description**

For 1-level inheritance types the discriminator property is to a static value that identifies the type. For 2-level inheritance types this is not the case which makes it not possible to detect the type that is returned by the api endpoint. 

Take for example `Pet` -> `Dog` -> `Poodle`. When a discriminator is specified for `Pet` this will be set on `Dog` but it will not be set on `Poodle`.

I don't know if this is excepted behavior, if it is than I will change this to a feature request.

| Name                 | Version                      |
| :------------------- | :--------------------------- |
| `openapi-typescript` | `6.2.4`                      |
| Node.js              | `16.19.0`                      |
| OS + version         | `Windows 11` |

**Reproduction**

I have a [swagger.json](https://github.com/melvinnijholt/openapi-typescript-discriminator-issue/blob/0fb92f6aabb08e62e7f5ebc47f52d19d4933c8bf/swagger.json) that contains the following definition for `Pet` that defines a discriminator named `_petType`.   

```json
"Pet": {
  "required": [
    "_petType"
  ],
  "type": "object",
  "properties": {
    "_petType": {
      "type": "string"
    },
    "nickName": {
      "type": "string",
      "nullable": true
    }
  },
  "additionalProperties": false,
  "discriminator": {
    "propertyName": "_petType"
  }
}
```

It also contains a definition for `Dog` that defines the `allOf` property to `Pet` 
```json
"Dog": {
  "type": "object",
  "allOf": [
    {
      "$ref": "#/components/schemas/Pet"
    }
  ],
  "properties": {
    "bark": {
      "type": "boolean"
    }
  },
  "additionalProperties": false
}
```

and a definition for `Poodle` that defines the `allOf` property to `Dog` 
```json
"Poodle": {
  "type": "object",
  "allOf": [
    {
      "$ref": "#/components/schemas/Dog"
    }
  ],
  "properties": {
    "showDog": {
      "type": "boolean"
    }
  },
  "additionalProperties": false
}
```

The generated [swagger.ts](https://github.com/melvinnijholt/openapi-typescript-discriminator-issue/blob/0fb92f6aabb08e62e7f5ebc47f52d19d4933c8bf/swagger.ts) contains `Dog` specifying a value `_petType: "Dog"` for the discriminator. The discriminator value for `Poodle` is missing.

```typescript
export interface components {
  schemas: {
    Pet: {
[...]

PR #1289, opened 2023-08-09
fix incorrect types generated when refs reference types with discriminators

adamschoenemann:
## Changes

This PR fixes the code, so that incorrect types are not generated when referencing a type (with a discriminator with an implicit mapping) inside a oneOf/anyOf

## How to Review

I added tests to both `index.test.ts` and  `schema-object.test.ts`, and I tried to fix the code, but I must admit it was kind of a shot in the dark, but it does seem to work :)

## Checklist

- [x] Unit tests updated
- [x] ~~`docs/` updated (if necessary)~~
- [x] `pnpm run update:examples` run (only applicable for openapi-typescript) (I ran this but nothing changed)


adamschoenemann:
Test macos failed, because the digital ocean test timed out. I'm pretty sure it's got nothing to do with this change... I observed this test being quite slow locally as well, on the main branch

packages/openapi-typescript/test/index.test.ts
```diff
@@ -625,6 +625,93 @@ export interface components {
 
 export type external = Record<string, never>;
 
+export type operations = Record<string, never>;
+`);
+    });
+
+    test("discriminator with implicit mapping (oneOf)", async () => {
```
drwpow, line 632: Awesome tests!

packages/openapi-typescript/test/schema-object.test.ts
```diff
@@ -374,48 +374,6 @@ describe("Schema Object", () => {
         expect(generated).toBe("0 | 1");
       });
 
-      test("empty object + oneOf is ignored", () => {
```
drwpow, line 377: We’ll want to keep this test—this is intended behavior.

adamschoenemann: Oh, I didn't mean to delete this test at all! I'll fix it when I'm back at my laptop tomorrow.
Also, thanks for the kind words, much appreciated ☺

packages/openapi-typescript/test/index.test.ts
adamschoenemann: Thanks!

packages/openapi-typescript/test/schema-object.test.ts
adamschoenemann: Un-deleted the test now :)

drwpow (approved):
Awesome work! 👏 Thank you

drwpow:
Just a heads up: In reviewing issues I realized #1158 does outline a bug that results in slightly-altered typegen in one of the tests added:

```diff
- components["schemas"]["parent"] | null
+ {
  operation: "schema-object";
} & (components["schemas"]["parent"] | null)
```

This seems weird, but I believe the discriminator spec does suggest that **if `propertyName` is present, it should generate that property even with no mapping**.

It’s weird, but seems to be how it’s outlined 🤷. Anyways that shouldn’t cause a regression in this fix at all; just wanted to flag an incoming change.

PR #1294, opened 2023-08-10
Add Windows tests (back), skip DO test in macOS/Win

drwpow:
## Changes

Adds Windows test, skips DigitalOcean test in macOS/Windows. The DigitalOcean API is the “meanest” test in the suite by far because it is a ridiculous abuse of remote $refs. But it takes too long in CI so that it times out (I’m guessing it’s [a read/write bottleneck](https://www.cncf.io/blog/2023/02/02/docker-on-macos-is-slow-and-how-to-fix-it/) because that’s especially slow and that’s most of what that test is doing), and I’d rather not slow down CI to let it finish

Keeping the DigitalOcean tests for the default ubuntu runners though because those are within the acceptable limits

## How to Review

- CI should pass

## Checklist

- [x] Unit tests updated


packages/openapi-typescript/test/cli.test.ts
```diff
@@ -1,8 +1,10 @@
 import { execa } from "execa";
-import fs from "node:fs";
-import { URL } from "node:url";
+import { URL, fileURLToPath } from "node:url";
+import os from "node:os";
+import { readFile } from "./helpers.js";
 
-const cwd = new URL("../", import.meta.url);
+const root = new URL("../", import.meta.url);
+const cwd = os.platform() === "win32" ? fileURLToPath(root) : root; // execa bug: fileURLToPath required on Windows
```
drwpow, line 7: this is one of the dumbest hours of my life I’ve ever spent

Issue #1018, opened 2022-12-12
Required object field with additionalProperties does not generate a required TypeScript field


sgrimm:
**Description**

If a payload has a required object field whose `additionalProperties` specifies a schema for the values, the generated type shouldn't allow `undefined` values.

**OpenAPI**

The real schema where I'm hitting this is [here](https://terraware.io/v3/api-docs.yaml) but this toy schema demonstrates the problem.

```yaml
openapi: 3.0.1
components:
  schemas:
    Payload:
      type: object
      required:
        - children
      properties:
        children:
          type: object
          additionalProperties:
            type: object
            required:
              - field
            properties:
              field: string
```

**Generated TS**

```ts
export type paths = Record<string, never>;

export type webhooks = Record<string, never>;

export interface components {
  schemas: {
    Payload: {
      children: {
        [key: string]: {
          field: string;
        } | undefined;
      };
    };
  };
  responses: never;
  parameters: never;
  requestBodies: never;
  headers: never;
  pathItems: never;
}

export type external = Record<string, never>;

export type operations = Record<string, never>;
```

**Expected TS**

Same as above but without the `| undefined`.

```ts
export type paths = Record<string, never>;

export type webhooks = Record<string, never>;

export interface components {
  schemas: {
    Payload: {
      children: {
        [key: string]: {
          field: string;
        };
      };
    };
  };
  responses: never;
  parameters: never;
  requestBodies: never;
  headers: never;
  pathItems: never;
}

export type external = Record<string, never>;

export type operations = Record<string, never>;
```

**Checklist**

- [X] My OpenAPI schema passes a validator
- [X] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/CONTRIBUTING.md))

acnebs:
Any update on this? I've been bumping up against it, and right now am just grepping occurrences in the output TS, which is obviously sub-optimal.

drwpow:
This behavior was a decision made to improve type safety because most users don’t have [noUncheckedIndexAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) enabled, and this library can’t check that’s set. But when dealing with `additionalProperties`, their whole nature is “this key may either exist or not” so the `| undefined` is necessary in order to not have runtime bugs (if the key was known, and required, it would be part of the schema).

[...]

PR #1295, opened 2023-08-11
Avoid adding undefined to the index signature for additionalProperties

### LINKED ISSUE #1267: Generated type of additionalProperties values should not include undefined
**Description**

Reposting this issue https://github.com/drwpow/openapi-typescript/issues/1070

I strongly believe that the library should generate 
```
type Foo = {
    [index: string]: string
}
```

instead of
```
type Foo = {
    [index: string]: string | undefined
}
```

The reasoning given in the issue above refers to the additional safety provided by the undefined, however this safety is now provided by `noUncheckedIndexAccess` mentioned here https://github.com/drwpow/openapi-typescript/issues/1070#issuecomment-1500751490

The current setup makes iteration difficult e.g.
```
const obj: Foo = {};
const values = Object.values(obj); -> string | undefined
```

When in reality, undefined is not present in JSON and therefore that is not accurate.

| `openapi-typescript` | `6.4.0`                      |
| Node.js              | `x.x.x`                      |

**Reproduction**

**Expected result**

_(in case it’s not obvious)_

**Checklist**

pvanagtmaal:
## Changes

Fixes #1267 and #1018 

## How to Review

A slightly tricky part about this is that we need to account for objects with not all properties required. In that case, we still need to add undefined to the index signature to avoid TypeScript errors

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/test/schema-object.test.ts
+++ b/packages/openapi-typescript/test/schema-object.test.ts
@@ -204,6 +204,35 @@ describe("Schema Object", () => {
 }`);
       });
 
+      test("additionalProperties with all required properties", () => {
+        const schema: SchemaObject = {
+          type: "object",
+          properties: { property: { type: "boolean" } },
+          additionalProperties: { type: "string" },
+          required: ["property"]
+        };
+        const generated = transformSchemaObject(schema, options);
+        expect(generated).toBe(`{
+  property: boolean;
+  [key: string]: string;
+}`);
+      });
+
+      test("additionalProperties with partly required properties", () => {
+        const schema: SchemaObject = {
+          type: "object",
+          properties: { property: { type: "boolean" }, property2: { type: "boolean" }},
+          additionalProperties: { type: "string" },
+          required: ["property"]
+        };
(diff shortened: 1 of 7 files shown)


drwpow:
This is great—thank you! I admit I was wrong on this original decision, and this is a better path forward for this library.

I do agree that a minor version is best—this is more than a patch change. But I don’t think it would _technically_ be a breaking change and it doesn’t feel right for a major version.

packages/openapi-typescript/test/schema-object.test.ts
```diff
@@ -204,6 +204,35 @@ describe("Schema Object", () => {
...
+          required: ["property"]
+        };
+        const generated = transformSchemaObject(schema, options);
+        expect(generated).toBe(`{
+  property: boolean;
+  [key: string]: string;
+}`);
+      });
+
+      test("additionalProperties with partly required properties", () => {
```
drwpow, line 221: 🙏 great tests!

drwpow (approved):

drwpow:
Released in `v6.5.0`!

PR #1296, opened 2023-08-14
Fix invalid typescript for an array of integers

jonsetzky:
## Changes

Fixed invalid typescript generation when items of an array is string instead of SchemaObject.

```
"values": {
  "type": "array",
  "items": "integer",
}
```
which would result in:
```
values: integer[]
```

## How to Review

Not sure if the test is in a sensible location or if this PR should've been an issue instead.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/tender-timers-sniff.md
+++ b/.changeset/tender-timers-sniff.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Fix invalid typescript for an array of integers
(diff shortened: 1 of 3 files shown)


wolfy1339:
Your schema is not completely accurate. While it may pass the spec it is misleading and it is better to be explicit.

It should be
```
"values": {
  "type": "array",
  "items": { "type: "integer" }
}
```

Or if you really meant it as the literal string
```
"values": {
  "type": "array",
  "items": { "type: "string", "enum": "integer" }
}
```

jonsetzky:
A service I'm using uses this non-explicit format and openapi-typescript therefore generates invalid code.

drwpow:
@wolfy1339 is correct this is definitely invalid syntax. Even the [Redocly CLI](https://redocly.com/docs/cli/commands/lint) throws this error parsing it:

```
Expected type `Schema` (object) but got `string`

 7 | values:
 8 |   type: array
 9 |   items: integer
```

Even if this exists in a schema you don’t control, invalid syntax is invalid syntax and this library won’t support it (because at a certain point, invalid syntax can and will disagree with valid syntax, so it’s in the name of reducing bugs).

Issue #1125, opened 2023-05-09
Svelte custom fetch


ESchouten:
Svelte provides a [custom fetch](https://kit.svelte.dev/docs/load#making-fetch-requests) in their load functions which you have to use in order to use SSR and their middleware handlers.
For this library Svelte is listed as supported, but I can't figure out how this Svelte fetch is used. Are there examples? Or is this underway?

pkb-pmj:
It's currently not possible to pass `fetch` to a single request like this
```ts
const { response } = await POST("/auth/login", {
    body: form.data,
    fetch,
});
```
And it seems necessary, because the `fetch` provided by SvelteKit inherits e.g. request-specific cookies.
Or is there some other way to use it which I have not thought of? If not, I can make a PR basically right now 😄

ESchouten:
@pkb-pmj I have not used this library (yet), but I guess you could do createClient<paths>({ baseUrl: "https://myapi.dev/v1/", fetch }) in your load function and create a client everywhere you want to fetch?

pkb-pmj:
I can also do it e.g. in `hooks.server.ts`:
```ts
export const handle = async ({ event, resolve }) => {
    const client = createClient<paths>({
        fetch: event.fetch,
        baseUrl: "/api",
    });
    event.locals.GET = client.GET;
    event.locals.PATCH = client.PATCH;
    event.locals.POST = client.POST;
    event.locals.PUT = client.PUT;

    return await resolve(event);
};
```
But I guess I don't want to create a new client on every request, when it's very easy to modify the client to accept custom per-request `fetch`.

drwpow:
Added [an example to the docs](https://openapi-ts.pages.dev/openapi-fetch/examples/#svelte--sveltekit) that uses the custom fetcher.

To @pkb-pmj’s suggestion of putting in `hooks.server.ts` I didn’t add to the example because I’m still not too familiar on that “magic” in order to recommend that. But PRs are welcome if you’d like to add it 🙂

> I guess I don't want to create a new client on every request

In most instances, yes, that’d probably be a concern. But I recently [added benchmarks](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/test/index.bench.js) and the `createClient()` setup is extremely lightweight. I profiled it at 2.5M ops/s on my machine. So yes as a general rule probably best to never do work you don’t need to. But also, based on QoL, calling `createClient()` every request won’t have any practical impacts.

[...]

PR #1310, opened 2023-08-19
Add examples/ folder for openapi-fetch

drwpow:
## Changes

To prepare for some upcoming expanded docs for openapi-fetch examples (e.g. #1125), add an `examples/` folder that moves the React Query example into, and shortens the docs. Overall this is a better experience seeing code in context.

## How to Review

Docs-only change.

## Checklist

- [ ] Unit tests updated


docs/src/content/docs/openapi-fetch/examples.md
```diff
@@ -68,194 +68,24 @@ client.GET("/some-authenticated-url", {
 });
 ```
 
-## Caching
```
drwpow, line 71: This may be slightly-controversial, but I removed the section on caching. I have a feeling like encouraging users to roll their own caching strategy in general is a bad idea. This should be replaced with more framework-specific examples that encourage good habits within your reactivity framework.

If a user is dedicated rolling their own client-side caching, then they probably aren’t the type to follow instructions 😄

docs/src/content/docs/openapi-fetch/examples.md
```diff
@@ -68,194 +68,24 @@ client.GET("/some-authenticated-url", {
...
-console.log(getUser("user-123")); // { loading: false, data: { … } }
-// Error
-console.log(getUser("user-123")); // { loading: false, error: { … } }
-```
-
-Keep in mind, though, that reactivity is difficult! You’ll have to account for stale data in the context of your application (if using a JS framework, there are already existing libraries that can handle this for you). Further, this is a contrived example that doesn’t handle invalidating the cache—both data and errors. Cache invalidation will be dependent on your specific needs.
-
-If you need to work with promises, then just using `openapi-fetch` as-is will usually suffice. If you need promises + caching, then you’ll have to come up with a solution that fits your own custom usecase.
-
-## React Query
```
drwpow, line 168: All this was just moved into `packages/openapi-fetch/examples/react-query`

Issue #1116, opened 2023-05-18
Referencing a type with discriminator generates wrong type 


sregg:
**Description**

When adding an `Owner` type that references a `Pet` in `discriminator (oneOf)` test case like so:

```
Owner: {
  type: "object",
  properties: {
    favoritePet: {
      nullable: true,
      allOf: [
        { $ref: "#/components/schemas/Pet" },
      ],
    },
  },
}
```

the generated type is wrong:

```
Owner: {
  favoritePet?: ({
    petType: "Owner";
  } & Omit<components["schemas"]["Pet"], "petType">) | null;
};
```

instead of 

```
Owner: {
  favoritePet?: components["schemas"]["Pet"];
};
```

| Name                 | Version                      |
|:---------------------|:-----------------------------|
| `openapi-typescript` | `6.2.4`                      |

**Reproduction**

See PR: https://github.com/sregg/openapi-typescript/pull/1

**Expected result**

See error message in: https://github.com/sregg/openapi-typescript/actions/runs/5009218680/jobs/8977922547?pr=1

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/CONTRIBUTING.md))

mitchell-merry:
Could you provide a full example of a schema? So we can reproduce this

sregg:
The example is in my fork:
https://github.com/sregg/openapi-typescript/pull/1/files#diff-825cc17817d6be362cee3768df588ac6fa8ee329b4b8990dcc85767d8ad3cf3b
```
const schema: OpenAPI3 = {
        openapi: "3.1",
        info: { title: "test", version: "1.0" },
        components: {
          schemas: {
            Pet: {
              oneOf: [
                { $ref: "#/components/schemas/Cat" },
                { $ref: "#/components/schemas/Dog" },
                { $ref: "#/components/schemas/Lizard" },
              ],
              discriminator: {
                propertyName: "petType",
                mapping: {
                  cat: "#/components/schemas/Cat",
                  dog: "#/components/schemas/Dog",
                  lizard: "#/components/schemas/Lizard",
                },
              },
            },
            Cat: {
              type: "object",
              properties: {
                name: { type: "string" },
                petType: { type: "string", enum: ["cat"] },
              },
              required: ["petType"],
            },
            Dog: {
              type: "object",
              properties: {
                bark: { type: "string" },
[...]

PR #1342, opened 2023-09-13
Fix discriminator propertyName inference

drwpow:
## Changes

Fixes an issue where `propertyName` wasn’t automatically generated for `discriminators` (mappings are not required; they only are needed to override the automatic inference).

Related to #1158 and #1116

## How to Review

- Tests added; tests should pass.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/src/load.ts
+++ b/packages/openapi-typescript/src/load.ts
@@ -1,4 +1,4 @@
-import type { ComponentsObject, Fetch, GlobalContext, OpenAPI3, OperationObject, ParameterObject, PathItemObject, ReferenceObject, RequestBodyObject, ResponseObject, SchemaObject, Subschema } from "./types.js";
+import type { ComponentsObject, DiscriminatorObject, Fetch, GlobalContext, OpenAPI3, OperationObject, ParameterObject, PathItemObject, ReferenceObject, RequestBodyObject, ResponseObject, SchemaObject, Subschema } from "./types.js";
 import fs from "node:fs";
 import path from "node:path";
 import { Readable } from "node:stream";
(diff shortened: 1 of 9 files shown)


packages/openapi-typescript/src/transform/schema-object.ts
```diff
@@ -159,11 +159,30 @@ export function defaultSchemaObjectTransform(schemaObject: SchemaObject | Refere
...
   }
 
-  // "type": "object" (explicit)
-  // "anyOf" / "allOf" (object type implied)
-
   // core type: properties + additionalProperties
   const coreType: string[] = [];
+
+  // discriminators
+  for (const k of ["oneOf", "allOf", "anyOf"] as ("oneOf" | "allOf" | "anyOf")[]) {
```
drwpow, line 166: Note: this was moved up **before property generation** due to the change in automatic `discriminator.propertyName` possibly adding keys to the object. We can’t detect if there may be a key conflict, but we can let the explicit properties override it.

packages/openapi-typescript/src/load.ts
```diff
@@ -296,7 +296,19 @@ export default async function load(schema: URL | Subschema | Readable, options:
...
         if (!node.discriminator) return;
-        options.discriminators[schemaID === "." ? makeTSIndex(nodePath) : makeTSIndex(["external", k, ...nodePath])] = node.discriminator;
+        const discriminator: DiscriminatorObject = { ...node.discriminator };
+
+        // handle child oneOf types (mapping isn’t explicit from children)
+        const oneOf: string[] = [];
+        if (Array.isArray(node.oneOf)) {
+          for (const child of node.oneOf) {
+            if (!child || typeof child !== "object" || !("$ref" in child)) continue;
+            oneOf.push(child.$ref);
```
drwpow, line 306: This adds automatic inference from parent `oneOf` types, but unfortunately the mapping is brittle (probably breaks for remote schemas).

I think the only way to fix this is if this library pre-bundled schemas via something from [Redocly](https://redocly.com/)

PR #1360, opened 2023-09-30
Fix CJS build for TypeScript

### LINKED ISSUE #1346: ESM is hell, how do we escape hell?
**Description**

I struggle to even come up with a description other than "something is wrong with openapi-fetch that causes it to be almost impossible to use in a typical Typescript project. I know this is never ending frustration for everyone, but I'm hoping perhaps someone can spot the problem. I'm using express, node20, and vitest.

**Reproduction**

Here's the TS config:

```
  "compilerOptions": {
    "lib": [
      "ES2022",
      "DOM"
    ],
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "target": "ES2022",
    "declaration": true,
    "sourceMap": true,
    "outDir": "./build",
    "incremental": true,
    "isolatedModules": true,
    "strict": true,
    "baseUrl": "./src",
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true,
    "experimentalDecorators": true,
    "paths": {
      "@/*": [
        "./*"
      ]
    }
  }
```

**Expected result**

Not this error:
```
The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("openapi-fetch")' call instead.
  To convert this file to an ECMAScript module, change its file extension to '.mts', or add the field `"type": "module"` to '/my-project/package.json'.ts(1479)
```

(using dynamic import doesn't help either)

**Checklist**

- [X] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-fetch/blob/main/CONTRIBUTING.md))

marcomuser:
## Changes
Should fix #1346 
- Output CJS build in separate `cjs` subdirectory and copy the type declaration file as `index.d.cts` from the dist folder
- Adjust package.json to point to correct builds/types for both cjs and esm

## How to Review

This changes only the build output for CJS projects. I created a small repo in https://github.com/marcomuser/cjsts where the mentioned bug report is reproducible. It is fixed when I update the dist folder in node_modules with the built openapi-fetch from this branch.

## Checklist

- [ ] Unit tests updated


djMax (approved):

drwpow (approved):
Thank you for this fix! I wasn’t aware the `.d.cts` was needed; I was under the false impression the type def worked for both. Fix seems great with me.

Issue #941, opened 2022-09-23
[Feature request] Generate Enums AND Union types


mkosir:
Hey 👋
I think it would be great if we could generate Enum types and Union types all together. Basically same thing as [Prisma](https://github.com/prisma/prisma) does when generating types for database models 🚀
Wdyt?

drwpow:
I think this could be offered as an option. I feel strongly that the default behavior of string unions shouldn’t change, because when data’s coming from an object it’s much easier to coerce a string union than a TypeScript enum.

One question I’d like to ask is: how would you envision this working? Like, what would the import path be? The following would have to be considered:

- Enums could come from `components` as well as `paths` and `responses`
- Account for deep-linking of enums (e.g. `components['schemas']['TopLevel']['NestedOne']['NestedTwo']['EnumProperty']`)
- Also account for invalid TypeScript characters (`Enum Type #1`)
- Lastly, whatever gets exported can’t have any conflicts with any other part of their schema (e.g. it would be impossible for them to add any property that conflicts)

Adding a proposal that takes into account those things would really help implementing (asking for anyone reading this thread!)

mkosir:
> I think this could be offered as an option. I feel strongly that the default behaviour of string unions shouldn’t change, because when data’s coming from an object it’s much easier to coerce a string union than a TypeScript enum.

Absolutely agree on this 💯 . I also always go with string unions instead of enums. 
But there are rare occasions where I "have to" use enum, usually that happens when we want to iterate over the values and render them on frontend (dropdown selection, table etc.) and of course we don't want to duplicate and hardcode those values on frontend, since we will need to keep them up to date all the time with backend (which is the source of truth).
Example: `Role` string union type defined in swagger, where on user creation page, we select predefined user role.

I wouldn't go with an option, but instead as you mentioned default behaviour of string unions shouldn't change, lets just add enums/object exported along with it.

If we check how Prisma does it:
If database model `User` and `Role` are defined as
```ts
model User {
  id        String   @id @default(uuid())
  email     String   @unique
  role      Role     @default(STANDARD)
}

enum Role {
  STANDARD
  APPRENTICE
  SUPERVISOR
  ADMINISTRATOR
}
```
Prisma CLI is going to generate types as:
```ts
export const Role: {
  STANDARD: 'STANDARD',
  APPRENTICE: 'APPRENTICE',
  SUPERVISOR: 'SUPERVISOR',
  ADMINISTRATOR: 'ADMINISTRATOR'
[...]

PR #1363, opened 2023-10-04
V7 Preview

drwpow:
# V7 Preview 🎉

This is one of the most exciting releases in this project’s history—this uses the Redocly CLI (#1344) to validate schemas and report errors ✨.

You can test it out today at:

```bash
npm i openapi-typescript@next
```

🙏 If you’re able to test this out and report any errors it’d be much appreciated. 

📖 [v7 Documentation](https://v7.openapi-ts.pages.dev/)

## Overview

- 🚀 Drastically-improved performance for multi-file schemas. Performance is a teeny bit slower for single-file schemas (an extra half-second) but not a drastic drop
- ✨ The Redocly CLI now validates & bundles schemas, which means better UX for users and reduced project maintenance
    - The codebase itself got a dramatic reduction in logic
- 🌲 Leveraging the TypeScript AST means greater accuracy without compromising speed
    - Also, using the official TypeScript lib means no feature drift, no chance of project abandonment, and bleeding-edge updates
- ✨ This also adds the hotly-requested feature top-level enums (#941).

<details>
<summary><h3>Full Changelog</h3></summary>

### Features

- ✨ **Feature**: automatically validate schemas with Redocly CLI ([docs](https://redocly.com/docs/cli/)). No more need for external tools to report errors! 🎉

    - By default, it will only throw on actual schema errors (uses Redocly’s default settings)
    - For stricter linting or custom rules, you can create a [redocly.yaml config](https://redocly.com/docs/cli/configuration/)
- ✨ **Feature**: bundle schemas with Redocly CLI
  - Any options passed into your [redocly.yaml config](https://redocly.com/docs/cli/configuration/) are respected
- ✨ **Feature**: add `enum` option to export top-level enums from schemas (#941)
- ✨ **Feature**: add `formatOptions` to allow formatting TS output
- ✨ **Feature**: header responses add `[key: string]: unknown` index type to allow for additional untyped headers
- ✨ **Feature:** allow configuration of schemas via `apis` key in redocly.config.yaml. [See docs](https://openapi-ts.pages.dev/cli/) for more info.

### Breaking Changes

-  ⚠️ **Breaking**: The Node.js API now returns the TypeScript AST for the main method as well as `transform()` and `postTransform()`. To migrate, you’ll have to use the `typescript` compiler API:

    ```diff
    + import ts from "typescript";

    + const DATE = ts.factory.createIdentifier("Date");
    + const NULL = ts.factory.createLiteralTypeNode(ts.factory.createNull());
  
      const ast = await openapiTS(mySchema, {
        transform(schemaObject, metadata) {
        if (schemaObject.format === "date-time") {
    -       return schemaObject.nullable ? "Date | null" : "Date";
    +       return schemaObject.nullable
    +         ? ts.factory.createUnionTypeNode([DATE, NULL])
    +         : DATE;
          }
        },
      };
    ```

    Though it’s more verbose, it’s also more powerful, as now you have access to additional properties of the generated code you didn’t before (such as injecting comments).

    For example syntax, search this codebae to see how the TypeScript AST is used.

    Also see [AST Explorer](https://astexplorer.net/)’s `typescript` parser to inspect how TypeScript is interpreted as an AST.

-  ⚠️ **Breaking**: Changing of several CLI flags and Node.js API options

    - The `--auth`, `--httpHeaders`, `--httpMethod`, and `fetch` (Node.js-only) options were all removed from the CLI and Node.js API
        - To migrate, you’ll need to create a [redocly.yaml config](https://redocly.com/docs/cli/configuration/) that specifies your auth options [in the http setting](https://redocly.com/docs/cli/configuration/#resolve-non-public-or-non-remote-urls)
    - You can also set your fetch client in redocly.yaml as well.
    - `--immutable-types` has been renamed to `--immutable`
    - `--support-array-length` has been renamed to `--array-length`

-  ⚠️ **Breaking**: Most optional objects are now always present in types, just typed as `:never`. This includes keys of the Components Object as well as HTTP methods.

-  ⚠️ **Breaking**: No more `external` export in schemas anymore. Everything gets flattened into the `components` object instead (if referencing a schema object from a remote partial, note it may have had a minor name change to avoid conflict).

-  ⚠️ **Breaking** `defaultNonNullable` option now defaults to `true`. You’ll now need to manually set `false` to return to old behavior.

-  ⚠️ **Breaking**: Remove globbing schemas in favor of `redocly.yaml` config. Specify multiple schemas with outputs in there instead. See [Multiple schemas](https://openapi-ts.pages.dev/docs/cli/#multiple-schemas) for more info.


wolfy1339:
This looks awesome!

There seems to be an issue compiling OpenAPI 3.1 specs.
```
Error: Unsupported schema format, expected `openapi: 3.x`
    at validateAndBundle (node_modules/openapi-typescript/dist/lib/redoc.js:88:15)
    at async openapiTS (node_modules/openapi-typescript/dist/index.js:31:20)
    at async run (openapi-webhooks/scripts/generate-types.js:90:9)
```

I looked into this with a debugger, and it seems that `document.parsed` is not an object but a string, which means that `document.parsed.{swagger,openapi}` will be undefined because those properties do not exist on the string object.
The value that I see for `document.parsed` is the path to the spec file.

Going further up the call chain, in the `parseSchema` function, the file doesn't seem to be loaded anywhere. It simply is the path to the spec that is passed around

You can find the spec I am trying to compile here: https://github.com/wolfy1339/openapi-webhooks/blob/main/packages/openapi-webhooks/generated/api.github.com.json
(It's GitHub's OpenAPI 3.1 specs with only webhooks, all routes deleted)

drwpow:
> There seems to be an issue compiling OpenAPI 3.1 specs.
>
> EDIT: This seems to happen only when passing a string path to openapi-typescript, passing a URL of the path

This might be behaving as expected, if you’re using the Node.js API. Since this is a major version, there are some minor breaking changes, especially with how schemas are loaded. Please check out the updated docs here and let me know if this fixes your issue: https://v7.openapi-ts.pages.dev/node/

packages/openapi-typescript/tsconfig.examples.json
```diff
@@ -0,0 +1,5 @@
+{
+  "extends": "./tsconfig.json",
+  "include": ["examples"],
+  "exclude": ["examples/digital-ocean-api.ts"]
```
drwpow, line 4: An overdue addition to v7 is enforcing that the generated snapshots are valid TypeScript. Before this more-or-less happened manually, but now it’s enforced in CI.

However, the DigitalOcean schema is a bit of a love–hate relationship. On the one hand, it does really stress-test many of the features that make this library robust and flexible. But on the other, the schema itself has errors 🙃.

Unfortunately, this means we can’t use DigitalOcean’s schema to check errors until they themselves resolve the problems with their schema. But I still like having it around for snapshot testing to just make sure this library _can_ generate _something_ there without blowing up.

drwpow:
The upcoming v7 will require more testing before release (release date TBD), but I’m going to merge this into `main`, and fork the current `main` into `6.x` for bugfixes and security updates.

PR #1364, opened 2023-10-05
Prevent using request body before cloning it

Fumaz:
## Changes

This PR aims to fix a bug where the library would read the response's body before cloning it, which would result in the cloned response having an empty body after being cloned (at least in Bun).

## How to Review

This is a pretty minor fix, so I don't think there's any specific things to keep in mind when reviewing the changes.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/src/index.ts
+++ b/packages/openapi-fetch/src/index.ts
@@ -68,10 +68,12 @@ export default function createClient<Paths extends {}>(clientOptions: ClientOpti
 
     // parse response (falling back to .text() when necessary)
     if (response.ok) {
-      let data: any = response.body;
+      let data: any;
       if (parseAs !== "stream") {
         const cloned = response.clone();
         data = typeof cloned[parseAs] === "function" ? await cloned[parseAs]() : await cloned.text();
+      } else {
+        data = response.clone().body;
       }
       return { data, response: response as any };
     }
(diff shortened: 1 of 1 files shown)


drwpow:
I’m in favor of this fix, but if this behavior is testable would it be possible to add a test (not actually loading Bun; just testing the desired behavior)? Or if not, maybe add a comment in the code explaining what to be aware of

Fumaz:
i didn't think this would require tests or comments since it seems like a pretty trivial fix, but it being an edge case (if its really an issue with bun only) might warrant a comment. if you'd like, i can add a comment explaining why this is necessary.

drwpow:
The major reasons for tests or comments is preventing regressions, and this change is the type of change that probably wouldn’t last long because there’s no clear intent why an empty variable would be initialized without a value without any explanation. It almost seems like a fault of style than an intentional line of code. A comment could go a long way.

But that said, the more I think about this, the more I think having Bun tests (even just basic “does this work” tests) would be valuable in general, so I’ll probably set that up either way, which makes the request somewhat moot. But until then, yes, a comment would help immensely 🙏

drwpow (approved):

Fumaz:
let me know if this looks good to you!

drwpow:
Looks great, thanks! Just the tip-off that this logic exists for Bun would prevent a future PR of someone unfamiliar with the code to change it without trying it in Bun first (hopefully). But again, will look into adding Bun tests.

Issue #1127, opened 2023-03-14
Optionless endpoints still require 2nd param


drwpow:
For simple endpoints that don’t require any parameters or requestBodies, the 2nd param is still required:

```ts
import { paths } from './v1';

const { get } = createClient<paths>();
get('/self');
// ~~~~~~~~~~
// ^ Expected 2 parameters, but got 1.
```

The solution is adding an empty `{}` to the call, like so:

```diff
- get('/self');
+ get('/self', {});
```

This is because it’s an intersection of `{ params: … } & { body: … } & RequestInit`, which are all required options, but in some scenarios end up having all keys optional which results in an empty object (`{}`). However, if we make the root types always optional, then the types break, and this library would be pointless.

If there’s a way to solve this without jeopardizing the types, PRs are welcome!

Though note that **this is only a cosmetic issue.** There are no actual performance or runtime bugs with providing an empty object as the 2nd param.

**Counterpoint to fixing**: Many APIs won’t have any endpoints with absolutely zero parameters. For the schemas that do, only a few endpoints realistically will deal with this. Further, simply having the 2nd param there can reduce Git diff noise if/when options are ever needed to be passed to that endpoint.

PR #1366, opened 2023-10-05
Fix empty object required 2nd param

drwpow:
## Changes

Fixes the elusive/annoying #1127. Now, rather than `.GET('/url', {})` being required, you can just write `.GET('/url')`. Minor QoL improvement, but very important fix!

## How to Review

Tests should pass.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/src/index.test.ts
+++ b/packages/openapi-fetch/src/index.test.ts
@@ -48,7 +48,7 @@ describe("client", () => {
 
       // data
       mockFetchOnce({ status: 200, body: JSON.stringify(["one", "two", "three"]) });
-      const dataRes = await client.GET("/string-array", {});
+      const dataRes = await client.GET("/string-array");
 
       // … is initially possibly undefined
       // @ts-expect-error
@@ -67,7 +67,7 @@ describe("client", () => {
 
       // error
       mockFetchOnce({ status: 500, body: JSON.stringify({ code: 500, message: "Something went wrong" }) });
-      const errorRes = await client.GET("/string-array", {});
+      const errorRes = await client.GET("/string-array");
 
       // … is initially possibly undefined
       // @ts-expect-error
@@ -92,7 +92,7 @@ describe("client", () => {
 
         // expect error on missing 'params'
         // @ts-expect-error
-        await client.GET("/blogposts/{post_id}", {});
+        await client.GET("/blogposts/{post_id}");
 
         // expect error on empty params
         // @ts-expect-error
@@ -120,7 +120,7 @@ describe("client", () => {
 
         // expet error on missing header
         // @ts-expect-error
-        await client.GET("/header-params", {});
+        await client.GET("/header-params");
 
         // expect error on incorrect header
         // @ts-expect-error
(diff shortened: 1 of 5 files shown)


packages/openapi-fetch/src/index.test.ts
```diff
@@ -48,7 +48,7 @@ describe("client", () => {
 
       // data
       mockFetchOnce({ status: 200, body: JSON.stringify(["one", "two", "three"]) });
-      const dataRes = await client.GET("/string-array", {});
+      const dataRes = await client.GET("/string-array");
```
drwpow, line 51: Note: the removal of all `, {})` in the test suite is a good sign. The test suite has quite a few `// @ts-expect-error` comments which are vital to testing the type assertions.

But in addition, the test suite also has runtime tests to ensure the function is behaving as-expected. The fact that no tests were modified, while still passing, is a great sign.

packages/openapi-fetch/src/index.ts
```diff
@@ -89,38 +93,55 @@ export default function createClient<Paths extends {}>(clientOptions: ClientOpti
...
+  type HeadFetchOptions<P extends HeadPaths> = FetchOptions<FilterKeys<Paths[P], "head">>;
+  type PatchFetchOptions<P extends PatchPaths> = FetchOptions<FilterKeys<Paths[P], "patch">>;
+  type TraceFetchOptions<P extends TracePaths> = FetchOptions<FilterKeys<Paths[P], "trace">>;
+
   return {
     /** Call a GET endpoint */
-    async GET<P extends PathsWithMethod<Paths, "get">>(url: P, init: FetchOptions<FilterKeys<Paths[P], "get">>) {
-      return coreFetch<P, "get">(url, { ...init, method: "GET" } as any);
+    async GET<P extends GetPaths>(url: P, ...init: HasRequiredKeys<GetFetchOptions<P>> extends never ? [GetFetchOptions<P>?] : [GetFetchOptions<P>]) {
+      return coreFetch<P, "get">(url, { ...init[0], method: "GET" } as any);
```
drwpow, line 116: The fix was weird, and not straightforward. Essentially, it involved multiple complicated steps:

1. First, convert the function parameters into an array (`...init`) (though for simplicity, we can do this starting at the 2nd param, rather than having to do it for the full spread)
2. Next, we’ll need to take each method’s individual FetchOptions and introspect whether **it has a single required key.** (we needed to create a helper for this)
3. Lastly, expand the parameter array into a tuple type, marking it either optional (`[T?]`) or not (`[T]`) based on the result of #2

drwpow: Again, all of this is very annoying, very strange TypeScript introspection. But that’s the name of the game here, so it’s all worth it… I guess.

packages/openapi-fetch/src/index.ts
```diff
@@ -145,8 +166,8 @@ export function defaultBodySerializer<T>(body: T): string {
 }
 
 /** Construct URL string from baseUrl and handle path and query params */
-export function createFinalURL<O>(url: string, options: { baseUrl?: string; params: { query?: Record<string, unknown>; path?: Record<string, unknown> }; querySerializer: QuerySerializer<O> }): string {
-  let finalURL = `${options.baseUrl ? options.baseUrl.replace(TRAILING_SLASH_RE, "") : ""}${url as string}`;
```
drwpow, line 149: Small micro-optimization: we don’t need to be running a dynamic RegEx to remove a trailing slash. It’s much more performant to just remove a single character if needed (done above on `createClient()`, also to save cycles on running a RegEx every request

PR #1367, opened 2023-10-05
Add Bun tests to CI

drwpow:
## Changes

CI only: run the test suite with Bun to ensure it’s working as expected (it’s only executing Vitest with Bun rather than Node, but will still flag any glaring issues)

## Che

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,8 +48,13 @@ jobs:
       - uses: pnpm/action-setup@v2
         with:
           version: latest
+      - uses: oven-sh/setup-bun@v1
+        with:
+          bun-version: latest
       - run: pnpm i
       - run: pnpm test
+      # ! Important: `bun test` != `bun run test`. We want the latter.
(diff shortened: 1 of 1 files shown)


drwpow:
@Fumaz 🤔 I added Bun testing to CI, however, when I revert #1364 locally, it passes all the same. I’m not able to recreate the error you were seeing that you were executing in Bun.

Is there something we can add here to recreate the error you were seeing originally?

Fumaz:
Honestly I'm not entirely sure how to reproduce. I had to debug a lot to figure out why the body was always empty by the time it got to the parsing function, and that's how I found out that it was causing that issue. I don't think it matters but I am using my own API framework in my project, which uses Bun too.

drwpow:
I’m going to punt on this for now. This just slows down CI without any clear gain. As far as I’ve tested, openapi-fetch works with Bun just fine. We can re-add these, possibly using Bun’s native test suite, at a future date when I have time.

PR #1373, opened 2023-10-06
feat: Custom fetch per call

### LINKED ISSUE #1369: Custom fetch support per call
**Description**

SvelteKit provides a custom fetch function for loaders(which provide data for components) with a bunch of QoL improvements - this fetch functions are specific to each call so they can't be provided on `createClient` call. Would be nice to have an option to provide custom fetch function per specific openapi-fetch call.

**Proposal**

Extent FetchOptions type with a fetch parameter akin to ClientOptions - default to createClient specified fetch if not provided.

**Checklist**

- [x] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-fetch/blob/main/CONTRIBUTING.md))

HugeLetters:
## Changes

Closes #1369

## How to Review

Adds the option to pass in custom fetch as fetchOption on per API call basis. Added a simple test to see that custom fetch overrides the one specified during client initialization.

## Checklist

- [x] Unit tests updated


HugeLetters:
@drwpow since this change was directly inspired by SvelteKit I would like to add some additional info to docs about making openapi-fetch work SvelteKit. The issue is not directly related to a custom fetch so perhaps you would prefer a separate PR?

Since your library reads content-length header when parsing a response, if you fetch during SSR SvelteKit will error because it doesn't serialize headers by default. 
The fix is rather simple and an error message tells you almost everything you need to know to fix it but I thought adding a small section with a code example in the docs would be nice.
image


The fix is this
```ts
// hooks.server.ts
import type { Handle } from '@sveltejs/kit';
export const handle: Handle = async ({ event, resolve }) => {
	return resolve(event, {
		filterSerializedResponseHeaders(name) {
			return name === 'content-length';
		}
	});
};
Can I add this example to this docs in this same PR?

**Comment by drwpow** [comment]
Docs improvements are always appreciated!

Maybe rather than having that snippet live just in docs, though, putting it in the `examples/sveltekit` example code itself would mean people don’t miss it when they’re setting up their project. Inline comments can be added if explanation is needed. Beyond that, I’d probably omit it in the docs unless there’s something critical for the reader to understand.

**Comment by drwpow** [comment]
But re: not serializing headers in SSR, that’s interesting! I’d be open to revisiting that default behavior after understanding why SvelteKit does that better (couldn’t find the “why” in a quick search, just other libraries requiring similar workarounds)

**Comment by HugeLetters** [comment]
@drwpow I've updated the examples

**Review by drwpow — APPROVED** [review]
This looks great, thank you!

Before releasing this version, though, I want to do a little testing first to make sure the updated types generated from openapi-typescript@7 didn’t introduce any accidental breaking changes from openapi-fetch (it shouldn’t have, but I need to double-check).

**Comment by drwpow** [comment]
One last thing—could you add a `patch` changeset to describe the feature? See the comment on this PR for instructions.

File: .changeset/wise-poems-do.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+"openapi-fetch": minor
```
drwpow, line 2:
```suggestion
"openapi-fetch": patch
```

For pre-1.0 versions, minor versions should indicate breaking changes. A patch would be better here, as requested

drwpow:
Thank you again!

Issue #1368, opened 2023-10-06
v7 Preview 🎉 


drwpow:
## v7 Preview

`openapi-typescript@7` is now in preview! 🎉 The major additions to this library are:

- ✨ Built-in validation using the [Redoc CLI](https://redocly.com/docs/cli/commands/lint/) — no more needing an external tool 
- ✨ Respecting `redocly.yaml` for API config (define your schemas there in one place, and use `openapi-typescript` to generate TypeScript without having to manage a separate config)
- ✨ Addition of the `--enum` flag for generating true TypeScript enums instead of unions (default behavior still unions)
- 📘 A mini-rewrite to use the TypeScript AST instead of string mashing

<details>
  <summary><h3>Full Changelog</h3></summary>

### Features

- ✨ **Feature**: automatically validate schemas with Redocly CLI ([docs](https://redocly.com/docs/cli/)). No more need for external tools to report errors! 🎉

    - By default, it will only throw on actual schema errors (uses Redocly’s default settings)
    - For stricter linting or custom rules, you can create a [redocly.yaml config](https://redocly.com/docs/cli/configuration/)
- ✨ **Feature**: bundle schemas with Redocly CLI
  - Any options passed into your [redocly.yaml config](https://redocly.com/docs/cli/configuration/) are respected
- ✨ **Feature**: add `enum` option to export top-level enums from schemas (#941)
- ✨ **Feature**: add `formatOptions` to allow formatting TS output
- ✨ **Feature**: header responses add `[key: string]: unknown` index type to allow for additional untyped headers
- ✨ **Feature:** allow configuration of schemas via `apis` key in redocly.config.yaml. [See docs](https://openapi-ts.pages.dev/cli/) for more info.

### Breaking Changes

-  ⚠️ **Breaking**: The Node.js API now returns the TypeScript AST for the main method as well as `transform()` and `postTransform()`. To migrate, you’ll have to use the `typescript` compiler API:

    ```diff
    + import ts from "typescript";

    + const DATE = ts.factory.createIdentifier("Date");
    + const NULL = ts.factory.createLiteralTypeNode(ts.factory.createNull());
  
      const ast = await openapiTS(mySchema, {
        transform(schemaObject, metadata) {
        if (schemaObject.format === "date-time") {
    -       return schemaObject.nullable ? "Date | null" : "Date";
    +       return schemaObject.nullable
    +         ? ts.factory.createUnionTypeNode([DATE, NULL])
    +         : DATE;
[...]

PR #1383, opened 2023-10-09
Improve composition

drwpow:
## Changes

Fixes `allOf` bug raised in #1368, and adds tests for this improvement in behavior. No changeset because this is likely a regression caused from the rewrite.

## How to Review

Tests should pass

## Checklist

- [x] Unit tests updated


packages/openapi-typescript/src/transform/schema-object.ts
```diff
@@ -163,31 +199,25 @@ export function transformSchemaObjectWithComposition(
...
+    const allOf: ts.TypeNode | undefined = allOfType.length
       ? tsIntersection(allOfType)
       : undefined;
-    // add required props
-    if (
-      allOf &&
-      "required" in schemaObject &&
-      Array.isArray(schemaObject.required)
-    ) {
-      allOf = tsWithRequired(
```
drwpow, line 177: Change: don’t indiscriminately add `WithRequired<…>`; instead, only add it when it’s actually needed, and with the proper keys

packages/openapi-typescript/src/transform/schema-object.ts
```diff
@@ -142,10 +142,46 @@ export function transformSchemaObjectWithComposition(
...
+
+        const resolved = options.ctx.resolve<SchemaObject>(item.$ref);
+        if (
+          resolved &&
+          typeof resolved === "object" &&
+          "properties" in resolved
+        ) {
+          // don’t try and make keys required if the $ref doesn’t have them
+          const validRequired = (required ?? []).filter(
+            (key) => !!resolved.properties![key],
```
drwpow, line 163: This is probably brittle across multiple nested compositions, but that’s OK for now. I don’t think that’s a common scenario to need to specify `required` keys on a different schema object from a deeply-nested `allOf` chain

packages/openapi-typescript/test/transform/schema-object/composition.test.ts
```diff
@@ -469,15 +469,66 @@ describe("composition", () => {
...
 } & {
-    green?: number;
-}, "red" | "blue" | "green">`,
+    green: number;
+}`,
         // options: DEFAULT_OPTIONS,
       },
     ],
+    [
+      "allOf > core properties",
```
drwpow, line 482: New test

PR #1413, opened 2023-10-30
Proposal to generate a separate interface for each omponents.schemas model

alex-aveva:
Provide public interface for each model defined in components.schemas


drwpow:
Thanks for opening @alex-aveva. In the future, this is better as an **Issue** rather than a Pull Request, so we can discuss.

This is currently being proposed already in #1260, so please give any thoughts/comments/feedback there.

PR #1417, opened 2023-11-01
Add questionToken option to transform

### LINKED ISSUE #1411: Transform nullable returns | null instead of optional ?
**Description**

I'm following the example from https://openapi-ts.pages.dev/node/, but the type becomes `file: Blob | null` instead of `file?: Blob`. Maybe this is the desired result, in that case, how can I make it optional with the question mark notation? I tried appending undefined as well, but that results in `file: Blob | undefined` which is also not what I'm looking for 😄 

| `openapi-typescript` | `7.0.0-next.2`                      |
| Node.js              | `18.x.x`                      |
| OS + version         | `Windows 11` |

**Reproduction**
```
import openapiTS from "openapi-typescript";
import ts from "typescript";

const BLOB = ts.factory.createIdentifier("Blob"); // `Blob`
const NULL = ts.factory.createLiteralTypeNode(ts.factory.createNull()); // `null`

const ast = await openapiTS(mySchema, {
  transform(schemaObject, metadata) {
    if (schemaObject.format === "binary") {
      return schemaObject.nullable
        ? ts.factory.createUnionTypeNode([BLOB, NULL])
        : BLOB;
    }
  },
});
```

**Expected result**
 `file?: Blob`

**Checklist**
- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

JorrinKievit:
## Changes

This adds the possibility of appending a questionToken to a property using the transform function. Fixes #1411 

## How to Review
I am not 100% confident I covered all use cases of the transform function. Keep in mind that with the limited knowledge of this library, I might have missed some things.

## Checklist

- [X] Unit tests updated
- [X] `docs/` updated (if necessary)

### DIFF
--- a/docs/src/content/docs/node.md
+++ b/docs/src/content/docs/node.md
@@ -118,7 +118,7 @@ That would result in the following change:
 
 ```diff
 - updated_at?: string;
-+ updated_at?: Date;
++ updated_at: Date | null;
 ```
 
 #### Example: `Blob` types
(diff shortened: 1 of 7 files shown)


docs/src/content/docs/node.md
```diff
@@ -158,7 +158,51 @@ Resultant diff with correctly-typed `file` property:
...
+    }
+  },
+});
+```
+
+Resultant diff with correctly-typed `file` property and "?" token:
+
+```diff
+- file: Blob;
++ file?: Blob | null;
```
drwpow, line 205: 👍 The docs look great! Thank you.

packages/openapi-typescript/src/types.ts
```diff
@@ -467,6 +467,11 @@ export type SchemaObject = {
   | {}
 );
 
+export interface TransformObject {
```
drwpow, line 470: 👍 This is great.

We can probably extend this pretty far before having to change this API.

packages/openapi-typescript/test/node-api.test.ts
```diff
@@ -382,9 +386,49 @@ export type operations = Record<string, never>;`,
...
-                ts.factory.createIdentifier("Date"),
-              );
+              return DATE;
+            }
+          },
+        },
+      },
+    ],
+    [
+      "options > transform with schema object",
```
drwpow, line 396: 💯 Fantastic tests!

packages/openapi-typescript/src/transform/components-object.ts
```diff
@@ -51,14 +53,31 @@ export default function transformComponentsObject(
     const items: ts.TypeElement[] = [];
     if (componentsObject[key]) {
       for (const [name, item] of getEntries(componentsObject[key], ctx)) {
-        const subType = transformers[key](item, {
+        let subType = transformers[key](item, {
```
drwpow, line 56: Yup I think this will work in all cases I can think of. Just handling keys on object types should be the main usecase here (we don’t ever want to have question tokens, for example, directly on the `components["schemas"]` object as that doesn’t make sense)

drwpow (approved):
> I am not 100% confident I covered all use cases of the transform function. Keep in mind that with the limited knowledge of this library, I might have missed some things.

All the tests pass, including the ones you added! So as far as I’m concerned, you handled all cases perfectly 🙂 (and if you missed something, then that’s the test suite’s fault and not yours)

drwpow:
Also the changeset can be ignored in this PR, as this will go into the v7 (beta) release (i.e. there’s no version to bump yet). But I’ll update the release notes for v7 and give you credit.

PR #1419, opened 2023-11-02
Update examples.md

WickyNilliams:
## Changes

adds example of using getter for dynamic header with openapi-fetch

## How to Review

eyeball code, make sure it looks ok

## Checklist

- [x] Unit tests updated

### DIFF
--- a/docs/src/content/docs/openapi-fetch/examples.md
+++ b/docs/src/content/docs/openapi-fetch/examples.md
@@ -72,6 +72,37 @@ client.GET("/some-authenticated-url", {
 });
 ```
 
+### Vanilla JS getter
+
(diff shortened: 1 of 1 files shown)


WickyNilliams:
i just stumbled on this approach when adding such functionality to my own codebase. thought it might be useful in the docs as it feels simpler than the existing examples. seems to work just fine in my testing

drwpow (approved):
Code looks great! Whoa this is much smarter than some of the other examples! Now I’m tempted to delete the other ones in favor of this one 😄

WickyNilliams:
Might be worth keeping them, but perhaps reorder so this is presented first :)

PR #1424, opened 2023-11-03
Manual type declarations in openapi-fetch

drwpow:
## Changes

Splits `index.ts` (runtime + types) into manual `index.js` and `index.d.ts` files. Added this section to the CONTRIBUTING docs to explain why:

>This project uses **manual type declarations** ([docs](https://www.typescriptlang.org/docs/handbook/2/type-declarations.html#dts-files)) that are kept separately from the runtime code. `index.d.ts` contains **type declarations**, and `index.js` contains **runtime code**. The most important code lives in `index.d.ts` because this library exists to provide correct type inference. `index.js`, is far less important, and is only doing bare-minimum work to support the API as efficiently as possible.
>
>In most projects, **this is not recommended practice** as the two can (and will) diverge, and you’re left with types that “lie” to you about what the runtime is doing. So this project does have that liability. However, due to the unique nature of this project implementing complex type inference from user-provided schemas, the type inference itself is so complex it was interfering with readable, maintainable, runtime code. By separating the two we won’t have the more complex parts of TypeScript interfering with the optimal runtime code. This would not be tenable in a larger project, but is perfect for a small codebase like this with minimal surface area.
>
>When writing code, it’s tempting to ignore `index.d.ts` and only make runtime updates, since that is generally simpler. But **please don’t!** We write tests in `*.test.ts` files intentionally so that the type inference can be typechecked as well as the runtime. Whenever you make a change to `index.js`, you probably need to also make a chance to `index.d.ts`, too. And **testing type correctness in tests is just as important as testing the runtime!**

This is an internal change—tests still pass—but it will make some upcoming changes much, much easier.

## How to Review

- Tests should pass

## Checklist

- [x] Unit tests updated


packages/openapi-fetch/test/index.test.ts
```diff
@@ -3,7 +3,7 @@ import { afterEach, beforeAll, describe, expect, it, vi } from "vitest";
 // @ts-expect-error
 import createFetchMock from "vitest-fetch-mock";
 import createClient from "../src/index.js";
-import type { paths } from "./v1.d.js";
+import type { paths } from "./fixtures/api.js";
```
drwpow, line 6: Moved into `/fixtures/` folder for cleanup

packages/openapi-fetch/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "lib": ["ESNext", "DOM"],
```
drwpow, line 7: Fun fact: I originally wanted to add `allowJs: true` to TSConfig, however, that broke everything—it tries to infer types from `.js` files and ignores `.d.ts`. So fortunately no change is necessary, and TypeScript likes `.d.ts` just as much as the original `.ts`

packages/openapi-fetch/CONTRIBUTING.md
```diff
@@ -69,3 +47,45 @@ To start the entire test suite in watch mode:
...
+
+The changelog is generated via [changesets](https://github.com/changesets/changesets), and is separate from Git commit messages and pull request titles. To write a human-readable changelog for your changes, run:
+
+```
+npx changeset
+```
+
+This will ask if it’s a `patch`, `minor`, or `major` change ([semver](https://semver.org/)), along with a plain description of what you did. Commit this new file along with the rest of your PR, and during the next release this will go into the official changelog!
+
+## Opening a Pull Request
```
drwpow, line 67: Not actually rewritten; just moved several sections out of the “Opening a Pull Request” heading and into a “Writing Code” heading

PR #1428, opened 2023-11-05
[feat]: parseAs options reports correct data type

### LINKED ISSUE #1370: parseAs doesn't change the data type
**Description**

Specifying `parseAs` as "string" or even "stream" still specifies `data` as JSON 

**Proposal**

A new generic parameter could be added to `coreFetch` to infer correct return type. 
- Either we modify `FetchResponse`,`FetchOptions` and the types it uses accordingly. This would technically be breaking for people who use these types in their codebase. We can instead of adding a third type parameter just modify the second one - something like `O extends {method: HttpMethod, parseAs: ParseAs}`.
- We could try to minimize the friction for people consuming these types by adding a third type parameter to `coreFetch`, then adding a second optional type parameter with a default value for `FetchOptions` and `RequestOptions`. And then also modify the return type of `coreFetch` directly instead of modifying `FetchResponse` type - something like 
```ts
Promise<parseAs extends "json" ? FetchResponse<M>, { data: ReturnType<Response[parseAs]> }>
```
First approach seems more "correct" but is breaking in a sense. The second one feels like a band-aid but shouldn't affect users.

**Checklist**

HugeLetters:
## Changes

Closes #1370

## How to Review

Improves typing of fetch function by introducing a fetchOptions generic which can then be used to report correct return type.

## Checklist

- [x] Unit tests updated 

### DIFF
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -39,14 +39,14 @@ export type QuerySerializer<T> = (
 export type BodySerializer<T> = (body: OperationRequestBodyContent<T>) => any;
 
 export type ParseAs = "json" | "text" | "blob" | "arrayBuffer" | "stream";
-export type ParseAsResponse<T, K> = K extends ParseAs
-  ? {
-      json: T;
-      text: Awaited<ReturnType<Response["text"]>>;
-      blob: Awaited<ReturnType<Response["blob"]>>;
-      arrayBuffer: Awaited<ReturnType<Response["arrayBuffer"]>>;
-      stream: Response["body"];
-    }[K]
+export type ParseAsResponse<T, K extends ParseAs> = K extends "text"
+  ? Awaited<ReturnType<Response["text"]>>
+  : K extends "blob"
+  ? Awaited<ReturnType<Response["blob"]>>
+  : K extends "arrayBuffer"
+  ? Awaited<ReturnType<Response["arrayBuffer"]>>
+  : K extends "stream"
+  ? Awaited<ReturnType<Response["body"]>>
   : T;
 
 export interface DefaultParamsOption {
@@ -71,11 +71,19 @@ export type RequestBodyOption<T> = OperationRequestBodyContent<T> extends never
 
 export type FetchOptions<T> = RequestOptions<T> & Omit<RequestInit, "body">;
 
-export type FetchResponse<T, O extends FetchOptions<any>> =
+/** This type helper makes the 2nd function param required if params/requestBody are required; otherwise, optional */
+export type MaybeOptionalInit<
+  P extends {},
+  M extends keyof P,
+> = HasRequiredKeys<FetchOptions<FilterKeys<P, M>>> extends never
+  ? [(FetchOptions<FilterKeys<P, M>> | undefined)?]
+  : [FetchOptions<FilterKeys<P, M>>];
+
+export type FetchResponse<T, R extends ParseAs = "json"> =
   | {
       data: ParseAsResponse<
         FilterKeys<SuccessResponse<ResponseObjectMap<T>>, MediaType>,
-        O["parseAs"]
+        R
       >;
       error?: never;
       response: Response;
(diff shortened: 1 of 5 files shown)


HugeLetters:
@drwpow I've made a draft PR which only changes GET requests - seem to work fine. Could you check if it looks fine to you? If yes, I'll update to include over http verbs too.

In general I think this change makes API more flexible in the future since any option property can be used now to modify return type.

packages/openapi-fetch/src/index.d.ts
```diff
@@ -39,6 +39,15 @@ export type QuerySerializer<T> = (
...
 
 export type ParseAs = "json" | "text" | "blob" | "arrayBuffer" | "stream";
+export type ParseAsResponse<T, K> = K extends ParseAs
+  ? {
+      json: T;
+      text: Awaited<ReturnType<Response["text"]>>;
+      blob: Awaited<ReturnType<Response["blob"]>>;
+      arrayBuffer: Awaited<ReturnType<Response["arrayBuffer"]>>;
+      stream: Response["body"];
+    }[K]
```
drwpow, line 49: 👍 This is a smart pattern. We could probably use this in other places to simplify some deeply-nested ternaries (maybe).

drwpow:
> Could you check if it looks fine to you? If yes, I'll update to include over http verbs too.

`ParseAsResponse` looks great to me—those look like the correct types. This looks like the right direction—now it’s just plugging everything up and making sure there aren’t any regressions in the test suite where `parseAs` is still the default (JSON)

HugeLetters:
@drwpow ok, I'm a bit stumped at the moment - this pattern doesn't seem to work for some requests which require body/params. I've been able to pinpoint the root of the issue somewhat but have no idea what's causing it - seems like non-string type in request schemas cause it to not work properly(e.g. a string literlal type(enums) or a number). Here's a video with an example - I can also link my schema file here, I'm not using the one from the examples but the one from my project since it has routes for other verbs.
[Screencast from 06-11-23 01:56:24.webm](https://github.com/drwpow/openapi-typescript/assets/119697239/9c3db7d8-282d-4c24-b7f3-ddcd539dd091)

drwpow:
Yeah this is a tricky problem for sure. Will dig in later this week; this will take some investigation

drwpow:
This is mostly unrelated, but problems like this are why in openapi-typescript@7 there are a lot more `requestBody?: never` shapes (and the like) because it reduces the inference from hundreds of possible permutations of requests to a much smaller number. My fear was for more complicated things like this, TypeScript would just fail when its inference was trying to create unions internally that were too complex for it.

Not saying that 7.x will solve this problem for us or anything—it’s the same problem here either way—but just saying I’ll probably first try and solve this in the 7.x schema types since it theoretically should be simpler, then backport it to 6.x types.

HugeLetters:
@drwpow sure, if you think it will be much easier to work with new types I can wait till these changes are made and make my PR around them. 
I don't think this particular parseAs is urgent for anyone anyways - the more important change of this PR probably is transition to options generic which would allow to add new options in the future which would also allow to modify return type in some new ways.

drwpow:
OK finally got some time to take a crack at it in #1442. In general it’s hard to explain what I did, and there’s also some arbitrary cleanup I did that just made it easier for me to debug, but didn’t actually contribute to the solution.

But the short version is your PR laid the correct groundwork but the inference was getting lost from the `GET()` function to the return type. If `...init` has type `O`, and we try and pass that as a generic argument all the way down (`FetchResponse<T, O>`), it’s really easy for TypeScript (or us) to get confused at the difference between `init`’s **generic typing** versus its **runtime value.** Often TS will try and reuse generic typing because it’s cheaper.

Instead, we want `FetchResponse<T, (typeof init)['parseAs']>` because this essentially has to infer the actual value of `init.parseAs`, and it will pass that down through all the helpers correctly.

I’m not saying that there’s _not_ a way to make `O` work here, but the way to do it involves stacking complex inferences in a way that’s tough to write and tough to maintain. But saving a primitive value is much easier to deal with. And that’s all my PR did—just flattened `parseAs` into its runtime value, and passed that along. The rest of the changes were, like I said, just cleanup that made the problem easier to see, and hopefully cleaned up some duplication.

I think we’ll have to test it a little more, but it’s close!

drwpow:
Anyway, feel free to cherrypick #1442 and/or pull that into your PR, and test it some more on your end.

packages/openapi-fetch/test/index.test.ts
```diff
@@ -605,12 +605,13 @@ describe("client", () => {
...
-        const { data } = await client.GET("/anyMethod", {
+        const { data } = (await client.GET("/anyMethod", {
           parseAs: "stream",
-        });
+        })) satisfies { data?: ReadableStream<Uint8Array> | null };
         if (!data) {
           throw new Error(`parseAs stream: error`);
         }
+        // todo - not sure what test to put here - previously it just checked for instanceof Buffer
         expect(data.byteLength).toBe(8);
```
HugeLetters, line 615: Not sure what test to put here - previously it just checked for instanceof Buffer

HugeLetters:
Ok, so the current iteration passes my mini test-suite I've set up with a more complicated openapi schema - I'll give it a day or two to try to make it work with an Options object as I was trying before. 
If not - I think it's okay as is.

packages/openapi-fetch/test/index.test.ts
drwpow: Oh that can change to whatever. I just wanted to have a type test because usually TS will let you check for `instanceof` on anything (that’s a runtime test). Really the `expect()` here is useless; it’s more for the test reader it’s an assertion. But any check will work where we try and access something off the `data` type immediately without guiding the inference first (which `instanceof` will do)

HugeLetters: @drwpow There's also this problem which I'm not sure we even need to address
https://github.com/node-fetch/node-fetch/issues/1308

The reported type from node's fetch(they use undici I think but I assume it's the same issue as with node-fetch) is ReadableStream as per browser spec but in nodejs it could either be a stream or a buffer.

HugeLetters: My proposed solution would be this
https://github.com/drwpow/openapi-typescript/pull/1428/commits/762efcab7be08753ea6765eff5b3b48954382da9

packages/openapi-typescript-helpers/index.d.ts
```diff
@@ -77,9 +77,7 @@ export type ErrorResponse<T> = FilterKeys<
 // Generic TS utils
 
 /** Find first match of multiple keys */
-export type FilterKeys<Obj, Matchers> = {
-  [K in keyof Obj]: K extends Matchers ? Obj[K] : never;
-}[keyof Obj];
+export type FilterKeys<Obj, Matchers> = Obj[keyof Obj & Matchers];
```
HugeLetters, line 80: Ok, so I've narrowed down the previous issue with literal types breaking inference to this type - this commit doesn't fix that yet.
But it's definitely the culprit

https://github.com/drwpow/openapi-typescript/pull/1428/files#diff-b38a8955c23641d93d6884093dd49817f791154bd4dbe0e979b8d359bcf17e35R68

If you change this line to just then suddenly it works - don't know why tho yet!
```ts
OperationRequestBodyMediaContent<T>[MediaType]
```

packages/openapi-fetch/test/index.test.ts
drwpow: Ah I see what you mean. Yeah, I agree with your approach. In general, this isn’t the only place where Node.js and browser types diverge, so the less work we can do here the better. This is also another advantage of us manually owning the types now—sometimes when building a library, your TSConfig settings change the generated types and you accidentally bake in Node.js types for a browser package or vice-versa. But now, there are no type surprises, but even if there are we can patch it without affecting runtime code.

HugeLetters:
This is pure insanity but I think I got it solved at least, haha

https://github.com/drwpow/openapi-typescript/assets/119697239/9e66c82f-3460-4fa5-851c-ae0fbdcf0860

drwpow:
> This is pure insanity but I think I got it solved at least, haha
> leproblem.mp4

wait—why is `const` necessary here? are those 2 strings an `enum`?

HugeLetters:
> > This is pure insanity but I think I got it solved at least, haha
> > leproblem.mp4
> 
> wait—why is `const` necessary here? are those 2 fields an `enum` (edit: in the OpenAPI schema)?

Yup, those two fields are enums - seems like TS inference isn't working properly here and it thinks `grade` and `title` are just `string` in provided object unless I put `as const` on them. At least now we know it's not the issue on our end. 
Extracting the value to a variable and annotating its type or just asserting the full type on an inline object works too .

A friend of mine showed me this and I think it's a related issue.

https://github.com/drwpow/openapi-typescript/assets/119697239/6e966ba3-c6ad-4c36-bdd8-b8d08829f159

drwpow:
Sorry—was there anything else you wanted to get in before merging? I think this looks great as-is, and happy to merge and kick the tires some more. But wanted to double-check just in case.

HugeLetters:
@drwpow Hi, yup, I'm done

drwpow (approved):

PR #1430, opened 2023-11-06
Add Node 21 to CI

drwpow:
## Changes

CI-only change: Add Node 21.x to CI to test upcoming compatibility for when 22.x gets a stable release.

Will keep `macos` and `windows` locked to 20.x because those runners are expensive, and GitHub doesn’t queue those up too many (tried to expand the matrix to include these, but it just blocked the CI pipeline for a half hour)

## How to Review

- CI should pass (🤞)

## Checklist

N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [18.x, 20.x]
+        node-version: [18.x, 20.x, 21.x]
     steps:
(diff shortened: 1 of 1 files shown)


drwpow:
🤔 I think the Cloudflare failure (docs site) is unrelated to this; I think it cached something it shouldn‘t have in the preview runner (not 100% sure, but not seeing that error locally). In either case, will fix separately from this PR.

drwpow:
Update: Cloudflare failure is from an upstream docs dependency bug unrelated to this PR or this library; will fix separately.

PR #1442, opened 2023-11-18
[openapi-fetch] Support parseAs type inference

drwpow:
## Changes

Adds parseAs inference; continues from #1428

## How to Review

- N/A

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -39,14 +39,14 @@ export type QuerySerializer<T> = (
 export type BodySerializer<T> = (body: OperationRequestBodyContent<T>) => any;
 
 export type ParseAs = "json" | "text" | "blob" | "arrayBuffer" | "stream";
-export type ParseAsResponse<T, K> = K extends ParseAs
-  ? {
-      json: T;
-      text: Awaited<ReturnType<Response["text"]>>;
-      blob: Awaited<ReturnType<Response["blob"]>>;
-      arrayBuffer: Awaited<ReturnType<Response["arrayBuffer"]>>;
-      stream: Response["body"];
-    }[K]
+export type ParseAsResponse<T, K extends ParseAs> = K extends "text"
+  ? Awaited<ReturnType<Response["text"]>>
+  : K extends "blob"
+  ? Awaited<ReturnType<Response["blob"]>>
+  : K extends "arrayBuffer"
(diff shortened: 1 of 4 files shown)


packages/openapi-fetch/src/index.d.ts
```diff
@@ -39,6 +39,15 @@ export type QuerySerializer<T> = (
 export type BodySerializer<T> = (body: OperationRequestBodyContent<T>) => any;
 
 export type ParseAs = "json" | "text" | "blob" | "arrayBuffer" | "stream";
+export type ParseAsResponse<T, K extends ParseAs> = K extends "text"
+  ? Awaited<ReturnType<Response["text"]>>
+  : K extends "blob"
```
drwpow, line 44: The object was much cleaner, however, I was not able to get that working.

This is messier, but it does have the advantage of _always_ falling back to the original type `T` whereas the cleaner object could return `unknown`

packages/openapi-fetch/src/index.d.ts
```diff
@@ -86,155 +106,107 @@ export default function createClient<Paths extends {}>(
...
     FetchResponse<
-      "get" extends infer T
-        ? T extends "get"
-          ? T extends keyof Paths[P]
-            ? Paths[P][T]
-            : unknown
-          : never
-        : never
+      Paths[P]["get"],
+      I[0]["parseAs"] extends ParseAs ? I[O]["parseAs"] : "json"
```
drwpow, line 118: This seems like there’s a way to avoid this, but I couldn’t think of a way. I needed this in my version for TS to fall back to `json` if `init?.parseAs` was omitted

packages/openapi-fetch/src/index.d.ts
```diff
@@ -86,155 +106,107 @@ export default function createClient<Paths extends {}>(
...
-    ...init: HasRequiredKeys<
-      FetchOptions<FilterKeys<Paths[P], "get">>
-    > extends never
-      ? [(FetchOptions<FilterKeys<Paths[P], "get">> | undefined)?]
-      : [FetchOptions<FilterKeys<Paths[P], "get">>]
+    ...init: I
   ): Promise<
     FetchResponse<
-      "get" extends infer T
-        ? T extends "get"
```
drwpow, line 99: This was autogenerated TypeScript defs, not something I think is necessary. I cleaned this up to be more readable, and at least in the tests I didn’t notice any regressions

PR #1445, opened 2023-11-18
Fix defaultNonNullable behavior

drwpow:
## Changes

`defaultNonNullable: true` can’t make parameters required even if a default value is given ([comment](https://github.com/drwpow/openapi-typescript/issues/1368#issuecomment-1794837901)).

## How to Review

- Tests should pass.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/src/transform/schema-object.ts
+++ b/packages/openapi-typescript/src/transform/schema-object.ts
@@ -447,7 +447,9 @@ function transformSchemaObjectCore(
         }
         let optional =
           schemaObject.required?.includes(k) ||
-          ("default" in v && options.ctx.defaultNonNullable)
+          ("default" in v &&
+            options.ctx.defaultNonNullable &&
+            !options.path?.includes("parameters")) // parameters can’t be required, even with defaults
             ? undefined
             : QUESTION_TOKEN;
         let type =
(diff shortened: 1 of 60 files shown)


packages/openapi-typescript/test/transform/operation-object.test.ts
```diff
@@ -155,6 +155,65 @@ responses: {
             "application/json": string;
         };
     };
+};`,
+      },
+    ],
+    [
+      "defaultNonNullable > parameters aren’t required even with defaults",
```
drwpow, line 162: This new test asserts that a `parameter` with a default is ignored, but a response object with a default is not

packages/openapi-typescript/src/transform/schema-object.ts
```diff
@@ -447,7 +447,9 @@ function transformSchemaObjectCore(
         }
         let optional =
           schemaObject.required?.includes(k) ||
-          ("default" in v && options.ctx.defaultNonNullable)
+          ("default" in v &&
+            options.ctx.defaultNonNullable &&
+            !options.path?.includes("parameters")) // parameters can’t be required, even with defaults
```
drwpow, line 452: This seems like the dumbest solution to the problem, but I actually think it’s the correct one—parameters go through normal SchemaObject transformation, however, in this one scenario we apply different rules because of where they live (in any `parameters` object). I can’t think of a single scenario where this wouldn’t be true.

I also feel like the alternative—trying to override `ctx.defaultNonNullable` conditionally based on the parent—is a worse approach that could lead to other unintentional bugs.

PR #1460, opened 2023-11-25
support $ref on deduplication of operator-object and path-item parameters

hrsh7th:
## Changes

openapi-typescript already support parameters deduplication but it doesn't care about `$ref`.
so if schema has `$ref` parameter with same key, it will be invalid typescript types.

## How to Review

I used `ctx.resolve` method but I don't know it's reasonable.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/src/transform/path-item-object.ts
+++ b/packages/openapi-typescript/src/transform/path-item-object.ts
@@ -80,17 +80,17 @@ export default function transformPathItemObject(
     }
 
     // fold top-level PathItem parameters into method-level, with the latter overriding the former
-    const keyedParameters: Record<string, ParameterObject | ReferenceObject> =
-      {};
+    const keyedParameters: Record<string, ParameterObject | ReferenceObject> = {};
     if (!("$ref" in operationObject)) {
       // important: OperationObject parameters come last, and will override any conflicts with PathItem parameters
       for (const parameter of [
         ...(pathItem.parameters ?? []),
         ...(operationObject.parameters ?? []),
       ]) {
-        // note: the actual key doesn’t matter here, as long as it can match between PathItem and OperationObject
-        keyedParameters["$ref" in parameter ? parameter.$ref : parameter.name] =
-          parameter;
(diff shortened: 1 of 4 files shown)


drwpow:
Great fix, and great test! One suggestion: could you make this its own test? It’s added to another test that’s looking for something different–that `$ref`’d parameters aren’t lost in the object.

You could make another test that specifically tests for parameter key collision, either in the same test file (if you wanted to test the whole schema). Or a simplified test inside `transform/operation-object.test.ts` (with mocked `ctx.resolve()`) would be great, too. Having tests focus on a specific behavior ensures that as tests get touched, there’s not secret test behavior lost.

drwpow:
> I used `ctx.resolve` method but I don't know it's reasonable.

That’s perfect! That’s exactly why that exists—to do cross-schema lookup.

drwpow:
> openapi-typescript already support parameters deduplication but it doesn't care about `$ref`.
so if schema has `$ref` parameter with same key, it will be invalid typescript types.

This is correct. Though technically this is an error with the schema itself (Redocly’s linter allows you to throw an error on duplicate param key), I agree we should just solve for it anyway because it’s trivial to do so.

Great PR!

hrsh7th:
IMO, the test seems to aim to check `parameter override` because https://github.com/drwpow/openapi-typescript/pull/1460/files#diff-825cc17817d6be362cee3768df588ac6fa8ee329b4b8990dcc85767d8ad3cf3bR155 has comment for it.

I'll try to change the test location.

hrsh7th:
Sorry. I think `operation-object.test.ts` does not allows the test that needed the path level transformation because it uses `transformOperationObject` in its test.
What do you think?

drwpow:
> IMO, the test seems to aim to check `parameter override` because https://github.com/drwpow/openapi-typescript/pull/1460/files#diff-825cc17817d6be362cee3768df588ac6fa8ee329b4b8990dcc85767d8ad3cf3bR155 has comment for it.

You know you do make a good argument! In that case, maybe just also add `description`s to the test you wrote and that will be good enough so it’s clear what behavior it’s testing.

drwpow (approved):
🙏 Thank you!

PR #1468, opened 2023-11-30
fix: correctly transform untyped nullable field

### LINKED ISSUE #1407: Empty Openapi Type being misinterpreted as Record<string, unknown> when nullable
I found that when a field has no type information, but `nullable: true`, the resulting Typescript is `Record<string, unknown> | null`. 

I was originally thinking I may need to justify _why_ this is wrong but after some testing I found that this case is actually handled correctly for `nullable: false` - the output is simply `unknown`. I would suggest that the nullable version of this should be `unknown | null`.

** Versions **
Using `openapi-typescript@6.5.0`

**Reproduction**

I tried to simplify the Yaml as much as possible for repro:
```
openapi: 3.0.0
info:
  title: example
  version: 0.1.0

paths:
  /my_route:
    get:
      responses:
        "200":
          description: "Description"
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Test"

components:
  schemas:
    Test:
      type: object
      properties:
        nullable_unknown_field:
          nullable: true

```

**Expected result**

As I said above, the result for this ends up as 
```
nullable_unknown_field?: Record<string, unknown> | null;
```
However, it seems like this should actually be:
```
nullable_unknown_field?: unknown | null;
```

**Checklist**

- [x] My OpenAPI schema passes the [Redocly validator](https://redocly.com/docs/cli/commands/lint/) (`npx @redocly/cli@latest lint`) - yes
- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md)) - potentially 😅

davidleger95:
## Changes

Fixes #1407

Correctly transform untyped nullable field to `unknown` instead of `Record<string, unknown> | null`

## How to Review

First PR, so are there any test cases I'm missing? Anything I should add? Anything I should keep in mind going forward?

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/examples/digital-ocean-api.ts
+++ b/packages/openapi-typescript/examples/digital-ocean-api.ts
@@ -9085,7 +9085,7 @@ export interface external {
      * @description The Droplet that the floating IP has been assigned to. When you query a floating IP, if it is assigned to a Droplet, the entire Droplet object will be returned. If it is not assigned, the value will be null.
      * @example null
      */
-    droplet?: (Record<string, unknown> | null) | external["resources/droplets/models/droplet.yml"];
+    droplet?: unknown;
     /**
      * @description A boolean value indicating whether or not the floating IP has pending actions preventing new ones from being submitted.
      * @example true
@@ -13948,7 +13948,7 @@ export interface external {
      * @description The Droplet that the reserved IP has been assigned to. When you query a reserved IP, if it is assigned to a Droplet, the entire Droplet object will be returned. If it is not assigned, the value will be null.
      * @example null
      */
-    droplet?: (Record<string, unknown> | null) | external["resources/droplets/models/droplet.yml"];
+    droplet?: unknown;
     /**
      * @description A boolean value indicating whether or not the reserved IP has pending actions preventing new ones from being submitted.
      * @example true
--- a/packages/openapi-typescript/test/schema-object.test.ts
+++ b/packages/openapi-typescript/test/schema-object.test.ts
@@ -374,6 +374,11 @@ describe("Schema Object", () => {
       const generated = transformSchemaObject({}, options);
       expect(generated).toBe(`unknown`);
     });
+
+    test("unknown nullable", () => {
+      const generated = transformSchemaObject({ nullable: true }, options);
+      expect(generated).toBe(`unknown`);
+    });
   });
 
   describe("schema composition", () => {
--- a/.changeset/large-ways-cheat.md
+++ b/.changeset/large-ways-cheat.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Correctly transform untyped nullable fields to `unknown`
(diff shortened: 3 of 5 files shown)


packages/openapi-typescript/test/schema-object.test.ts
```diff
@@ -374,6 +374,11 @@ describe("Schema Object", () => {
       const generated = transformSchemaObject({}, options);
       expect(generated).toBe(`unknown`);
     });
+
+    test("unknown nullable", () => {
+      const generated = transformSchemaObject({ nullable: true }, options);
+      expect(generated).toBe(`unknown`);
+    });
```
davidleger95, line 381: The issue suggested the type should be `unknown | null`, but `unknown` is effectively the same and `tsUnionOf()` is won't do a union if the type is already `unknown`.

drwpow:
Ah tests failed for the gigantic list of snapshots. You’ll have to run the [update snapshots script](https://github.com/drwpow/openapi-typescript/blob/6.x/packages/openapi-typescript/CONTRIBUTING.md#updating-snapshot-examples) and commit the changes (which is also a helpful sanity check)

davidleger95:
> Ah tests failed for the gigantic list of snapshots. You’ll have to run the [update snapshots script](https://github.com/drwpow/openapi-typescript/blob/6.x/packages/openapi-typescript/CONTRIBUTING.md#updating-snapshot-examples) and commit the changes (which is also a helpful sanity check)

Ahh, I wasn't watching and rebuilding when I made the changes, so the snapshot script was referencing an old build without the changes.

CI catches it but wondering if it's worth running a build as part of the `update:examples` script to ensure it's always running with the most recent changes. The tradeoff is it's redundant if your build is already up to date. It's fast though so 🤷

I can add it in a new PR if that tradeoff makes sense.

drwpow:
> CI catches it but wondering if it's worth running a build as part of the update:examples script to ensure it's always running with the most recent changes. The tradeoff is it's redundant if your build is already up to date. It's fast though so 🤷

Oh yup that’s a great idea! And yeah build only takes a sec so that could happen.

Also the update script is a little janky/race condition-y so feel free to improve that however you feel like. That was just a quick slap-together.

packages/openapi-typescript/examples/digital-ocean-api.ts
```diff
@@ -9085,7 +9085,7 @@ export interface external {
      * @description The Droplet that the floating IP has been assigned to. When you query a floating IP, if it is assigned to a Droplet, the entire Droplet object will be returned. If it is not assigned, the value will be null.
      * @example null
      */
-    droplet?: (Record<string, unknown> | null) | external["resources/droplets/models/droplet.yml"];
```
drwpow, line 9088: Nice! Much better

drwpow (approved):
Looks great, thank you!

Since the `main` branch is currently the v7 beta, I’ll release this manually on 6.x. I’ll do that until v7 is stable, then changesets creates release PRs that just need merging.

WickyNilliams:
@davidleger95 just in case you missed the above comments. I'd be interested to hear your thoughts

Issue #1478, opened 2023-12-04
Error when compiling for CJS: "The current file is a CommonJS module..."


darwish:
**Description**

When including `openapi-fetch` from a CommonJS module, we get the following error:

```
node_modules/openapi-fetch/dist/cjs/index.d.cts:10:8 - error TS1479: The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("openapi-typescript-helpers")' call instead.

10 } from "openapi-typescript-helpers";
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in node_modules/openapi-fetch/dist/cjs/index.d.cts:10
```

**Reproduction**

See the following proof-of-concept repo: https://github.com/darwish/openapi-typescript-poc

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

PR #1479, opened 2023-12-04
Separate outputs for CommonJS and ESM when building openapi-typescrip…

darwish:
…t-helpers

## Changes

Fixes https://github.com/drwpow/openapi-typescript/issues/1478

Creates separate output directories for CommonJS and ESM for the output of `openapi-typescript-helpers`.

## How to Review

1. Run `pnpm build`
2. Check that both `packages/openapi-typescript-helpers/dist/` and `packages/openapi-typescript-helpers/dist/cjs/` contain an `index.ts` and an `index.d.ts` file.

## Checklist

- [ ] Unit tests updated

### DIFF
--- a/.changeset/hot-terms-vanish.md
+++ b/.changeset/hot-terms-vanish.md
@@ -1,0 +1,6 @@
+---
+"openapi-fetch": patch
+"openapi-typescript-helpers": patch
+---
+
+Fixed build of openapi-typescript-helpers for CommonJS environments
--- a/package.json
+++ b/package.json
@@ -10,8 +10,7 @@
   "scripts": {
     "build": "run-p -s build:*",
     "build:openapi-typescript": "cd packages/openapi-typescript && pnpm run build",
-    "build:openapi-typescript-helpers": "cd packages/openapi-typescript-helpers && pnpm run build",
-    "build:openapi-fetch": "cd packages/openapi-fetch && pnpm run build",
+    "build:openapi-fetch": "cd packages/openapi-typescript-helpers && pnpm run build && cd ../openapi-fetch && pnpm run build",
     "lint": "run-p -s lint:*",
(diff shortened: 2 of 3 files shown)


drwpow (approved):
Nice fix! Thank you!

drwpow:
Hm the failing tests might be from a race condition while building `openapi-fetch` and `openapi-typescript-helpers`. It seems that because `openapi-typescript-helpers` now requires a build step to work that it didn’t need before, it has to happen before `openapi-fetch`.

As a workaround, in the root, I’d recommend doing something like:

```diff
  "scripts": {
    "build": "run-p -s build:*",
    "build:openapi-typescript": "cd packages/openapi-typescript && pnpm run build",
-   "build:openapi-typescript-helpers": "cd packages/openapi-typescript-helpers && pnpm run build",
-   "build:openapi-fetch": "cd packages/openapi-fetch && pnpm run build",
+   "build:openapi-fetch": "cd packages/openapi-typescript-helpers && pnpm run build && cd ../openapi-fetch && pnpm run build"
```

The TS helpers build is basically instant/trivial. And this keeps openapi-fetch and openapi-typescript built in parallel, which while quick, still makes a noticeable difference in CI runs

drwpow (approved):
This is working for me locally, and seems to just be a weird CI issue. I’ll merge this, then poke at it separately.

drwpow:
Oh I figured it out—it was a CI thing. It was just in the `test` runs, each individual package was building itself but not others (i.e. we updated it in the wrong place, the building of `openapi-typescript-helpers` needed to happen inside `openapi-fetch`’s package.json itself).

But really it’s just a flimsy setup; I’ll make some changes that make that better (also we shouldn’t have a `prepare` step on `openapi-fetch` itself)

Issue #1122, opened 2023-05-22
 Some sort of middleware support


drwpow:
Original message by [@djMax](https://github.com/djMax) on https://github.com/drwpow/openapi-fetch/issues/41

---

**Description**

The request interception, response interception, etc is quite useful and I wouldn't think would add much to the library. Is there a way to simulate this on top of the library? (I don't see an obvious one)

**Proposal**

Add a single pre-flight and post-flight hook argument to the methods.

**Checklist**

- [X] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-fetch/blob/main/CONTRIBUTING.md))

djMax:
I started messing with this, and I now think the answer is "don't do it here." ClientOptions already supports a fetch argument, and you can basically do all you want with that, I think.

adarshaacharya:
@djMax so did you found any solution for this using fetch argument?

djMax:
for collecting basic metrics? Yes, you just wrap the fetch function with an observer. You can use something like this:

https://www.npmjs.com/package/fetch-hooks

I've been meaning to make a more purpose built module for this particular library but haven't had time. The one thing I think should change in this lib is the direct call of response.json(). Instead there should probably be some way to set what method is called to handle the response.

adarshaacharya:
@djMax i'm trying to implement refresh token using `openapi-typescript` and running out of idea for implementing that.
Used to do it using axios interceptor checking the status code, how can we implement here ?

djMax:
You would just make a fetch something like this:

```
async function customFetch(...args) {
  // Modify the request as necessary, e.g. with a Bearer token auth header
 const result = await fetch(...args);
 if (result.status === 403) {
   // refresh the token, call customFetch again, do something to protect against repeated calls
   return customFetch(...args);
 }
  return result;
}
```

WickyNilliams:
it's slightly off topic, but this i ended up on this issue when investigating how to throw on error instead of returning a data structure like `{ data, error }`. i am using tanstack query, so throwing on error is a must. i didn't want to manually do this every time.

i struggled to write an object that wraps each `GET`/`POST`/etc function without breaking the types. or having to restate all the generics, which is error prone.

i have landed on exposing the client like this:

```ts
const client = createApiClient<paths>({ ... })
export type Client = typeof client
export type ApiCall =  (client: Client) => Promise<FetchResponse<T>>

export function useApi() {
[...]

Issue #1502, opened 2024-01-14
Integration with tanstack-query


adiletto64:
**Description**
Using with react-query or vue-query [(this library)](https://tanstack.com/query/latest) you need manually return data or raise error:
```js
import createClient from "openapi-fetch";
import type { paths } from "./v1";

const client = createClient<paths>({ baseUrl: "https//localhost:8000" });

const {data, isLoading, isError} = useQuery({
    queryFn: async () => {
        const { data, error } = await client.GET("/articles/", { params });
        if (data) return data; // here
        throw new Error(error); // and here
    }
})
```
**Proposal**
Make wrapper or another createClient, so it elegantly works with tanstack-query
```js
const client = createQueryClient<paths>({ baseUrl: "https//localhost:8000" });

const {data, isLoading, isError} = useQuery({
    queryFn: client.GET("/articles/", { params }) // returns data or throws an error
})
```

**Checklist**

- [ ] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

drwpow:
I’m against the core library throwing an error by default, because it’s this library’s design to match the fetch API. Most applications should gracefully handle error codes without a try/catch, which is not only verbose; it can be slow in instances.

Requiring an error is a design decision of Tanstack Query that goes against the design of the fetch API. That said, I’m not opposed to a wrapper designed for Tanstack, but maybe there would be other functionality in addition to just simply throwing?

drwpow:
Update: with openapi-fetch adding middleware, there may be a more automatic way to achieve this: https://github.com/drwpow/openapi-typescript/pull/1521/files#diff-decdbf43769e15babbf0e979dcea2acadd305ca28f41e453db3329ad1f317b80R4

Issue #1486, opened 2023-12-13
"Unexpected end of JSON input" with bun.js


mihon73:
Hey there, thanks for your work.
Looks like needed one more fix for bun (with node works good).

**Description**
If fetch with the ban, it looks like an object that has not yet been fully received is cloned.
```js
86 |       }
87 |       const cloned = response.clone();
88 |       return {
89 |         data:
90 |           typeof cloned[parseAs] === "function"
91 |             ? await cloned[parseAs]()
                         ^
SyntaxError: Unexpected end of JSON input
      at /someLongPath/project/node_modules/openapi-fetch/dist/index.js:91:21
```

**Reproduction**
error occur with using bun only.
client configuration:
```js
const {GET} = createClient<paths>({
    baseUrl: 'https://some.awesome.api.endpoint/api/3.1',
    headers: {
        Authorization: "Bearer " + super_secret_token,
        "Content-type": "plain/text",
    },
})
```

The cloned response and the original are different, there is with some debug info before the error line:

---
in code:
```js
      // await Bun.sleep(3000)
      const cloned = response.clone()
      const clonedText = await cloned.text()
      const responseText = await response.text()
      console.log('cloned:', clonedText.length)
      console.log('response:', responseText.length)
      process.exit(0)

      return {
        data:
          typeof cloned[parseAs] === "function"
```
in output:
```
cloned: 0
response: 56084
```
---
And if we wait a little before cloning:
in code:
```js
      await Bun.sleep(3000)
      const cloned = response.clone()
      const clonedText = await cloned.text()
      const responseText = await response.text()
      console.log('cloned:', clonedText.length)
      console.log('response:', responseText.length)
      process.exit(0)

      return {
        data:
          typeof cloned[parseAs] === "function"
```
in output:
```
cloned: 56084
response: 56084
```

mihon73:
openapi-fetch 0.8.2
bun 1.0.16

titouv:
Getting the same error with a bearer authorized request if that can help

titouv:
After investigating it seems the problem is from the clone before getting the json from the body (at least in my case), removing the cloning (and probably breaking other functionalities) works for me.

titouv:
Seems like this issue is known in the Bun repo https://github.com/oven-sh/bun/issues/6969#issuecomment-1847705004
[...]

PR #1521, opened 2024-01-28
Middleware

drwpow:
## Changes

Addresses #1122 by adding Middleware support. This is an **<abbr title="request for comment">RFC</abbr>** so feedback is welcome! This will be used for testing & feedback before stable release.

You can download the beta version at the `@next` tag:

```sh
npm i openapi-fetch@next
```

### Dev Notes

[Please read the docs first](https://middleware.openapi-ts.pages.dev/openapi-fetch/middleware-auth) to learn about the API. The following are more “meta” notes helpful for early reviewers:

- The API was suggested by @denisw 🎉. I found it to be easy-to-use and clear in typing.
  - I pulled inspiration from tRPC links, Hono middleware, Next.js middleware, and Axios interceptors, but ultimately landed on a semi-unique API that’s a good blend of idiomatic and performant/lightweight.
  - A lot of middleware complexity stems from server streaming, which this library isn’t concerned with (as the 99% usecase is JSON, which is not streamable).
  - A simple middleware approach is not only easy-to-use; it’s also performant and lightweight (the final implementation only added a couple dozen lines to the codebase, and no performance impact).
- There’s still no `next()` callback.
  - In server middleware, this is necessary when streaming, or wanting to return a response to the user earlier but have some server work to do (logging, etc.).
  - But in `fetch()` land, and in my early testing, I found it to be a little overcomplicated since JSON responses (which this relies on) doesn’t utilize streaming. And to start I wanted to prefer a simpler middleware structure over a complex one (i.e. not encourage over-engineering clientside server setups)
- The `.use(middleware)` seemed to be a ubiquitous way to register middlewares
  - Originally, there was a `createClient({ middlewares: […] })` array, which doesn’t have any advantages, and only limits how/when middlewares are loaded.
  - The less-commonly-used `.eject()` was inspired by Axios interceptors. It may or may not get use, but was essentially free. So why not?

## How to Review

- 📚 [Read the docs](https://middleware.openapi-ts.pages.dev/openapi-fetch/middleware-auth)
- Test out the latest version at `openapi-fetch@next`.

## Checklist

- [x] Unit tests updated


packages/openapi-fetch/README.md
```diff
@@ -16,18 +16,18 @@ The syntax is inspired by popular libraries like react-query or Apollo client, b
...
 import type { paths } from "./api/v1"; // generated by openapi-typescript
 
-const { GET, PUT } = createClient<paths>({ baseUrl: "https://myapi.dev/v1/" });
+const client = createClient<paths>({ baseUrl: "https://myapi.dev/v1/" });
 
 const {
   data, // only present if 2XX response
   error, // only present if 4XX or 5XX response
-} = await GET("/blogposts/{post_id}", {
+} = await client.GET("/blogposts/{post_id}", {
```
drwpow, line 24: Extremely in-the-weeds: when writing the docs for middleware, I found `const { use } = createClient()` to be an awkward syntax (it it just me or does a floating `use()` just seem weird?). Renaming it (`const { use: useMiddleware }`) also felt awkward. `client.use(…)` felt like the most natural syntax, and matched other libraries.

As a result, it felt a little weird in other places to both keep `client` around and destructure it, so I opted to simplify and just keep `client` around in all places. Really, there’s no recommended syntax, and people can do what they want. I’m just trying to write the docs in the most “copy-pasteable” way :shrug:

packages/openapi-fetch/examples/react-query/src/lib/api/index.ts
```diff
@@ -1,5 +1,20 @@
-import createClient from "openapi-fetch";
+import createClient, { type Middleware } from "openapi-fetch";
 import type { paths } from "./v1";
 
+const throwOnError: Middleware = {
```
drwpow, line 4: People have [asked for this](https://github.com/drwpow/openapi-typescript/issues/1502): a more automatic way to interface with React Query. Hopefully this is sufficient!

docs/openapi-fetch/examples.md
```diff
@@ -4,130 +4,28 @@ title: openapi-fetch Examples
 
 # Examples
 
-## Authentication
```
drwpow, line 7: With middleware, these awkward, one-off workarounds can go away. Now it can be handled more-automatically.

packages/openapi-fetch/src/index.js
```diff
@@ -81,26 +116,17 @@ export default function createClient(clientOptions) {
     if (response.ok) {
       // if "stream", skip parsing entirely
       if (parseAs === "stream") {
-        // fix for bun: bun consumes response.body, therefore clone before accessing
-        // TODO: test this?
-        return { data: response.clone().body, response };
```
drwpow, line 86: I removed a lot of the `.clone()`-ing that was happening internally. I profiled this, and this can be a pretty big bottleneck, in addition to a waste of memory for large responses.

Instead, I’d rather users learn about safe vs unsafe `body` access rather than shield them from this. `openapi-fetch` will consume the body by calling `.json()` automatically (because 99% of the time that’s what users want). And if users need to read the raw body, then they’re doing something advanced and should know to `.clone()`.

packages/openapi-fetch/examples/react-query/src/lib/api/index.ts
WickyNilliams: I feel this may be generally useful enough to deserve a call out in the middleware docs (even if just cross linking to the example). It may get missed otherwise

docs/openapi-fetch/api.md
```diff
@@ -83,3 +83,100 @@ const { data, error } = await PUT("/submit", {
...
+  // …
+}
+```
+
+`onRequest()` takes 2 params:
+
+| Name      |        Type         | Description                                                                                                                                                                          |
+| :-------- | :-----------------: | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `req`     | `MiddlewareRequest` | A standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) with `schemaPath` (OpenAPI pathname) and `params` ([params](/openapi-fetch/api#fetch-options) object) |
+| `options` |   `MergedOptiosn`   | Combination of [createClient](/openapi-fetch/api#create-client) options + [fetch overrides](/openapi-fetch/api#fetch-options)                                                        |
```
djMax, line 127: Typo

docs/openapi-fetch/api.md
```diff
@@ -83,3 +83,100 @@ const { data, error } = await PUT("/submit", {
...
+onResponse(res, options) {
+  // …
+}
+```
+
+`onResponse()` also takes 2 params:
+| Name | Type | Description |
+| :-------- | :-----------------: | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `req` | `MiddlewareRequest` | A standard [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response). |
+| `options` | `MergedOptiosn` | Combination of [createClient](/openapi-fetch/api#create-client) options + [fetch overrides](/openapi-fetch/api#fetch-options) |
```
djMax, line 146: Typo

docs/openapi-fetch/middleware-auth.md
```diff
@@ -0,0 +1,192 @@
...
+onResponse(res, options) {
+  // …
+}
+```
+
+`onResponse()` also takes 2 params:
+| Name | Type | Description |
+| :-------- | :-----------------: | :----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `req` | `MiddlewareRequest` | A standard [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response). |
+| `options` | `MergedOptiosn` | Combination of [createClient](/openapi-fetch/api#create-client) options + [fetch overrides](/openapi-fetch/api#fetch-options) |
```
djMax, line 74: Typo. Now I'm thinking these must be generated? In which case should they really be checked in?

packages/openapi-fetch/src/index.js
drwpow: This related Bun issue here (#1486) I think is a mixture of a potential bug in Bun (which may be patched by now) along with user error. In either case, will investigate separately rather than cause potential performance bottlenecks for everyone.

Noahkoole:
Found another issue with type safety. https://github.com/drwpow/openapi-typescript/issues/1525

docs/openapi-fetch/index.md
```diff
@@ -4,34 +4,34 @@ title: openapi-fetch
 
 <img src="/assets/openapi-fetch.svg" alt="openapi-fetch" width="216" height="40" />
 
-openapi-fetch is a typesafe fetch client that pulls in your OpenAPI schema. Weighs **4 kb** and has virtually zero runtime. Works with React, Vue, Svelte, or vanilla JS.
+openapi-fetch is a typesafe fetch client that pulls in your OpenAPI schema. Weighs **5 kb** and has virtually zero runtime. Works with React, Vue, Svelte, or vanilla JS.
 
 | Library                    | Size (min) | “GET” request              |
 | :------------------------- | ---------: | :------------------------- |
-| openapi-fetch              |     `4 kB` | `278k` ops/s (fastest)     |
+| openapi-fetch              |     `5 kB` | `278k` ops/s (fastest)     |
```
drwpow, line 11: 🤷 sometimes the bytes do things

docs/openapi-fetch/testing.md
```diff
@@ -0,0 +1,67 @@
...
+title: Testing
+---
+
+# Testing
+
+Testing openapi-fetch is best done in a test runner that supports TypeScript, such as [Vitest](https://vitest.dev/) or Jest.
+
+## Mocking Requests
+
+To test requests, the `fetch` option can be supplied with any spy function like `vi.fn()` (Vitest) or `jest.fn()` (Jest).
```
drwpow, line 11: 🤔 I’ll probably rewrite this to recommend [MSW](https://mswjs.io/) as I’ve had some good success with it, but will probably wait until I refactor this codebase to use it for testing. Treat this as more of a placeholder.

packages/openapi-fetch/test/index.test.ts
```diff
@@ -187,13 +190,14 @@ describe("client", () => {
...
-          const post_id = "post?id = 🥴";
           await client.GET("/blogposts/{post_id}", {
-            params: { path: { post_id } },
+            params: { path: { post_id: "post?id = 🥴" } },
           });
 
           // expect post_id to be encoded properly
-          expect(fetchMocker.mock.calls[0][0]).toBe(`/blogposts/${post_id}`);
+          expect(fetchMocker.mock.calls[0][0].url).toBe(
+            `/blogposts/post?id%20=%20🥴`,
```
drwpow, line 199: Side-effect of using `new Request()`—we’re not encoding anything manually. Seems OK? Still tests UTF-8 sticks

PR #1524, opened 2024-01-30
Fix: keep type information for nullable fields when possible

WickyNilliams:
## Changes

The PR https://github.com/drwpow/openapi-typescript/pull/1468 was too aggressive in its use of `unknown` for nullable fields. It makes sense to use `unknown` when there is no `type` defined in the schema. But in cases where there is type information available, we should not discard it. 

The changes in the snapshots show that richer types are preserved when possible. I wrote a [comment](https://github.com/drwpow/openapi-typescript/pull/1468#issuecomment-1838526799) on the original PR relating to this, but it's possible it got missed. 

Essentially, we use the "old" behavior where we have some type info available in the schema, and the "new" behavior otherwise. The original bug report https://github.com/drwpow/openapi-typescript/issues/1407 says (emphasis mine):

> I found that when a field **has no type information**, but `nullable: true`, the resulting Typescript is Record<string, unknown> | null.

This PR ensures the "has no type information" part is respected.

## How to Review

The snapshots and tests should give enough context to evaluate the PR, and hopefully my justification makes sense.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/loud-items-hope.md
+++ b/.changeset/loud-items-hope.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Keep type information for nullable fields when possible
--- a/packages/openapi-typescript/examples/digital-ocean-api.ts
+++ b/packages/openapi-typescript/examples/digital-ocean-api.ts
@@ -9085,7 +9085,7 @@ export interface external {
      * @description The Droplet that the floating IP has been assigned to. When you query a floating IP, if it is assigned to a Droplet, the entire Droplet object will be returned. If it is not assigned, the value will be null.
      * @example null
      */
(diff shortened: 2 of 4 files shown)


WickyNilliams:
@davidleger95 i'd be interested in your thoughts, since this is a refinement of your PR

drwpow:
I’m OK with this change! Would welcome @davidleger95’s thoughts async on this as well, but I also think this is an improvement (I wouldn’t consider changes to `unknown` a significant breaking change)

drwpow (approved):

WickyNilliams:
Appreciate it! Can probably close #1407 as fixed now?

PR #1530, opened 2024-02-07
refactor(openapi-fetch): simplify createClient types

wydengyre:
## Changes

Adds a ClientMethod utility type to simplify the object definition of the return type of createClient.

## How to Review

This is a refactor adding a utility type and cleaning up some code. All tests should pass and no dependencies should be affected. Have a look at it and see if it seems more readable.

## Checklist

None of these apply to this change. It should not affect tests, not affect docs, and not touch the examples.

- [x] Unit tests updated

### DIFF
--- a/.changeset/sixty-oranges-arrive.md
+++ b/.changeset/sixty-oranges-arrive.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Exports the ClientMethod utility type.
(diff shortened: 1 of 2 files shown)


drwpow:
Thanks! I’m in favor of cleanup like this—looks much better. However, something is off, because the tests are failing (the `// @ts-expect-error`s are actually a critical part of the test—they make sure errors show when they should). But I don’t see why your approach wouldn’t work with some adjustments to fix the tests

wydengyre:
This is very strange. When I run `pnpm test` on the `main` branch on my dev computer, I get these exact same errors.

wydengyre:
@drwpow tests are passing when I run locally now

drwpow (approved):
Very nice!

PR #1534, opened 2024-02-15
Improve query + path serialization

### LINKED ISSUE #1533: Query parameter serialization for objects is not the openapi default
**Description**

When serializing query parameters for objects using the library, the serialization method employed does not match the OpenAPI 3.0 specification's default behavior. Specifically, the library uses a "deepObject" serialization style instead of the expected "form" serialization with explode=true, leading to an incorrect query parameter format.
[Openapi spec](https://swagger.io/docs/specification/serialization/#query)

**Reproduction**
Calling the:
```
const {
  data, // only present if 2XX response
  error, // only present if 4XX or 5XX response
} = await GET("/user", {
  params: {
    path: {id: {role: "admin", firstName: "Alex"}},
  },
});
```

the url parameters are: "id[role]=admin&id[firstName]=Alex"

**Expected result**

the url parameters should be: "role=admin&firstName=Alex"

**Checklist**

- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

drwpow:
## Changes

Resolves #1533

Overhauls serialization to support advanced [path serialization](https://swagger.io/docs/specification/serialization/#path), and allows for an alternate, simpler `querySerializer` syntax [according to the spec](https://swagger.io/docs/specification/serialization):

```ts
createClient({
  querySerializer: {
    array: { style: "form", explode: true },
    object: { style: "deepObject", explode: true },
    allowReserved: false,
  },
})
```

This change is **backwards-compatible**, meaning you can still provide a function to `querySerializer()`. However, it does introduce a ⚠️ **minor breaking change** in that deeply-nested objects/arrays aren’t supported (which hopefully no one was relying on, as that behavior is specifically outside the spec therefore there’s no predictable way to handle that).

## How to Review

- Tests should pass

## Checklist

- [x] Unit tests updated


packages/openapi-fetch/src/index.js
```diff
@@ -3,6 +3,8 @@ const DEFAULT_HEADERS = {
   "Content-Type": "application/json",
 };
 
+const PATH_PARAM_RE = /\{[^{}]+\}/g;
```
drwpow, line 6: Note (mostly for myself): I tested this simple RegEx approach, and a more low-level “parse-y” approach, and V8 seemed to favor this RegEx + `.match()` for performance. Before, this library used a static `.replace()`, and I wasn’t able to find a more performant alternative easily. There shouldn’t be any runtime impact (even splitting hairs).

Though I do value readability in code in general, when developing a library I think it is important to sweat the micro-optimizations, and not introduce runtime slowness just for the sake of readable code. I think that’s a luxury only afforded at the application level, not the library level.

packages/openapi-fetch/src/index.js
```diff
@@ -144,77 +162,224 @@ export default function createClient(clientOptions) {
...
+      default: {
+        return final;
+      }
+    }
   }
-  if (typeof value === "number" || typeof value === "boolean") {
-    return `${deepObjectPath(key)}=${String(value)}`;
+
+  // explode: true
+  for (const k in value) {
```
drwpow, line 224: In the vein of performance, both `Object.entries()` and `for … in` are (really) fast. But `for … in` is about 2× faster. And though this is totally splitting hairs, I’d be remiss not to use whatever is an order of magnitude faster if there’s no discernable readability/complexity impact.

This was tested in current V8, and this smells like a temporary thing that will flatten out so both will probably be equally performant over time. But I saw the numbers today, and had to do it 🤷

PR #1535, opened 2024-02-15
Support single operation parameter

tvanier:
## Changes

First, thanks for `openapi-typescript`!
I came across a generation failure when compiling the spec below (Azure OpenAI inference).
It appears some operations are defined with one single parameter object, not as array - see an example [here](https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/AzureOpenAI/inference/preview/2023-12-01-preview/examples/audio_transcription_object.json).

I am not 100% sure this is valid OpenAPI, but since `openapi-typescript@7` compiles it successfully, I thought it would be good to fix the problem in `6.x`.

```
openapi-typescript https://raw.githubusercontent.com/Azure/azure-rest-api-specs/main/specification/cognitiveservices/data-plane/AzureOpenAI/inference/preview/2023-12-01-preview/inference.json

file:///.../node_modules/openapi-typescript/dist/transform/operation-object.js:19
                for (const param of operationObject.parameters ?? []) {
                                                    ^

TypeError: object is not iterable (cannot read property Symbol(Symbol.iterator))
    at transformOperationObject (file:///.../node_modules/openapi-typescript/dist/transform/operation-object.js:19:53)
    at openapiTS (file:///.../node_modules/openapi-typescript/dist/index.js:137:39)
    at async generateSchema (file:///.../node_modules/openapi-typescript/bin/cli.js:94:18)
    at async main (file:///.../node_modules/openapi-typescript/bin/cli.js:170:5)
```

## How to Review

All iteration over operation or path `parameters` should call `getParametersArray` to make sure parameters are iterable.

## Checklist

- [X] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/test/operation-object.test.ts
+++ b/packages/openapi-typescript/test/operation-object.test.ts
@@ -164,4 +164,40 @@ describe("Operation Object", () => {
   };
 }`);
   });
+
+  it("handles non-array parameters", () => {
+    const schema: OperationObject = {
+      parameters: {
+        in: "query",
+        name: "search",
+        schema: { type: "string" },
+      },
+      responses: {
+        "200": {
+          description: "OK",
+          content: {
+            "application/json": {
+              schema: { type: "string" },
+            },
+          },
+        },
+      },
+    };
+    const generated = transformOperationObject(schema, options);
+    expect(generated).toBe(`{
+  parameters: {
+    query?: {
+      search?: string;
+    };
+  };
+  responses: {
+    /** @description OK */
+    200: {
+      content: {
+        "application/json": string;
+      };
+    };
+  };
+}`);
+  });
 });
--- a/.changeset/breezy-trees-brake.md
+++ b/.changeset/breezy-trees-brake.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Support single operation parameter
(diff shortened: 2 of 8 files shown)


drwpow:
Awesome, thank you! Happy to approve & merge this if you’d be so kind as to add a `patch` changeset (instructions in the 🦋 comment)

drwpow:
> I am not 100% sure this is valid OpenAPI

Many people use OpenAPI schemas in production that probably violate some aspect of the technical spec in some way or another. Since we’re converting to static TypeScript types, too, there can often be more opinionated interpretation than most realize. So the general approach of this library is to **Support as much of the official spec as possible (only barring technical limitations of TypeScript).**

Beyond the spec, we’ll generally allow for fixes to common footguns/rough edges so long as it doesn’t deviate from the spec, and isn’t a huge maintenance burden to maintain. Your PR is a great example of this—all the existing tests pass; we just covered an additional usecase without any breaking changes. That’s always a win in my book.

tvanier:
Thanks @drwpow , fully agreed on the flexible approach. I think I added a `patch` changeset, let me know if anything is missing or incorrect.
For info I use openapi-typescript + fetch to connect to different AI providers (OpenAI, Azure, Mistral, Ollama etc), works great!

packages/openapi-typescript/test/operation-object.test.ts
```diff
@@ -162,6 +162,42 @@ describe("Operation Object", () => {
       };
     };
   };
+}`);
+  });
+
+  it("handles non-array parameters", () => {
```
drwpow, line 168: Great tests

drwpow (approved):

drwpow:
Thanks again! Changeset is perfect 🙂

tvanier:
Hi @drwpow , may I ask if there is a plan to release these PR changes in version `6.7.5` when you get a chance? Thanks!

drwpow:
@tvanier yes thank you for the nudge! I needed the nudge. Normally these get deployed automatically but I had to pause it until `7.x` is stable (which is very close; just needs the last couple items done on the roadmap). Will release it later today.

drwpow:
Released!

tvanier:
Thanks @drwpow , release `6.7.5` works good! And I started to migrate to `v7` ;-)

PR #1545, opened 2024-02-16
fix: replace # characters in operation IDs with a slash

### LINKED ISSUE #1542: Operation types are being mapped incorrectly (v7)
**Description**

The operation type mapping no longer uses the correct `operationId` keys.

| `openapi-typescript` | `7.0.0-next.7`                      |
| Node.js              | `21.0.0`                      |
| OS + version         | `macOS 14.2.1` |

**Reproduction**

Given an OpenAPI v3 spec

```yml
openapi: 3.0.3
info:
    title: Web API
    version: "1.0"
servers:
    - url: https://api.hgraph.com
paths:
    /v1/accounts/{id}:
        get:
            tags:
                - Accounts
            summary: Get Accounts
            description: Get an account.
            operationId: Accounts#Get
            parameters:
                - name: id
                  in: path
                  description: The ID of the account
                  required: true
                  schema:
                    type: string
                    description: The ID of the account
                    example: d9f08fe2-cc42-11ee-a737-6eaad63d76fb
                    format: uuid
                  example: d9f09118-cc42-11ee-a737-6eaad63d76fb
            responses:
                "200":
                    description: OK response.
                    content:
                        application/json:
                            schema:
                                $ref: '#/components/schemas/Account'
                            example:
                                created_at: "1974-07-14T21:12:28Z"
                                id: d9da03c6-cc42-11ee-a737-6eaad63d76fb
                                personal_org_id: d9da051a-cc42-11ee-a737-6eaad63d76fb
                                primary_email_address_id: d9da0664-cc42-11ee-a737-6eaad63d76fb
```

The following types are generated:

```ts
export type paths = {
  "/v1/accounts/{id}": {
    parameters: {
      query?: never;
      header?: never;
      path?: never;
      cookie?: never;
    };
    /**
     * Get Accounts
     * @description Get an account.
     */
    get: operations["Get"]; // NOTE: here is the bug
    put?: never;
    post?: never;
    delete?: never;
    options?: never;
    head?: never;
    patch?: never;
    trace?: never;
  };
}

export type operations = {
  "Accounts#Get": {
    parameters: {
      query?: never;
      header?: never;
      path: {
        /**
         * @description The ID of the account
         * @example d9f09118-cc42-11ee-a737-6eaad63d76fb
         */
        id: string;
      };
      cookie?: never;
    };
    requestBody?: never;
    responses: {
      /** @description OK response. */
      200: {
        headers: {
          [name: string]: unknown;
        };
        content: {
          "application/json": components["schemas"]["Account"];
        };
     },
}
```

As you can see, the `get` method is mapped to `operations["Get"];` , however the operation key is actually `Accounts#Get`. This only started happening today as far as I can tell. It has generated correctly to date.

**Expected result**

```ts
export type paths = {
  "/v1/accounts/{id}": {
    parameters: {
      query?: never;
      header?: never;
      path?: never;
      cookie?: never;
    };
    /**
     * Get Accounts
     * @description Get an account.
     */
    get: operations["Accounts#Get"]; // NOTE: here is the bug
    /**
     * Update Accounts
     * @description Update an account.
     */
    put: operations["Accounts#Update"];
    post?: never;
    delete?: never;
    options?: never;
    head?: never;
    patch?: never;
    trace?: never;
  };
}

export type operations = {
  "Accounts#Get": {
    parameters: {
      query?: never;
      header?: never;
      path: {
        /**
         * @description The ID of the account
         * @example d9f09118-cc42-11ee-a737-6eaad63d76fb
         */
        id: string;
      };
      cookie?: never;
    };
    requestBody?: never;
    responses: {
      /** @description OK response. */
      200: {
        headers: {
          [name: string]: unknown;
        };
        content: {
          "application/json": components["schemas"]["Account"];
        };
     },
  },
}
```

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

jaredLunde:
## Changes

Fixes #1542

A [recent change](https://github.com/Redocly/redocly-cli/pull/1408) to redocly affected the way JSON pointers are parsed, causing operation IDs with the "#" character to generate the incorrect type mapping to operations referenced in the path item object. This works around the issue by replacing "#" characters in operation IDs with a "/" character.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/src/transform/path-item-object.ts
+++ b/packages/openapi-typescript/src/transform/path-item-object.ts
@@ -105,7 +105,7 @@ export default function transformPathItemObject(
     // if operationId exists, move into an `operations` export and pass the reference in here
     else if (operationObject.operationId) {
       // workaround for issue caused by redocly ref parsing: https://github.com/drwpow/openapi-typescript/issues/1542
-      const operationId = operationObject.operationId.replace(/#/g, "/");
+      const operationId = operationObject.operationId.replace(HASH_RE, "/");
       operationType = oapiRef(createRef(["operations", operationId]));
       injectOperationObject(
         operationId,
@@ -132,3 +132,5 @@ export default function transformPathItemObject(
 
   return ts.factory.createTypeLiteralNode(type);
 }
+
+const HASH_RE = /#/g;
--- a/packages/openapi-typescript/src/transform/path-item-object.ts
+++ b/packages/openapi-typescript/src/transform/path-item-object.ts
@@ -104,11 +104,11 @@ export default function transformPathItemObject(
     }
     // if operationId exists, move into an `operations` export and pass the reference in here
     else if (operationObject.operationId) {
-      operationType = oapiRef(
-        createRef(["operations", operationObject.operationId]),
-      );
+      // workaround for issue caused by redocly ref parsing: https://github.com/drwpow/openapi-typescript/issues/1542
+      const operationId = operationObject.operationId.replace(/#/g, "/");
+      operationType = oapiRef(createRef(["operations", operationId]));
       injectOperationObject(
-        operationObject.operationId,
+        operationId,
         { ...operationObject, parameters: Object.values(keyedParameters) },
         { ...options, path: createRef([options.path!, method]) },
       );
--- a/packages/openapi-typescript/test/index.test.ts
+++ b/packages/openapi-typescript/test/index.test.ts
@@ -488,6 +488,84 @@ export type operations = Record<string, never>;`,
         // options: DEFAULT_OPTIONS
       },
     ],
+
+    [
+      "operations > # character is parsed correctly",
+      {
+        given: {
+          openapi: "3.1",
+          info: { title: "Test", version: "1.0" },
+          paths: {
+            "/accounts": {
+              get: {
+                operationId: "Accounts#List",
+                responses: {
+                  200: {
+                    description: "OK",
+                    content: {
+                      "application/json": {
+                        schema: { type: "string" },
+                      },
+                    },
+                  },
+                },
+              },
+            },
+          },
+        },
+        want: `export interface paths {
+    "/accounts": {
+        parameters: {
+            query?: never;
+            header?: never;
+            path?: never;
+            cookie?: never;
+        };
+        get: operations["Accounts/List"];
+        put?: never;
+        post?: never;
+        delete?: never;
+        options?: never;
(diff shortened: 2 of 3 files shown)


packages/openapi-typescript/src/transform/path-item-object.ts
```diff
@@ -104,11 +104,11 @@ export default function transformPathItemObject(
     }
     // if operationId exists, move into an `operations` export and pass the reference in here
     else if (operationObject.operationId) {
-      operationType = oapiRef(
-        createRef(["operations", operationObject.operationId]),
-      );
+      // workaround for issue caused by redocly ref parsing: https://github.com/drwpow/openapi-typescript/issues/1542
+      const operationId = operationObject.operationId.replace(/#/g, "/");
```
drwpow, line 108: For RegExs, it’s good to pull them out into the top-level module scope as constants, that way they don’t recompile every function run. This one may be negligible, but some RegExs can perceptibly slow things down especially when looping; better to just do it regardless.

```suggestion
      const operationId = operationObject.operationId.replace(HASH_RE, "/");
```

packages/openapi-typescript/test/index.test.ts
```diff
@@ -488,6 +488,84 @@ export type operations = Record<string, never>;`,
         // options: DEFAULT_OPTIONS
       },
     ],
+
+    [
+      "operations > # character is parsed correctly",
```
drwpow, line 493: 👍 Great test!

drwpow (commented):
Looks great, and great test! Just one minor nitpick on RegExs; otherwise happy to approve & merge

packages/openapi-typescript/src/transform/path-item-object.ts
drwpow: Just confirming (to myself) I’m not crazy and this is still valid: still a perceptible difference in V8 between the two:

CleanShot 2024-02-15 at 20 48 34@2x

17% isn’t a lot, but this is the best-case scenario. Some larger RegExs can be orders of magnitude slower. Also, in application-level code this is splitting hairs, but for libraries like this I think splitting hairs like this can be helpful to the end user.

jaredLunde: For sure- good call

drwpow (approved):

Issue #1560, opened 2024-02-22
Document Chinese support


phk422:
Hi, @drwpow ! For the convenience of Chinese users, I want the document to support Chinese and I can participate in the construction and translation

drwpow:
We would love that! Vitepress makes i18n easy; I just haven’t set it up yet. I will make some placeholder documents and let you (and others) contribute as they’re able to.

What would be a good ISO URL path to set up placeholders? `/zh/…`, `/yue/…`, or something else?

drwpow:
Opened up #1561 for feedback! Please suggest any/all changes. Also see the [preview deployment](https://i18n.openapi-ts.pages.dev/). Ideally you could open up a few PRs against this branch, and we could merge in, just so a couple articles are filled in. Then happy to merge when there’s any content at all, and we can build it over time (doesn’t have to be complete 1st pass or anything)

phk422:
Hello @drwpow ,Before we get started, I realized when using the light theme, the Get start button on the home page is not visible.
![image](https://github.com/drwpow/openapi-typescript/assets/59734322/feeadaf8-22ab-4a92-9786-7c2f5929e76e)

drwpow:
The button color is set with [oklch](https://caniuse.com/?search=oklch); does your browser/OS not support that? And if not, what are you using? (I had thought it had wide support now but I guess I was mistaken)

drwpow:
Oh! It’s not `oklch()` that’s the problem; it’s that the color is in the [P3 gamut](https://oklch.com/#40,0.2,260,100) and needs a fallback for non-P3 monitors. Will fix! Thank you.

phk422:
> The button color is set with [oklch](https://caniuse.com/?search=oklch); does your browser/OS not support that? And if not, what are you using? (I had thought it had wide support now but I guess I was mistaken)

My Google browser hasn't been upgraded for a long time, but when I upgraded to the latest version, it worked fine. Maybe it's a browser version compatibility issue

drwpow:
It’s live! 🙌 Thanks so much for all your hard work! 🙂 There are still a few more pages to translate, but I wanted to at least get all the pages you have done up so that the work can happen as you have time to (and also for others to help contribute—maybe this will even inspire other languages). Thanks so much for making this happen ❤️

drwpow:
This has been completed! 🎉 While there will always be room for additional PRs and fixes, I think this ticket has served its purpose.

PR #1561, opened 2024-02-22
Chinese Translation

drwpow:
## Changes

Adds placeholders for #1560 Thanks to @phk422 

## How to Review

- Is the ISO code correct?
- Is `简体中文` (and the flag) correct in the config?
- Should untranslated pages simply link to the English version? Should they appear in the sidebar as `Node (EN)`? Or should they remain as stubs?
- Would you change anything else the structure at all?

## Checklist

N/A


phk422:
#1562

PR #1562, opened 2024-02-23
WIP: Chinese translation

phk422:
I have modified the organization of relevant configuration files, please confirm whether this is feasible? If so, I can start translating


docs/.vitepress/zh.ts
```diff
@@ -0,0 +1,60 @@
+import { defineConfig } from 'vitepress'
+
+export const zh = defineConfig({
```
drwpow, line 3: Ah this is smart! I wasn’t thinking about all the additional links and navigation items that needed to be translated.

docs/.vitepress/en.ts
```diff
@@ -0,0 +1,117 @@
...
+            },
+            { text: "Testing", link: "/openapi-fetch/testing" },
+            { text: "Examples", link: "/openapi-fetch/examples" },
+            { text: "API", link: "/openapi-fetch/api" },
+            { text: "About", link: "/openapi-fetch/about" },
+          ],
+        },
+      ],
+      // 🌐
+      "/zh/": [
```
drwpow, line 44: We can probably remove `zh` from the English sidebar, right?

drwpow (approved):
Looks great!

docs/.vitepress/en.ts
phk422: Yes! I've removed it.

drwpow:
This is looking great! If you’re OK with it, I’m going to merge this, and deploy it. We can add additional pages in followup PRs.

PR #1566, opened 2024-02-27
update openapi-typescript-fetch bench

ajaishankar:
## Changes

Update bench to use openapi-typescript correctly. 

Create and reuse the fetch method instead of instantiating it on every invocation.

This should speed up the benchmark slightly.

## How to Review

This change moves the method creation outside of the bench

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/test/index.bench.js
+++ b/packages/openapi-fetch/test/index.bench.js
@@ -41,13 +41,14 @@ describe("get (only URL)", () => {
   openapiTSFetch.configure({
     init: { baseUrl: BASE_URL },
   });
+  let openapiTSFetchGET = openapiTSFetch.path("/url").method("get").create();
 
   bench("openapi-fetch", async () => {
     await openapiFetch.GET("/url");
   });
 
   bench("openapi-typescript-fetch", async () => {
-    await openapiTSFetch.path("/url").method("get").create()();
+    await openapiTSFetchGET();
   });
 
   bench("openapi-typescript-codegen", async () => {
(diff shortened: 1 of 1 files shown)


ajaishankar:
@drwpow it's great that lots of progress has been made on a (very elegant) fetch implementation and it is now part of openapi-typescript! 👍 

The fetch was completely missing two years back when I first published openapi-typescript-fetch.

Just wanted to update the bench with correct usage...

drwpow (approved):

ajaishankar:
I was surprised why there would be so much of a difference in the benchmark.

Turns out my target was ES2015 and Typescript was transpiling async/await into some generators.

Once the target was set to ESNext that overhead went away and I'm getting same/better ops/s as openapi-fetch  (on my machine 😉)

bench

Again, not sure how relevant these benchmarks are.

API wise openapi-fetch is so much cleaner, thanks for this and also openapi-typescript! 👍

drwpow:
Yeah as I said the _absolute_ numbers are kinda garbage from my findings, and I need to replace it with something more reliable. But the _relative_ numbers (e.g. `1.38x faster than axios`) tend to be fairly consistent.

Also as someone that is always learning about benchmarks, you’re benchmarking in V8 (Node, Chrome), which rewards certain particular operations and punishes others. JavaScriptCore (Safari) and SpiderMonkey (Firefox) will have different benchmarks on the same code. But in general, things like RegExs, functional programming (`.reduce()` vs `for … of`), and classic [Big O Notation](https://en.wikipedia.org/wiki/Big_O_notation) will all affect the runtime. For a fetch library, IMHO it is very important to sweat every little drop of performance, because these are the core of so many applications, and get executed billions of times. In many applications this is splitting hairs, but for fetching, I think it matters.

ajaishankar:
Got it!

Time permitting, I'll try to add the following to the benchmark that exercises the following:

1. Path construction from request payload args
2. Query string serialization
3. Exception handling
4. Middleware handling (loop vs recursion in openapi-typescript-fetch)

In my case I'd to make a clone of the payload because of a design choice (all params are mashed into a single argument and deconstructed) whereas in openapi-fetch this overhead may not be there.

Will also publish a esnext version of the library for comparison.

Issue #1569, opened 2024-02-27
Custom fetch options are stripped from the request. Next.js cache doesn't work properly.


petercsaki:
**Description**

Openapi-fetch uses a request object for its fetch. https://github.com/drwpow/openapi-typescript/blob/9c277fb0a10c3513de46765a4381ccb722a72af4/packages/openapi-fetch/src/index.js#L101
The request object created by openapi-fetch strips away non standard fetch options, i.e. how Next.js passes its cache related params: 
```
{ next: { revalidate: 3600 } }
```

More info on [Next.js fetching data + caching](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating#fetching-data-on-the-server-with-fetch)

**Reproduction**

Try to pass next cache params in https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/examples/nextjs/app/page.tsx
It won't change its behaviour, it always falls back to defaults, because the non standard ones are stripped away.

**I've tried**

I changed this line 
https://github.com/drwpow/openapi-typescript/blob/9c277fb0a10c3513de46765a4381ccb722a72af4/packages/openapi-fetch/src/index.js#L101

to
```
let response = await fetch(request, requestInit);
```
and it works, but I don't know if it would cause other problems.

drwpow:
Ah, hm. I was somewhat afraid this would happen, and did see it come up in one other test failure, but had thought it was fixed.

Would gladly welcome any PRs for this! This should be a fairly-straightforward issue to write a failing test for, and find a fix.

petercsaki:
PR created.

JE-lee:
For the workaround, this could be ok as well as the cache mechanism, regardless a little wired.
```typescript
  const res = await client.GET("/api/v1/space", {
    fetch: (request: unknown) => {
      return fetch(request as Request, { next: { revalidate: 40 } } )
    }
  })
```
I have tested it in Next.js 14

FreeAoi:
Hi, a pull request fixing this was merged #1653 recently

sazzouz:
[...]

PR #1571, opened 2024-02-28
NextJs options added to fetch

petercsaki:
## Changes
NextJs uses fetch's options to pass information about caching. These were removed by openapi-fetch.

https://github.com/drwpow/openapi-typescript/issues/1569#issuecomment-1966999115

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/rare-pets-lie.md
+++ b/.changeset/rare-pets-lie.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Fixed fetch options used by NextJs to be passed to fetch.
--- a/packages/openapi-fetch/examples/nextjs/app/page.tsx
+++ b/packages/openapi-fetch/examples/nextjs/app/page.tsx
@@ -9,7 +9,7 @@ async function getFact() {
     next: {
       revalidate: 10,
       tags: ["cat"],
-    }
+    },
   });
 }
 
--- a/packages/openapi-fetch/examples/nextjs/app/page.tsx
+++ b/packages/openapi-fetch/examples/nextjs/app/page.tsx
@@ -6,6 +6,10 @@ async function getFact() {
     params: {
       query: { max_length: 500 },
     },
+    next: {
+      revalidate: 10,
+      tags: ["cat"],
+    }
   });
 }
 
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -108,10 +108,12 @@ export type RequestBodyOption<T> =
       : { body: OperationRequestBodyContent<T> };
 
 export type FetchOptions<T> = RequestOptions<T> &
-  Omit<RequestInit, "body" | "headers"> & NextJsFetchOptions;
+  Omit<RequestInit, "body" | "headers"> &
+  NextJsFetchOptions;
 
-export type NextJsFetchOptions = 
-  {next?: {revalidate?: false | 0 | number, tags?: string[]}};
+export type NextJsFetchOptions = {
+  next?: { revalidate?: false | 0 | number; tags?: string[] };
+};
 
 /** This type helper makes the 2nd function param required if params/requestBody are required; otherwise, optional */
 export type MaybeOptionalInit<P extends {}, M extends keyof P> =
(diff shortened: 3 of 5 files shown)


drwpow:
Thank you for opening! I’m +1 in favor of fixing this issue, and the test is perfect! But I would like to solve this **without having any Next.js-specific code in the runtime**. This probably means somehow sniffing out [RequestInit](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#options) and passing it along before/after each middleware.

If we can make that test pass in a generic way without looking for `.next`, then we’ll not only solve for Next.js; we’ll solve for all other libraries that are experiencing the same issue.

drwpow (changes requested):
+1 in favor of this merging, and this is a great fix! But as we discussed I’d **like to solve this without manually declaring `.next`.** Perhaps there’s a more universal way we can solve this (such as manually rebuilding `new Request()`s, or another creative solution)?

If we had to have framework-specific code in `openapi-fetch`, we’d always be playing catchup. There’d always be a new vendor prefix (or existing ones could change) and there’d always be something left out and another version to ship. But solving for **any arbitrary property added to the request** would let us solve this problem once and for all.

petercsaki:
Yep, I get it. I'll need to think about how to get the parameters that aren't used in creating the Request object and should be passed as the second parameter for fetch.

JE-lee:
After checking the document of the fetch in Next.js and MDN, maybe there is no need to add an external option to achieve that.
We can just add the Next.js options  like this 
```typescript
  const res = await client.GET("/api", { next: {revalidate: 10}})
```
I haven't tested it yet. Have you tried this @petercsaki

petercsaki:
Yes. Openapi-fetch uses them as parameters when creating the Request object:

https://github.com/drwpow/openapi-typescript/blob/9c277fb0a10c3513de46765a4381ccb722a72af4/packages/openapi-fetch/src/index.js#L72C5-L75C7

but the request object only uses the properties of it that it knows about and others are "forgotten". The code uses only the Request object from then on.

JE-lee:
> Yes. Openapi-fetch uses them as parameters when creating the Request object:
> 
> https://github.com/drwpow/openapi-typescript/blob/9c277fb0a10c3513de46765a4381ccb722a72af4/packages/openapi-fetch/src/index.js#L72C5-L75C7
> 
> but the request object only uses the properties of it that it knows about and others are "forgotten". The code uses only the Request object from then on. So any custom properties are ignored.

Haha, I got what you mean. I have tested some cases. It seems like The Next.js extended Fetch api only takes the second parameter as the cache option.
The following code works as well as the cache mechanism.

```typescript
  const request = new Request('http://localhost:3000/api/v1/space', {
    method: 'GET',
  })
  const res = await fetch(request, { next: { revalidate: 10 } }) 
```
so you are in the right way. keep going.

IngoVals:
Would it be possible to switch `NextJsFetchOptions` out for `AdditionalProps` which is defined as an empty object that we spread into the fetch. Then in each case users could ts declare AdditionalProps to what they want ( nextjs props for example ).

```ts
declare module 'openapi-fetch' {
    type AdditionalProps {
        next?: { revalidate?: false | 0 | number; tags?: string[] };
     }
}
```

Just brainstorming.

drwpow:
@IngoVals love that suggestion! That would be a great addition, and I see the value in typechecking additional properties like that. That would be much better than just adding `[key: string]: unknown` to the shape.

That could either be part of this PR at the author’s discretion, or as a followup.

ryami333:
Is this even necessary? Because NextJS uses "ambient modules" to mutate `RequestInit`, which `openapi-typescript` already extends for `FetchOptions`.

petercsaki:
> Is this even necessary? 

Openapi-fetch loses the next specific parameters before it gets to Next's fetch.

FreeAoi:
How's going this PR?

petercsaki:
I've abandoned this, sorry. Someone pointed out that you can pass a custom fetch to each request, and that can use next's cahcing. That solution works for us. Creating a generic solution is not in my scope. Feel free to pick this up.

drwpow:
No worries! I’m going to close this then, but someone else is free to pick this up, provided my restriction up above: the solution should allow for **any arbitrary properties,** not just `.next`

PR #1575, opened 2024-03-03
[version 5.x] chore: Bump the pnpm lockfile version to 6.0 and fix tests

mitchell-merry:
## Changes

The pnpm lockfile version is bumped to 6.0. I ran `pnpm install` and it automatically did this, so I figure probably a good idea to commit upstream...

I'm doing this because I want to make other changes (which will be in a subsequent PR)

Edit: Also had to fix the remote reference to test files since they no longer exist on main.

## How to Review

Dunno, tests pass?

## Checklist

Unit tests updated, rest were n/a!

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,56 +1,75 @@
-lockfileVersion: 5.4
-
-specifiers:
-  '@types/js-yaml': ^4.0.5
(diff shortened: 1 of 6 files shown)


drwpow (approved):
Awesome, thanks! Old versions I have to release manually, but happy to do so

PR #1576, opened 2024-03-03
feat: add `--x-nullable-as-nullable` to treat schemas with `x-nullable` as being able to specify `null`

mitchell-merry:
## Changes

Essentially adds the `nullable` feature from 3.0 into 2.0 as well, behind the `--x-nullable-as-nullable` flag.

## How to Review

Stacked on https://github.com/drwpow/openapi-typescript/pull/1575!

## Checklist

- [x] Unit tests updated
Specifically didn't add too much so as not to over-test. This is not a commonly used version, anywho


drwpow:
Thanks for adding this! I’m good with this addition. While I know 3.x encourages `type: [null]` as the recommended way, `nullable: true` and `x-nullable: true` are still really common patterns I see in 3.x schemas and I think supporting both doesn’t stand in the way of the 3.x specs while providing value to users. +1 to this landing.

mitchell-merry:
Specifically, this is to support a 2.x schema, where I believe there is no official pattern for nullable (outside of this `x-nullable` thing) (which is why I am PRing into 5.x 😄).

drwpow (approved):

mitchell-merry:
Thanks for the merge! Could we get a release of 5.x with this change?

drwpow:
Published!

PR #1592, opened 2024-03-17
replace vitest-fetch-mock with msw

### LINKED ISSUE #1581: Replace vitest-fetch-mock with msw in tests
**Description**

`vitest-fetch-mock` could be replaced with [msw](https://mswjs.io/) in the tests for better reliability (and better testing with msw in general). Would love if someone opened a PR to do this! 🙏

vipentti:
## Changes

- replaces vitest-fetch-mock with msw in openapi-fetch tests (including `index.bench.js`)

Fixes #1581

## How to Review

I've made an effort to keep the existing tests as close as possible to their original implementation, with the exception of replacing `mockFetch*` with the equivalent `msw` test handler. Due to `msw`, it's necessary to pass `baseUrl` around to ensure we're using the correct request URLs. Therefore, it has been provided for each `createClient` and equivalent test handler.

Please note, the current usage of `msw` contradicts `msw`'s best practices ([Avoid Request Assertions](https://mswjs.io/docs/best-practices/avoid-request-assertions)). However, given the nature of this library, I believe this approach is acceptable in this context. In most of these tests, we aim to validate that the client is indeed calling the expected handler with the expected request parameters.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/test/index.test.ts
+++ b/packages/openapi-fetch/test/index.test.ts
@@ -128,7 +128,7 @@ function useTestRequestHandler<
   );
 
   return {
-    getCookies: () => receivedCookies!,
+    getRequestCookies: () => receivedCookies!,
     getRequest: () => receivedRequest!,
     getRequestUrl: () => new URL(requestUrl),
   };
@@ -1259,7 +1259,7 @@ describe("client", () => {
     it("respects cookie", async () => {
       const client = createClient<paths>({ baseUrl });
 
-      const { getCookies } = useTestRequestHandler({
+      const { getRequestCookies } = useTestRequestHandler({
         baseUrl,
         method: "get",
         path: "/blogposts",
@@ -1272,7 +1272,7 @@ describe("client", () => {
         },
       });
 
-      const cookies = getCookies();
+      const cookies = getRequestCookies();
       expect(cookies).toEqual({ session: "1234" });
     });
   });
--- a/packages/openapi-fetch/test/index.test.ts
+++ b/packages/openapi-fetch/test/index.test.ts
@@ -28,7 +28,7 @@ const baseUrl = "https://api.example.com";
  * Test path helper, returns a an absolute URL based on
  * the given path and base
  */
-function testPath(path: string, base: string = baseUrl) {
+function toAbsoluteURL(path: string, base: string = baseUrl) {
   // If we have absolute path
   if (URL.canParse(path)) {
     return new URL(path).toString();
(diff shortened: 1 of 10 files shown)


packages/openapi-fetch/test/index.test.ts
```diff
@@ -1,33 +1,137 @@
...
+/**
+ * Default baseUrl for tests
+ */
+const baseUrl = "https://api.example.com";
+
+/**
+ * Test path helper, returns a an absolute URL based on
+ * the given path and base
+ */
+function testPath(path: string, base: string = baseUrl) {
```
vipentti, line 31: This is only an internal helper, but could probably be re-named to something more descriptive.

packages/openapi-fetch/test/index.test.ts
```diff
@@ -1,33 +1,137 @@
...
+        return HttpResponse.json(body, {
+          status: status ?? 200,
+          headers,
+        }) as AsyncResponseResolverReturnType<ResponseBodyType>;
+      },
+    ),
+  );
+
+  return {
+    getCookies: () => receivedCookies!,
```
vipentti, line 131: This should maybe be renamed to `getRequestCookies` for consistency.

packages/openapi-fetch/test/index.test.ts
```diff
@@ -1,33 +1,137 @@
...
-  status: number;
-  body: any;
+  status?: number;
+  handler?: HttpResponseResolver<Params, RequestBodyType, ResponseBodyType>;
 }
 
-function mockFetch(res: MockResponse) {
-  fetchMocker.mockResponse(() => res);
-}
+function useTestRequestHandler<
```
vipentti, line 89: Not sure if this is the best name either, naming things is hard :).

Possible alternatives: `useMockRequestHandler`, `mockRequestHandler`, `mockHandler`

drwpow:
This is a great PR, thank you! Very in favor of this direction—I had been wanting to do this but you beat me to it.

I’m not so much particular about the test setup details, so long as the tests themselves don’t change.

packages/openapi-fetch/test/index.test.ts
drwpow: This (and all the other names) are fine! Better than the current versions anyway

vipentti:
I don't have permissions to merge this, feel free to do so :)

drwpow:
This is great, thanks again!

drwpow:
@jshike this is for testing; there’s nothing to publish

Issue #1291, opened 2023-08-09
Error types cannot be used (expression of type '"..."' can't be used to index type 'NonNullable<FilterKeys<...>>)


EmilePerron:
**Description**

When making an API call using the library, using fairly simple code like below, the error object does not allow access to any of the keys that are supposed to be available.

Ex.:

```ts
const client = createClient<paths>({ baseUrl: apiBaseUrl, headers: { ...baseHeaders, Authorization: `Bearer ${jwt}` } });
const response = await client.post("/rtm/notes/create", { body: data });

if (response.error) {
  enqueueSnackbar({
    message: response.error["hydra:description"],
    variant: "error",
    autoHideDuration: 10000,
  });

  return;
}
```

Gives the following error when using `response.error["hydra:description"]` (or any other key/property on the error):

```
Element implicitly has an 'any' type because expression of type '"hydra:description"' can't be used to index type 'NonNullable<FilterKeys<{ "application/ld+json": { "@context": string; "@type": string; "hydra:title": string; "hydra:description": string; }; "application/json": { code?: string | undefined; message?: string | undefined; }; } | { ...; } | { ...; }, `${string}/${string}`>>'.
  Property 'hydra:description' does not exist on type 'NonNullable<FilterKeys<{ "application/ld+json": { "@context": string; "@type": string; "hydra:title": string; "hydra:description": string; }; "application/json": { code?: string | undefined; message?: string | undefined; }; } | { ...; } | { ...; }, `${string}/${string}`>>'.ts(7053)
```

Here is the Typescript error in VS Code:

<img width="963" alt="image" src="https://github.com/drwpow/openapi-typescript/assets/9383532/c6b50f75-c9e1-4362-b1d4-cec7ed96a8ed">

My OpenAPI schema for the responses is the following:
```yaml
400:
  description: 'Invalid input'
  content:
    application/ld+json:
      schema:
        type: object
        properties:
          '@context': { readOnly: true, type: string, nullable: false }
          '@type': { readOnly: true, type: string, nullable: false }
          'hydra:title': { readOnly: true, type: string, nullable: false }
          'hydra:description': { readOnly: true, type: string, nullable: false }
        required:
          - '@context'
          - '@type'
          - 'hydra:title'
          - 'hydra:description'
    application/json:
      schema:
        type: object
        properties:
          code: { readOnly: true, type: string, nullable: false }
          message: { readOnly: true, type: string, nullable: false }
        required:
          - '@context'
[...]

PR #1597, opened 2024-03-27
Allow to select the response content type

armandabric:
## Changes

This is a simple change that for now only allow to select the content type of the response. Inspired by the @NikolaStojicic patch. 

Fix https://github.com/drwpow/openapi-typescript/issues/1291 

This solution is not perfect but allow to fix the issue. Clearly a more robust solution could be to infer the `MediaType` by reading the `Accept` header. If multiple content and the `Accept` header not set, the lib could fallback to an `unknown` data.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/few-tomatoes-try.md
+++ b/.changeset/few-tomatoes-try.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Allow to select the response content type
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -111,10 +111,10 @@ export type RequestBodyOption<T> =
 export type FetchOptions<T> = RequestOptions<T> &
   Omit<RequestInit, "body" | "headers">;
 
-export type FetchResponse<T, O> =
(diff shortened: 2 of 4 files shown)


drwpow:
This is an elegant solution to allow a workaround while we come up with a better fix, thank you! While it is a core goal of this library to not have to use generics for anything, this does need a temporary workaround. Agree on not documenting it while a more automatic solution is explored.

drwpow:
This looks great! If you could just add a `patch` changeset (see comment), I’d be happy to merge

armandabric:
@drwpow I add the changeset as requested

drwpow (approved):

armandabric:
@drwpow Thanks for the merge. Can you publish a new version (even a `next`)?

drwpow:
> @drwpow Thanks for the merge. Can you publish a new version or just update the `next` tag?

Just published a new `next` version! Thanks again!

drwpow:
Er, sorry; got my PRs mixed up. Meant to say this is published at `openapi-fetch` latest. Went ahead and removed the `next` tag for now as it’s not needed at the moment (will use that later, perhaps when a 1.0 RC gets closer)

Issue #1598, opened 2024-03-27
Export component's and path's types with friendly names


enkot:
**Description**

First of all, I want to thank you for this great lib! :) 
To make it even better, I suggest exporting component types and paths like open-api-generator does.

**Proposal**

The first attempt to implement this is here https://github.com/drwpow/openapi-typescript/pull/1260.
The idea is to generate types like this:
Components:
```ts
export type SchemaSimpleUser = components["schemas"]["simple-user"];
export type SchemaInstallation = components["schemas"]["installation"];
...
export type ResponseNotFound = components["responses"]["not_found"];
export type ResponseRequiresAuthentication = components["responses"]["requires_authentication"];
...
export type ParameterPerPage = components["parameters"]["per-page"];
export type ParameterPage = components["parameters"]["page"];
...
export type HeaderLink = components["headers"]["link"];
export type HeaderXRateLimitLimit = components["headers"]["x-rate-limit-limit"];
```
Paths:
```ts
export type RequestEnterpriseAdminListGlobalWebhooks = paths["/admin/hooks"]["get"]["parameters"];
export type RequestEnterpriseAdminGetGlobalWebhook = paths["/admin/hooks/{hook_id}"]["get"]["parameters"];
```
It would be also nice to generate types for `requestBody` and `responses`, but not sure how to properly handle content type (`application/json`, `application/yaml` etc.)

We also need to resolve the duplication issue described in this comment https://github.com/drwpow/openapi-typescript/pull/1260#issuecomment-1658694431. This can be solved by adding a number as a suffix to the duplicate name.

**Checklist**

- [x] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

drwpow:
Just for cleanup: this already exists at #1414; closing this to track that.

This solution has been frequently-requested, but is fraught with problems and gotchas (since OpenAPI allows so many characters that are not valid JS identifiers, not to mention conflicts, it makes this approach non-deterministic and full of edge cases that amount to a significant amount of work, to result in what is only a cosmetic difference of what’s available in the library today.

The current status of this is if someone can open a PR that meets full requirements this will be accepted, but so far it’s been too much work to take on.

PR #1599, opened 2024-03-27
feat: root types

enkot:
## Changes

#1598 
Exports component's and path's types from the root:
```ts
export type RequestEnterpriseAdminListGlobalWebhooks = paths["/admin/hooks"]["get"]["parameters"];
export type RequestEnterpriseAdminGetGlobalWebhook = paths["/admin/hooks/{hook_id}"]["get"]["parameters"];
...
export type SchemaSimpleUser = components["schemas"]["simple-user"];
export type SchemaInstallation = components["schemas"]["installation"];
...
export type ResponseNotFound = components["responses"]["not_found"];
export type ResponseRequiresAuthentication = components["responses"]["requires_authentication"];
...
export type ParameterPerPage = components["parameters"]["per-page"];
export type ParameterPage = components["parameters"]["page"];
...
export type HeaderLink = components["headers"]["link"];
export type HeaderXRateLimitLimit = components["headers"]["x-rate-limit-limit"];
```
I reworked the transformers a bit to return an array where the first element is a TypeNode, the second is a list of aliases (TypeAliasDeclaration[]). For each element in the list, a corresponding "friendly name" type will be generated, alongside with `paths` and `components`.

Additionally, duplicate alias names for components are suffixed with an incremented number using the `renameDuplicates` function.

Names and implementation are subject to improve 🙈 

But it's an open question which level of nested types for `paths` should be aliased and what is the most convenient way to use them:
```ts
RequestEnterpriseAdminListGlobalWebhooks['query']
// or
RequestEnterpriseAdminListGlobalWebhooksQuery
// or
EnterpriseAdminListGlobalWebhooksQuery
// etc
```

- [ ] We must have tests for output. Not only testing ideal cases; test for:
  - [ ] Illegal JS chars
  - [x] Conflicts (foo.bar vs foo-bar—not only how do they get transformed to PascalCase, but is the output deterministic—do the conflicts resolve the same way regardless of document order every time?)
  - [ ] Deep-nested $refs
  - [ ] $defs support (does it work with root types?)
- [ ] Test for support for #/components/parameters, #/components/responses, and #/components/responseBodies (and testing for conflicts between all of them)
- [ ] Test for support for operation IDs

## How to Review

Run `./bin/cli.js` with `--root-types` flag.

## Checklist

- [ ] Unit tests updated


packages/openapi-typescript/package.json
```diff
@@ -75,6 +75,7 @@
     "del-cli": "^5.1.0",
     "esbuild": "^0.20.1",
     "execa": "^7.2.0",
+    "scule": "^1.3.0",
```
drwpow, line 78: I’m fine with adding this since it’s tiny, well-documented, and well-written. But this will need to be in `dependencies` if it’s required for runtime! Otherwise it won’t install, and users will get a “module not found” error

packages/openapi-typescript/src/transform/paths-object.ts
```diff
@@ -26,8 +27,9 @@ const PATH_PARAM_RE = /\{[^}]+\}/g;
 export default function transformPathsObject(
   pathsObject: PathsObject,
   ctx: GlobalContext,
-): ts.TypeNode {
+): [ts.TypeNode, ts.TypeAliasDeclaration[]] {
```
drwpow, line 30:
```suggestion
): { node: ts.TypeNode, aliases: ts.TypeAliasDeclaration[] } {
```

This is a personal opinion, but let’s not return tuple types for any functions; those don’t scale well. If a function needs to return multiple things, a **named object** is much better (same for parameters—named objects are more maintainable than ordered params or tuples)

packages/openapi-typescript/src/transform/index.ts
```diff
@@ -17,13 +17,18 @@ type SchemaTransforms = keyof Pick<
...
+  (
+    node: any, // eslint-disable-line @typescript-eslint/no-explicit-any
+    options: GlobalContext,
+  ) => [ts.TypeNode, ts.TypeAliasDeclaration[]]
 > = {
   paths: transformPathsObject,
   webhooks: transformWebhooksObject,
   components: transformComponentsObject,
-  $defs: (node, options) =>
+  $defs: (node, options) => [
```
drwpow, line 28: Another tuple type to remove: let’s change this to an `interface` instead

drwpow:
Thanks for opening this! This is off to a great start, but we need the following things:

**Required**
1. We must have tests for output. Not only testing ideal cases; test for:
  - [ ] Illegal JS chars
  - [ ] Conflicts (`foo.bar` vs `foo-bar`—not only how do they get transformed to PascalCase, but **is the output deterministic**—do the conflicts resolve the same way regardless of document order every time?)
  - [ ] Deep-nested `$refs`
  - [ ] `$defs` support (does it work with root types?)
2. Test for support for `#/components/parameters`, `#/components/responses`, and `#/components/responseBodies` (and—you guessed it—testing for conflicts between all of them)
  - You mentioned content types, but that shouldn’t factor in at all since the top-level keys for all these have “friendly names” and should all generate the same way
4. Test for support for operation IDs (agree or disagree, people will ask for this feature 😅)

I know it’s a lot to ask, but historically, flags have been implemented with gaps, and it’s led to frustration when a feature doesn’t work as expected. I’d like to avoid this for this flag, and release something that works for everyone and not just a limited usecase. And these problems were all reasons the original 1.x moved away from this naming structure. I think we can solve for these now, though, but it will take just a little more planning and testing this time

drwpow:
This PR has gone a while without updates. Closing not as a “won’t accept;” but just to leave room for others to pick up the work if-needed.

PR #1621, opened 2024-04-18
Add propertiesRequired option to Node.js API and CLI #1570

### LINKED ISSUE #1570: Is there some way to generate responses as non-optional even if not 'required' in the schema?
**Description**

I have a number of APIs that I need to talk to internally - however they all have openapi schemas generated using a library (C#'s Swashbuckle) which does not mark fields as required, even if they are not nullable in the original types.

This means every field in every response I get is optional:

```typescript
primary?: boolean;
```

even nullable fields are both nullable and optional:

```typescript
applicantId?: string | null;
```

I'm aware that the OpenAPI spec says the response fields need to be `required` explicitly - but it'd be nice to be able to override this where people aren't following the spec correctly.

**Proposal**

add a `--response-fields-non-optional` CLI parameter to allow all response fields to be treated as non-optional

**Checklist**

- [ ] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

TzviPM:
## Changes

Closes #1570.

Adds a `propertiesRequired` option to the Node.js API and global context and exposes this as `--properties-required` in the CLI.

### More nuanced description

Schema objects are allowed to have a `required` property that is set to an array of keys in `properties` that are considered required. The default for `required` is `[]` as specified by https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-6.5.3 and adopted by openapi via https://swagger.io/specification/#data-types. This may not be the intent of many API designers, so the `propertiesRequired` option changes the interpretation of section 6.5.3 from

> Omitting this keyword has the same behavior as an empty array.

to the following:

> Omitting this keyword has the same behavior as specifying an array containing the names of all properties in the instance.

## How to Review

Pay special attention to the naming of the option and cli flag. I'm open to any better names you can come up with 😄 My approach was inspired by the behaviour of `additionalProperties` and `defaultNonNullable`.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -101,6 +101,7 @@ The following flags are supported in the CLI:
 | `--alphabetize`           |       | `false`  | Sort types alphabetically                                                                                           |
 | `--array-length`          |       | `false`  | Generate tuples using array `minItems` / `maxItems`                                                                 |
 | `--default-non-nullable`  |       | `false`  | Treat schema objects with default values as non-nullable                                                            |
+| `--properties-required`   |       | `false`  | Treat schema objects without `required` as having all properties required.                                          |
 | `--empty-objects-unknown` |       | `false`  | Allow arbitrary properties for schema objects with no specified properties, and no specified `additionalProperties` |
 | `--enum`                  |       | `false`  | Generate true [TS enums](https://www.typescriptlang.org/docs/handbook/enums.html) rather than string unions.        |
 | `--exclude-deprecated`    |       | `false`  | Exclude deprecated fields from types                                                                                |
--- a/packages/openapi-typescript/test/transform/schema-object/object.test.ts
+++ b/packages/openapi-typescript/test/transform/schema-object/object.test.ts
@@ -29,6 +29,27 @@ describe("transformSchemaObject > object", () => {
       },
     ],
     [
+      "basic > options.propertiesRequired: tre",
+      {
+        given: {
+          type: "object",
+          properties: {
+            required: { type: "boolean" },
+            optional: { type: "boolean" },
+          },
+          required: ["required"],
+        },
+        want: `{
+    required: boolean;
+    optional?: boolean;
+}`,
+        options: {
+          ...DEFAULT_OPTIONS,
+          ctx: { ...DEFAULT_OPTIONS.ctx, propertiesRequired: true },
+        },
+      },
+    ],
+    [
       "empty",
       {
         given: { type: "object" },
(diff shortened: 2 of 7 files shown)


docs/cli.md
```diff
@@ -101,6 +101,7 @@ The following flags are supported in the CLI:
 | `--alphabetize`           |       | `false`  | Sort types alphabetically                                                                                           |
 | `--array-length`          |       | `false`  | Generate tuples using array `minItems` / `maxItems`                                                                 |
 | `--default-non-nullable`  |       | `false`  | Treat schema objects with default values as non-nullable                                                            |
+| `--properties-required`   |       | `false`  | Treat schema objects without `required` as having all properties required.                                          |
```
drwpow, line 104: Minor change, and this is a little more verbose, but is a little more descriptive as to what it does. 

```suggestion
| `--properties-required-by-default`   |       | `false`  | Treat schema objects without `required` as having all properties required.                                          |
```

packages/openapi-typescript/test/transform/schema-object/object.test.ts
```diff
@@ -28,6 +28,27 @@ describe("transformSchemaObject > object", () => {
         // options: DEFAULT_OPTIONS,
       },
     ],
+    [
+      "basic > options.propertiesRequired: tre",
```
drwpow, line 32: nit: not particular about the naming of the test, but it should just describe it’s not _just_ about setting the flag to `true`; it’s about the behavior of `true` + the required array

```suggestion
      "options > propertiesRequired + required array",
```

packages/openapi-typescript/test/transform/schema-object/object.test.ts
```diff
@@ -278,6 +299,19 @@ describe("transformSchemaObject > object", () => {
         },
       },
     ],
+    [
+      "options > propertiesRequired: true",
+      {
+        given: { type: "object", properties: { string: { type: "string" } } },
+        want: `{
+    string: string;
+}`,
```
drwpow, line 308: 👍 

Great test

drwpow:
This looks good! I’m going to fix CI on `main`, and will rebase off that when that’s fixed

drwpow (approved):
Thanks so much! If it’s all right, I’m going to merge since CI is failing, and there’s a lint error, but otherwise your PR looks great.

Since your PR allowed this, I think I’ll add a minor change to the flag but keep everything else as-is. Thanks! Will release this in a new release of `@next`

Issue #1631, opened 2024-04-26
Send init parameter(s) into fetch requests which is currently unused (= undefined)


mellster2012:
**Description**
v0.9.3, 0.9.x
Need a way to set the dispatcher in the request, as setGlobalDispatcher doesn't work to disable certificate checking

**Proposal**
replacing line 107 in dist/index.cjs from
    `response = await fetch(request);`
to
    `let response = await fetch(request, requestInit);`
    will allow to pass a custom dispatcher and other parameters via
```
const agent = new Agent({
  connect: {
    rejectUnauthorized: false,
  },
});
```
and
```
export const client = createClient<paths>({
  baseUrl: myUrl,
  dispatcher: agent as never,
});
```

**Checklist**

- [x] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

drwpow:
Thanks for opening, but can you describe what a “dispatcher” is? Is this a concept from another library? Why won’t the custom `fetch` or middleware work? What prior art are you referencing?

I’m not opposed to adding this, but I want to be 100% sure before we commit to adding breaking changes to the API there’s not a simpler solution to solve the problem (and I do not understand the problem)

mellster2012:
Thanks for the comments!
* Previously you could provide an Agent to override default configurations, such as not failing on self-signed certificates. Node's global fetch does not use the http/https stack, but an internally wired undici client and you cannot replace the global agent easily, see here: https://stackoverflow.com/questions/73817412/why-is-the-agent-option-not-available-in-node-native-fetch
* Replacing fetch with a custom implementation is a much larger change than providing a custom dispatcher - which is an undici specific Agent different from Node's http(s) Agent. Also we've ran into signature issues/incompatibilities when swapping out fetch. So this seems like the simplest solution (as suggested in the linked SO thread above) to pass custom configurations.
* We already forked openapi-fetch for this, there must be many others with the same need. AFAIK adding an optional dispatcher to the configuration option should not be breaking changes to the (public facing) API, but I could be wrong.

mellster2012:
There actually seems to be an option to modify the globalFetch Dispatcher:
`globalThis[Symbol.for('undici.globalDispatcher.1')] = yourDispatcher;`
[...]

PR #1636, opened 2024-04-27
Allow client option for custom dispatcher into fetch requests (e.g. to disable certificate validation) #1631

mellster2012:
## Changes

Fix for https://github.com/drwpow/openapi-typescript/issues/1631

## How to Review

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/stale-donuts-smoke.md
+++ b/.changeset/stale-donuts-smoke.md
@@ -1,5 +1,0 @@
----
-"openapi-fetch": minor
----
-
-Add support for `client["/endpoint"].GET()` style calls
--- a/packages/openapi-fetch/CHANGELOG.md
+++ b/packages/openapi-fetch/CHANGELOG.md
@@ -1,5 +1,11 @@
 # openapi-fetch
 
+## 0.11.0
+
+### Minor Changes
+
+- [#1791](https://github.com/openapi-ts/openapi-typescript/pull/1791) [`a956d5d`](https://github.com/openapi-ts/openapi-typescript/commit/a956d5d8480834402536283ee2f24ce8086698dc) Thanks [@gzm0](https://github.com/gzm0)! - Add support for `client["/endpoint"].GET()` style calls
+
 ## 0.10.6
 
 ### Patch Changes
--- a/packages/openapi-fetch/package.json
+++ b/packages/openapi-fetch/package.json
@@ -1,7 +1,7 @@
 {
   "name": "openapi-fetch",
   "description": "Fast, type-safe fetch client for your OpenAPI schema. Only 5 kb (min). Works with React, Vue, Svelte, or vanilla JS.",
-  "version": "0.10.6",
+  "version": "0.11.0",
   "author": {
     "name": "Drew Powers",
     "email": "drew@pow.rs"
--- a/packages/openapi-react-query/CHANGELOG.md
+++ b/packages/openapi-react-query/CHANGELOG.md
@@ -1,5 +1,12 @@
 # openapi-react-query
 
+## 0.1.1
+
+### Patch Changes
+
+- Updated dependencies [[`a956d5d`](https://github.com/openapi-ts/openapi-typescript/commit/a956d5d8480834402536283ee2f24ce8086698dc)]:
+  - openapi-fetch@0.11.0
+
 ## 0.1.0
 
 ### Minor Changes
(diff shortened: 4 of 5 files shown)


mellster2012:
Is there anything we can do to re-run failed workflows? Thanks!

drwpow:
> Is there anything we can do to re-run failed workflows? Thanks!

Unfortunately no; there’s just a requirement that new contributors need to have manually-run workflows until they’ve gotten a PR merged. I think it’s just GitHub saving $ (which is fine by me, because of how generously they provide so many resources for OSS).

drwpow:
See comment on #1631. I think we need a little more discussion / clarification on what this PR is accomplishing, because on first glance it’s not clear. So that’s priority 1.

I’ll also leave a few comments on this PR—even if it gets reworked—just to give a sense of what we’re looking for when contributing to this project.

packages/openapi-fetch/test/index.with.dispatcher.test.ts
```diff
@@ -0,0 +1,1608 @@
...
+      throw new Error(`No request handler found for ${request.method} ${request.url}`);
+    },
+  });
+});
+
+afterEach(() => server.resetHandlers());
+
+afterAll(() => server.close());
+
+describe("TypeScript checks", () => {
```
drwpow, line 25: Appreciate you adding so many test cases to this PR, but we **don’t** want to just flat-out copy the entire test suite just for one change. I would suspect most of the tests don’t actually test this code path at all; we probably only need 1 or 2 new tests to start.

There may be some confusion with the v7-beta.test.ts file; that’s a separate case. It’s because when openapi-typescript 7.x gets a stable release, we’ll have to make a breaking release for openapi-fetch as well, and I’m just making sure that transition is smooth. But again, for a single option, we don’t need to copy all the tests

package.json
```diff
@@ -20,6 +20,7 @@
     "@changesets/cli": "^2.27.1",
     "del-cli": "^5.1.0",
     "prettier": "^3.2.5",
-    "typescript": "^5.4.5"
+    "typescript": "^5.4.5",
+    "undici": "^6.14.1"
```
drwpow, line 24: We’re only using `undici` in a single test in `openapi-fetch`, so we want to add it to devDependencies in that `package.json`; not for the entire monorepo.

packages/openapi-fetch/src/index.d.ts
```diff
@@ -15,6 +15,8 @@ import type {
 export interface ClientOptions extends Omit<RequestInit, "headers"> {
   /** set the common root URL for all API requests */
   baseUrl?: string;
+  /** custom dispatcher */
+  dispatcher?: unknown;
```
drwpow, line 19: We’ll want a more accurate type for this.

packages/openapi-fetch/test/index.with.dispatcher.test.ts
mellster2012: Removed the extra test file and just added one test under options specifying a custom dispatcher - without it hitting a 'bad cert' website it only validates that specifying a custom dispatcher does not break general functionality.

package.json
mellster2012: Fixed.

packages/openapi-fetch/src/index.d.ts
mellster2012: Fixed.

mwotton:
any chance of this getting merged? i'm getting hit with this problem with a long-running fetch call (undici hardcodes the timeout when used through the fetch api)

drwpow:
Apologies; I dropped off reviewing this PR because it still had failing tests, and it seemed like there was still work to be done. @mellster2012 if we get all the tests and checks passing again, I’d be open to approving and merging 🙂. Thanks for addressing my previous comments, by the way!

mellster2012:
> Apologies; I dropped off reviewing this PR because it still had failing tests, and it seemed like there was still work to be done. @mellster2012 if we get all the tests and checks passing again, I’d be open to approving and merging 🙂. Thanks for addressing my previous comments, by the way!

Sure, I will definitely work on it and resubmit once everything is back in sync and tests are passing, but it may take a week or more until I can get started on it.

mellster2012:
Redone PR here: https://github.com/openapi-ts/openapi-typescript/pull/1837

PR #1645, opened 2024-04-29
Fix Redocly config load error

drwpow:
## Changes

- Fixes issue with loading some parts of Redocly configs
- Also adds support for OpenAPI schemas declared in [JSON5](https://www.npmjs.com/package/json5) format cuz why not (also frankly any parser errors are better than the default `JSON.parse()` errors)

## How to Review

- Tests added; tests should pass.

## Checklist

- [x] Unit tests updated


packages/openapi-typescript/test/cli.test.ts
```diff
@@ -131,5 +129,16 @@ describe("CLI", () => {
...
     });
+
+    test("lint error", async () => {
+      const cwd = new URL("./fixtures/redocly-lint-error", import.meta.url);
+
+      try {
+        await execa("../../../bin/cli.js", { cwd: fileURLToPath(cwd) });
+        throw new Error("Linting should have thrown an error");
+      } catch (err) {
+        expect(stripAnsi(String(err))).toMatch(/✘ {2}Servers must be present/);
```
drwpow, line 140: Not exhaustive, but strikes a good balance between asserting that the linting throw an error with a specific message, vs tying tests too-heavily to Redocly’s specific output that could change.

packages/openapi-typescript/bin/cli.js
```diff
@@ -92,7 +95,7 @@ async function generateSchema(schema, { redoc, silent = false }) {
       exportType: flags.exportType,
       immutable: flags.immutable,
       pathParamsAsTypes: flags.pathParamsAsTypes,
-      redoc,
+      redocly,
```
drwpow, line 98: Turns out you have to name variables the same in order for them to get passed properly! Whodathunk

packages/openapi-typescript/src/lib/redoc.ts
```diff
@@ -64,7 +65,7 @@ export async function parseSchema(schema: unknown, { absoluteRef, resolver }: Pa
     if (schema[0] === "{") {
       return {
         source: new Source(absoluteRef, schema, "application/json"),
-        parsed: JSON.parse(schema),
+        parsed: parseJson(schema),
```
drwpow, line 68: Just adds nicer syntax errors for better DX. In my benchmarks I can’t notice a significant performance difference between this and native `JSON.parse()`.

PR #1646, opened 2024-04-29
bump openapi-typescript-fetch version

ajaishankar:
## Changes

Bumps the version of openapi-typescript-fetch for benchmark

## How to Review

No code changes, all tests should pass

## Checklist

- [ ] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/package.json
+++ b/packages/openapi-fetch/package.json
@@ -71,7 +71,7 @@
     "msw": "^2.2.14",
     "openapi-typescript": "^6.7.5",
     "openapi-typescript-codegen": "^0.25.0",
-    "openapi-typescript-fetch": "^1.1.3",
+    "openapi-typescript-fetch": "^2.0.0",
     "superagent": "^9.0.0",
     "typescript": "^5.4.5",
     "vitest": "^1.5.2"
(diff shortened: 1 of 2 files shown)


drwpow (approved):
Thank you! You know I think I need to use a different method of running benchmarks, because Vitest’s is still marked beta and is unreliable. Will re-run benchmarks, and if you have any other solutions / suggestions, I’m all ears 🙏

ajaishankar:
Thanks!

Was looking into the [Hono](https://hono.dev/) framework and they use [Mitata](https://github.com/honojs/hono/blob/main/benchmarks/utils/src/loop.js) for their benchmarks

Not sure if it would make a difference.

Feel the benchmark tests should exercise more functionality like request/response serialization and middleware.

That would be a better comparison.

Sorry have not got a chance to look into adding those...

btw you might find their [RPC](https://hono.dev/guides/rpc) strongly typed client/server implementation interesting...

drwpow:
Will take a look at Mitata! Probably anything will be an improvement over Vitest’s benchmarks. 

> btw you might find their [RPC](https://hono.dev/guides/rpc) strongly typed client/server implementation interesting...

Oh that’s interesting! It looks like it uses Zod, which I personally wouldn’t want to use in this project because of the weight/runtime cost. But I’m sure in the inference piping it’s doing, there are probably some interesting ideas worth looking into. Good suggestion!

PR #1653, opened 2024-05-03
add custom properties to request object

FreeAoi:
## Changes

This pull request lets request object attach custom properties like `next` prop on nextjs environment.

image
image

## How to Review
It should works

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/red-pants-cover.md
+++ b/.changeset/red-pants-cover.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Let request object have custom properties
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -166,7 +166,7 @@ export type ClientMethod<Paths extends Record<string, PathMethods>, M extends Ht
   I extends MaybeOptionalInit<Paths[P], M>,
 >(
   url: P,
-  ...init: HasRequiredKeys<I> extends never ? [I?] : [I]
+  ...init: HasRequiredKeys<I> extends never ? [(I & { [key: string]: unknown })?] : [I]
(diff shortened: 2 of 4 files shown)


drwpow:
Thanks for adding. Would it be possible to add a test, to ensure the correct usage is working as intended?

FreeAoi:
Sure

FreeAoi:
How I should the unit test?

drwpow:
> How I should do the unit test?

Inside the [main test suite, inside `middleware`](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/test/index.test.ts#L828), I’d just write a test that creates example middleware and adds the arbitrary property.

Then from the MSW side, assert that the server received the request with that additional property on it (see the [can modify request test](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/test/index.test.ts#L829)) as an example).

The bonus part of adding a test is **the types will be validated, too.** The code you wrote looks good, and it looks like it achieves the desired result! But adding a test guarantees it works as-advertised, and doesn’t throw any errors in TypeScript either.

FreeAoi:
Should be ready, check it out and let me know your thoughts

drwpow (approved):
This is great, thank you! Fantastic test!

drwpow:
🤔 I’m wondering if we need to add docs for this; this may just be an invisible feature. I’m going to release this, and will skip docs for now, but also open to a PR adding docs if it’s helpful.

FreeAoi:
Hi, I think write docs for this is not necessary because It is not common (it could very a bad practice in fact) monkeypatching request object.

This pr it's principal objective is let `request` object support custom properties in cases where frameworks like nextjs modified the request interface

PR #1672, opened 2024-05-19
fix: only add custom properties that aren't in the request object

### LINKED ISSUE #1666: Uncaught error when body is `undefined` in request init
**Description**

> Uncaught TypeError: Cannot set property body of #<Request> which has only a getter

Screenshot 2024-05-17 at 4 53 14 PM

https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/src/index.js#L16-L20

Basically, `body` is a getter on the `Request` object and the linked code is trying to write to it. One work around could be to wrap it in a try/catch. Otherwise I think you'll have to ignore some properties in the request init

**Reproduction**

Add a `body` to the request init object in `openapi-fetch@0.9.6` e.g. `client.POST('/', {body:undefined})`.

Happens in all major browsers AFAICT

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

jaredLunde:
## Changes

Closes #1666

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/wet-apples-hide.md
+++ b/.changeset/wet-apples-hide.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Fixes issue where native properties were not excluded from custom properties in the CustomRequest class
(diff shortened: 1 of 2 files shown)


drwpow (approved):
Great fix! 🤦 I always get bitten by this

PR #1681, opened 2024-05-27
requestBody fields optional even with defaults

fitztrev:
## Changes

Noticed that when a request body field had a `default` value, that it was being specified as required. This would change it to be treated like a parameter with a default value.

## How to Review

Let me know if the premise correct and then I could look into updating the tests.

## Checklist

- [x] Unit tests updated


drwpow:
I think this is an improvement, thank you! Yes if you update the tests and it’s clear there are no breakages I’d be open to this change.

drwpow:
This looks great, thanks again! Examples updated makes sense, too.

To merge this, would love to add a test, maybe in `request-body-object.test.ts`. The changes to the examples look correct, but those don’t always get checked for every single change so a unit test would ensure this doesn’t regress in the future. Great work!

fitztrev:
Thanks! Good call, added a test

drwpow (approved):
Amazing! Thank you 🙏

PR #1697, opened 2024-06-17
Expose original request on Middleware.onResponse

armandabric:
## Changes

Add the `req` as thrid params of the onResponse middleware. Its useful to create log on the on response (the `Response` did not contains some information of the original request)

## Checklist

- [x] Unit tests updated
- [x] `docs/` updated (if necessary) (will be done if

### DIFF
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -144,7 +144,11 @@ export function onRequest(
   req: MiddlewareRequest,
   options: MergedOptions,
 ): Request | undefined | Promise<Request | undefined>;
-export function onResponse(res: Response, options: MergedOptions): Response | undefined | Promise<Response | undefined>;
+export function onResponse(
+  res: Response,
+  options: MergedOptions,
+  req: MiddlewareRequest,
+): Response | undefined | Promise<Response | undefined>;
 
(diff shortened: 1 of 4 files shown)


packages/openapi-fetch/src/index.d.ts
```diff
@@ -144,7 +144,11 @@ export function onRequest(
   req: MiddlewareRequest,
   options: MergedOptions,
 ): Request | undefined | Promise<Request | undefined>;
-export function onResponse(res: Response, options: MergedOptions): Response | undefined | Promise<Response | undefined>;
+export function onResponse(
+  res: Response,
+  options: MergedOptions,
+  req: MiddlewareRequest,
```
drwpow, line 150: 👍 This is a good non-breaking change. Will probably opt  for a more “logical” ordering in a future breaking change

drwpow (approved):
Great addition. Thanks for adding! There’s no reason not to have this available, and it wasn’t ever omitted intentionally.

armandabric:
Two suggestions for the v7:
- it could nice de rework the paramater order `(req, res, options)`
- and avoid to surcharge the standard Request and pass the data in the `options` arguments

drwpow:
@armandabric suggestions taken, and updated in `0.10.0`!

PR #1698, opened 2024-06-17
Support corporate proxy based on standard http_proxy environment vari…

Reggino:
…able

## Changes

This change will make the CLI command support an environment variable called ```http_proxy```. This is a standard environment variable in environments where a http proxy is required to access the internet.

## How to Review

Setup a http proxy server. Then start the openapi-typescript CLI command with an environment variable http_proxy. E.g.:

```
http_proxy=http://myproxy.company.lan:1234 ./node_modules/.bin/openapi-typescript 

```

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -203,3 +203,9 @@ This results in more explicit typechecking of array lengths.
 _Note: this has a reasonable limit, so for example `maxItems: 100` would simply flatten back down to `string[];`_
 
 _Thanks, [@kgtkr](https://github.com/kgtkr)!_
+
+## Proxy support
+
+When the use of a corporate proxy is required to access the internet and/or download external definitions, please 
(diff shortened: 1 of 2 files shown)


drwpow:
I’m in favor of this change, but could you add docs for this please? This feels like it’s worth a mention somewhere (anywhere is fine, but probably `cli.md` would be my recommendation)

Reggino:
Thanks, i've updated the documentation.

drwpow:
This looks great, but I think you meant to open the PR against the `6.x` branch? Currently `main` is for the upcoming `7.x` version that will be released soon, and it no longer relies on `undici` (just uses native Node fetch).

I tried changing the base branch, but it resulted in a pretty confused git history so I think you’ll have to rebase it against 6.x on your end. But with that change, happy to approve & merge (feel free to open another PR if it’s easier, too)

Issue #1694, opened 2024-06-14
Empty body should not result in Content-Type header being set


goce-cz:
**Description**

Any HTTP method is free to choose whether it sends a payload or not (GET might be debatable, but that's not important).
When it sends one, it should declare what type it is, by providing `Content-Type` header. When it doesn't send one, it **MUST NOT** declare the header.

The current implementation declares `application/json` for every call, even `GET` without specifying any body:

We are currently using a workaround through a middleware:

```ts
client.use({
  onRequest: req => {
    if (req.body === null) {
      req.headers.delete('Content-Type')
    }
    return req
  },
})
```

I know that there are at least two closed issues related to this, but I want to emphasize that this indeed is a bug which makes the library non-compliant with rules of HTTP protocol. I really like the library and I would like to remove this little stain to make it shine even more.

**Reproduction**

Just fire a POST request against a strict API server, such as [Fastify](https://fastify.dev/).

```ts
await client.POST('/some-endpoint')
```
and see it respond with 400 Bad Request:
```
Body cannot be empty when content-type is set to 'application/json'
```

Fastify rightfully complains about the malformed incoming request.

**Expected result**

There should be no `Content-Type` header present, when we're not sending any body - regardless of the HTTP method.

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

eduardohilariodev:
This also makes test fail when mocking a GET request

drwpow:
> I want to emphasize that this indeed is a bug which makes the library non-compliant with rules of HTTP protocol. 

Thanks for raising this. Ideally this follows the `fetch` implementation as closely as possible, so that it has interop everywhere. It’s why despite requests to throw error on non-2XX status codes, this library won’t because that deviates from the spec.

Some QoL improvements were added around the native fetch implementation, of course, but if there’s ever a departure between spec-compliance and opinion, this library will choose spec compliance every time. In that regard, I do think you make a good argument that this is a bug, and not intentional behavior. I don’t think changing this default behavior would significantly negatively impact any consumers, but removing it would yield benefits.

The closed issues you’re referring to hopefully were regarding other aspects of the behavior (but if they made the same point, apologies if I missed something or failed to understand before).

So all that to say **I’d welcome a PR to change this behavior**, and it’ll get released as a breaking change just for safety.

[...]

Issue #1695, opened 2024-06-14
`null` and `''` (empty string) are serializable to JSON and should therefore be considered a valid body payload


goce-cz:
**Description**

Calling the client with `body: null` or `body: ''` will result in no body sent to the server (same behavior as with `undefined`).
I believe that they should be serialized to `null` and `""` respectively as these are valid JSON values and should be sent to the server.

I believe the problem lies in this condition:
https://github.com/drwpow/openapi-typescript/blob/19ab7342594df2ec7bb60d891818fe22b3e96e38/packages/openapi-fetch/src/index.js#L79-L81

where it should rather use strict comparison: `requestInit.body !== undefined`.

While the issue manifests only with `null` and `''`, the problem is bigger here as any other falsy value (`false`, `0`) will bypass the body serializer too. They are sent "correctly" to the server, just because they fall back to default serialization provided by the browser, where `null` and `''` is apparently considered as "no-body".

**Reproduction**

```ts
client.POST('/some-endpoint', {
  body: null // or ''
})
```
results in no body being sent.

```ts
const client = createClient<paths>({
  bodySerializer: body => {
    console.log('serializing')
    return JSON.stringify(body)
  },
})

client.POST('/some-endpoint', {
  body: 0 // or false
})
```
results in body being sent, but does not yield `serializing` in the console.

**Expected result**

All mentioned values go through the serializer and are sent to the server.

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

drwpow:
This logic makes sense! This seems like an improvement.

PR #1702, opened 2024-06-19
fix: add verification when request body is falsy

eduardohilariodev:
## Changes

https://github.com/openapi-ts/openapi-typescript/issues/1695
#1694

## How to Review

The `instanceof` was throwing an error when the body was falsy.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -80,7 +80,7 @@ export default function createClient(clientOptions) {
       requestInit.body = bodySerializer(requestInit.body);
     }
     // remove `Content-Type` if serialized body is FormData; browser will correctly set Content-Type & boundary expression
-    if (requestInit.body &&requestInit.body instanceof FormData) {
+    if (requestInit.body && requestInit.body instanceof FormData) {
       requestInit.headers.delete("Content-Type");
     }
     let request = new CustomRequest(createFinalURL(url, { baseUrl, params, querySerializer }), requestInit);
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -80,7 +80,7 @@ export default function createClient(clientOptions) {
       requestInit.body = bodySerializer(requestInit.body);
     }
     // remove `Content-Type` if serialized body is FormData; browser will correctly set Content-Type & boundary expression
-    if (requestInit.body instanceof FormData) {
+    if (requestInit.body &&requestInit.body instanceof FormData) {
       requestInit.headers.delete("Content-Type");
     }
     let request = new CustomRequest(createFinalURL(url, { baseUrl, params, querySerializer }), requestInit);
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -78,10 +78,10 @@ export default function createClient(clientOptions) {
     };
     if (requestInit.body) {
       requestInit.body = bodySerializer(requestInit.body);
-    }
-    // remove `Content-Type` if serialized body is FormData; browser will correctly set Content-Type & boundary expression
-    if (requestInit.body && requestInit.body instanceof FormData) {
-      requestInit.headers.delete("Content-Type");
+      // remove `Content-Type` if serialized body is FormData; browser will correctly set Content-Type & boundary expression
+      if (requestInit.body instanceof FormData) {
+        requestInit.headers.delete("Content-Type");
+      }
     }
     let request = new CustomRequest(createFinalURL(url, { baseUrl, params, querySerializer }), requestInit);
     // middleware (request)


drwpow (approved):
Good fix! I’m fine merging this without tests as it’s simple enough, but would also love a test that prevents this regressing if possible.

eduardohilariodev:
> Good fix! I’m fine merging this without tests as it’s simple enough, but would also love a test that prevents this regressing if possible.

I am getting this error when running `pnpm test`

```sh
pnpm test

> openapi-fetch@0.9.7 pretest /home/eduardo/dev/forks/openapi-typescript/packages/openapi-fetch
> pnpm run generate-types


> openapi-fetch@0.9.7 generate-types /home/eduardo/dev/forks/openapi-typescript/packages/openapi-fetch
> node ./scripts/generate-types.js

file:///home/eduardo/dev/forks/openapi-typescript/node_modules/.pnpm/execa@8.0.1/node_modules/execa/lib/error.js:60
                error = new Error(message);
                        ^

Error: Command failed with exit code 1: ../openapi-typescript/bin/cli.js ./test/fixtures/api.yaml -o ./test/fixtures/v7-beta.d.ts
node:internal/modules/esm/resolve:264
    throw new ERR_MODULE_NOT_FOUND(
          ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/eduardo/dev/forks/openapi-typescript/packages/openapi-typescript/dist/index.js' imported from /home/eduardo/dev/forks/openapi-typescript/packages/openapi-typescript/bin/cli.js
    at finalizeResolution (node:internal/modules/esm/resolve:264:11)
    at moduleResolve (node:internal/modules/esm/resolve:917:10)
    at defaultResolve (node:internal/modules/esm/resolve:1130:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:396:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:365:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:240:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:85:39)
    at link (node:internal/modules/esm/module_job:84:36) {
  code: 'ERR_MODULE_NOT_FOUND',
  url: 'file:///home/eduardo/dev/forks/openapi-typescript/packages/openapi-typescript/dist/index.js'
}

Node.js v20.11.1
    at makeError (file:///home/eduardo/dev/forks/openapi-typescript/node_modules/.pnpm/execa@8.0.1/node_modules/execa/lib/error.js:60:11)
    at handlePromise (file:///home/eduardo/dev/forks/openapi-typescript/node_modules/.pnpm/execa@8.0.1/node_modules/execa/index.js:124:26)
    at async Promise.all (index 1)
    at async generate (file:///home/eduardo/dev/forks/openapi-typescript/packages/openapi-fetch/scripts/generate-types.js:13:3) {
  shortMessage: 'Command failed with exit code 1: ../openapi-typescript/bin/cli.js ./test/fixtures/api.yaml -o ./test/fixtures/v7-beta.d.ts',
  command: '../openapi-typescript/bin/cli.js ./test/fixtures/api.yaml -o ./test/fixtures/v7-beta.d.ts',
  escapedCommand: '"../openapi-typescript/bin/cli.js" "./test/fixtures/api.yaml" -o "./test/fixtures/v7-beta.d.ts"',
  exitCode: 1,
  signal: undefined,
  signalDescription: undefined,
  stdout: '',
  stderr: 'node:internal/modules/esm/resolve:264\n' +
    '    throw new ERR_MODULE_NOT_FOUND(\n' +
    '          ^\n' +
    '\n' +
    "Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/eduardo/dev/forks/openapi-typescript/packages/openapi-typescript/dist/index.js' imported from /home/eduardo/dev/forks/openapi-typescript/packages/openapi-typescript/bin/cli.js\n" +
    '    at finalizeResolution (node:internal/modules/esm/resolve:264:11)\n' +
    '    at moduleResolve (node:internal/modules/esm/resolve:917:10)\n' +
    '    at defaultResolve (node:internal/modules/esm/resolve:1130:11)\n' +
    '    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:396:12)\n' +
    '    at ModuleLoader.resolve (node:internal/modules/esm/loader:365:25)\n' +
    '    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:240:38)\n' +
    '    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:85:39)\n' +
    '    at link (node:internal/modules/esm/module_job:84:36) {\n' +
    "  code: 'ERR_MODULE_NOT_FOUND',\n" +
    "  url: 'file:///home/eduardo/dev/forks/openapi-typescript/packages/openapi-typescript/dist/index.js'\n" +
    '}\n' +
    '\n' +
    'Node.js v20.11.1',
  cwd: '/home/eduardo/dev/forks/openapi-typescript/packages/openapi-fetch/',
  failed: true,
  timedOut: false,
  isCanceled: false,
  killed: false
}

Node.js v20.11.1
 ELIFECYCLE  Command failed with exit code 1.
 ELIFECYCLE  Command failed with exit code 1.
 ```

**Comment by drwpow** [comment]
You’ll need to run `pnpm run build` first (either in the root, or in `packages/openapi-typescript`; doesn’t matter).

The [Contributing guide](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md#testing) has some additional information on testing locally.

**Comment by drwpow** [comment]
> I am having problems in understanding the trigger for the removal of this header. I am not familiar with the inner workings of this lib.

At a high level, there are 2 ways of interpreting #1694:

1. We should unset `"content-type"` for all `GET` requests
2. We should respect the user’s preference if they set `"content-type"`, but not send that header by default for `GET` requests.

This is opinionated, but the first option seems overreaching, and having read a lot of bug reports and strange ways APIs behave, I feel confident that there will likely be some users that have APIs that require `"content-type"` to be set on `GET` requests (regardless of what any specification says). The second option feels much safer, as it ultimately lets the user decide, but also provides good defaults. **I would strongly suggest the second approach.**

For #1695, https://github.com/openapi-ts/openapi-typescript/issues/1694, what we want to accomplish is only set `"content-type": "application/json"` if body is not `undefined` (`if (requestInit.body !== undefined)`). 

Solving the 2 together, on [this line](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-fetch/src/index.js#L40) where we initialize the headers, we should **just remove that.** Unsetting a header is tricky **because we don’t know if we set it or the user.** So we can just delay that until the actual request.

Later, when we [create the request object](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-fetch/src/index.js#L77), we are again merging headers (which had the default `"content-type": "application/json"` header unless the user overrode it). Right before that happens, we want to add something like:

```diff
+     if (requestInit.body !== `undefined`) {
+       baseHeaders = mergeHeaders(DEFAULT_HEADERS, baseHeaders);
+     }
+ 
      const requestInit = {
```

Which we couldn’t do earlier, because we didn’t know if there was a body on the request or not. Also, making sure `DEFAULT_HEADERS` is first in the list is important, because any conflict will make sure it isn’t set.

> I made the start of a test, but I don't think it's quite it:

The existing logic is there because we _want_ to have `"content-type": "multipart/form-data"` when FormData is passed for the body. That’s unrelated to your fix. Your test(s) needs to cover:

1. For all `GET` requests, `"content-type":` is `undefined` by default (when the user never specified)
4. For `GET` requests, `"content-type"` keeps options passed into `createClient()` if the user manually specified one
5. For `GET` requests, `"content-type"` keeps any value passed in on individual requests (overrides)
6. For (any non-`GET`; doesn’t matter which) requests, `"content-type"` is `"application/json"` by default

With the danger here being we accidentally invert the order, and unset `"content-type"` if the user has manually declared it.

Remember that the approach would differ if we had chosen the first option in the beginning—we’d simply delete that header regardless of what the user did. But as that feels more dangerous/unpredictable, we structured our code and the tests to make sure we don’t override users’ opinions, and instead just improve the defaults.

packages/openapi-fetch/src/index.js
```diff
@@ -80,7 +80,7 @@ export default function createClient(clientOptions) {
       requestInit.body = bodySerializer(requestInit.body);
     }
     // remove `Content-Type` if serialized body is FormData; browser will correctly set Content-Type & boundary expression
-    if (requestInit.body instanceof FormData) {
+    if (requestInit.body && requestInit.body instanceof FormData) {
```
imranbarbhuiya, line 83:
```suggestion
    if (!requestInit.body || requestInit.body instanceof FormData) {
```

altho out of scope for this pr but we should delete the content type when body is empty, currently there's no proper way to remove the body and when I don't need to send any body to backend, backend throws content type json was used but no body provided

imranbarbhuiya (changes requested):

packages/openapi-fetch/src/index.js
drwpow: Yeah I may have approved prematurely because I didn‘t realize at first this was attempting to solve 2 issues with 1 PR. My suggestions should be a ~2-line change in code, and a couple test additions to satisfy both.

eduardohilariodev:
@drwpow Do you think it's necessary to perform these tests on this specific PR? I am encountering this bug every time I reinstall this lib, having to change it locally to be able to test my app (with Vitest). Do you think your suggestions can be put into another issue that has a more comprehensive context? The way I see them, they alter the library's behaviour (much more overarching than this simple hotfix). Please, do correct me if I am ignorant about this.

drwpow:
> @drwpow Do you think it's necessary to perform these tests on this specific PR? I am encountering this bug every time I reinstall this lib, having to change it locally to be able to test my app (with Vitest). Do you think your suggestions can be put into another issue that has a more comprehensive context? The way I see them, they alter the library's behaviour (much more overarching than this simple hotfix). Please, do correct me if I am ignorant about this.

Yes we can merge as-is, if you’d like. But I mostly asked for tests because in order to fix all the problems outlined in #1694 and #1965, we’d probably have to overwrite your changes here (which would be a slight behavior change, as you pointed out, but IMO nothing significant or breaking). I believe your original issue overlaps a _little_ with those issues, but is actually different. Tests are just good reinforcers that when you write code, it sticks around (because you taking your valuable time to open a PR and fix an issue means a lot! It really does. And this project is only possible from people like yourself. And tests just ensure your good work sticks around).

Given the time on this, and the need to fix #1694 and #1695, I’ll merge, and push a fix for those separately. But feel free to open another issue (or another PR) if those changes cause a regression for your scenario.

Issue #1712, opened 2024-06-20
Remove nanoid


imranbarbhuiya:
**Description**
_A brief description of the bug. Provide either a screenshot or the full error message_

Hi, thanks for this awesome lib, the new version added `nanoid` for middleware usage, I'm using this package in my rn project and `crypto` isn't available there. Not only rn but there's other platforms/runtimes too where crypto available so adding that to dep and using it even when we arent using middleware makes it impossible for me to update. For now, I patched the package to remove `nanoid`. I can create a pr to fix it but need to discuss what should we use instead.

**Reproduction**
_How can this be reproduced / when did the error occur? Does the issue occur in a specific browser, or all browsers?_
Use the package in any env where crypto isn't available

**Expected result**
It should work and specially when I'm not even using middleware so id is useless for me.
_(in case it’s not obvious)_

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

drwpow:
I’m not particular about the library used; it’s swappable. Actually looking at benchmarks I would like to use something faster, but without incurring weight. I may even end up handrolling something because it doesn’t need to be secure, or that unique (it can collide much more often because it’s only meant to compare requests right next to each other, not months apart)

imranbarbhuiya:
can we use simple `Math.random().toString(36).slice(2, 11)`?

drwpow:
> can we use simple `Math.random().toString(36).slice(2, 11)`?

That’s not a bad suggestion! Though this is something I do want to split hairs on to at least match nanoid’s performance and ideally beat if possible. For many libraries that’s overkill, but a design decision of this library is to be at the center of every fetch request it should be functionally equivalent to native `fetch()` and be as thin and performant as absolutely possible. So microoptimizations are important here IMO.

drwpow:
The package’s [CONTRIBUTING guide](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md#testing) has some basic instructions on test setup. You can also inspect `.github/workflows/ci.yml` to see the exact commands CI runs (which happens on a clean environment so it contains all the setup)

Also make sure you have pnpm 9 (current) and Node 22 (current) installed.

imranbarbhuiya:
we should also need to skip generating the id when no middleware is registered

[...]

PR #1719, opened 2024-06-21
Fix 0.10.0 errors

### LINKED ISSUE #1715: Upgrading to 0.10.0 results in Failed to execute 'fetch' on 'Window'
**Description**

Upgraded to 0.10.0 (and 7.0.0 of openapi-typescript). Perform `.POST(..)` method and receive:

```
Failed to execute 'fetch' on 'Window': The `duplex` member must be specified for a request with a streaming body

at coreFetch (openapi-fetch.js?v=c62ec3c6:100:26)
```

**Reproduction**

As above, previously working POST request on 0.9.0, upgrade to latest 0.10.0. Execute same POST request

**Expected result**

To not throw.

**Checklist**

drwpow:
## Changes

Fixes #1715 and #1712. Adds Playwright tests to catch client behavior that doesn’t exist in Node’s `fetch()` implementation.

## How to Review

- Tests added; tests should pass

## Checklist

- [x] Unit tests updated


packages/openapi-fetch/src/index.js
```diff
@@ -84,56 +90,63 @@ export default function createClient(clientOptions) {
...
-          id,
-        });
-        if (result) {
-          if (!(result instanceof Request)) {
-            throw new Error("onRequest: must return new Request() when modifying the request");
+    if (middlewares.length) {
+      id = randomID();
+
+      // middleware (request)
+      options = Object.freeze({
```
drwpow, line 101: Micro-optimization: don’t run certain codepaths if no middleware loaded

packages/openapi-fetch/test/fixtures/e2e/app.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { paths } from "./e2e.d.ts";
+
+const client = createClient<paths>({
+  baseUrl: "/api/v1",
+});
+
+/**
+ * Test 1: GET /api/v1/get
+ */
+async function testGet() {
```
drwpow, line 11: New tests that get loaded in a true browser environment. Not comprehensive, but we at least have:

- GET request (no body)
- POST request (with body)
- POST request with multi-part/form data

packages/openapi-fetch/package.json
```diff
@@ -54,15 +54,17 @@
     "build:cjs": "esbuild --bundle src/index.js --format=cjs --outfile=dist/cjs/index.cjs && cp dist/index.d.ts dist/cjs/index.d.cts",
     "format": "biome format . --write",
     "lint": "biome check .",
-    "generate-types": "openapi-typescript test/fixtures/api.yaml -o test/fixtures/api.d.ts",
+    "generate-types": "openapi-typescript -c test/redocly.yaml",
```
drwpow, line 57: Use fancy new 7.x ability to generate multiple schemas with one config file

packages/openapi-fetch/src/index.js
```diff
@@ -84,56 +90,63 @@ export default function createClient(clientOptions) {
...
+            request = result;
           }
-          request = result;
         }
       }
     }
 
     // fetch!
-    let response = await fetch(request.url, request);
+    let response = await fetch(request);
```
drwpow, line 128: This is the central error fix—`Request` can’t be substituted for `RequestInit` in every environment. Trying to hack around this results in even nastier code when you consider different browsers implement different stages of `RequestInit` and experimental APIs under-the-hood. This library shouldn’t take on any of that burden, and instead use a `Request` object which is universal and spec-compliant.

packages/openapi-fetch/src/index.d.ts
```diff
@@ -15,7 +15,7 @@ export interface ClientOptions extends Omit<RequestInit, "headers"> {
   /** set the common root URL for all API requests */
   baseUrl?: string;
   /** custom fetch (defaults to globalThis.fetch) */
-  fetch?: (input: string, init?: RequestInit) => Promise<Response>;
+  fetch?: (input: Request) => Promise<Response>;
```
drwpow, line 18: This is _technically_ a breaking change, but since it was a breaking change from `0.9.0` I’m just going to pretend this was an accidental regression and never happened 🙈

package.json
```diff
@@ -12,14 +12,18 @@
     "lint": "pnpm run -r --parallel --aggregate-output lint",
     "format": "pnpm run -r --parallel --aggregate-output format",
     "test": "pnpm run -r --parallel --aggregate-output test",
+    "test-e2e": "pnpm run -r --parallel --aggregate-output test-e2e",
```
drwpow, line 15: ⚠️ **VERY** important that this is not run as part of `"test"`. We don’t need to run this multiple times in different OSs and and different Node versions. We only need to run this **ONCE** in a single environment (Ubuntu, Node latest because that’s the cheapest/fastest).

PR #1733, opened 2024-06-26
Remove sourceRoot configuration to enable proper source file mapping

phk422:
Hi, @drwpow! I've encountered an issue where VSCode's debugger cannot correctly navigate to the source files.
image

## Changes

- Removed `sourceRoot` configuration from `tsconfig.json` .
- This fix resolves the issue where source maps were not correctly mapping to the original TypeScript files.


drwpow (approved):
Ah good catch; I think this isn’t intentional

PR #1746, opened 2024-07-05
fix: Correct handling of default parameter values in referenced component schema

### LINKED ISSUE #1741: bug: request body fields with default values are treated as required
**Description**

We have generated TS interfaces using our OpenAPI 3.1.0 specification. There are some parameters that have default values (so they are not required in request, and the body without them will be still valid):

image

However, these parameters are treated as required when generating paths:

image

| `openapi-typescript` | `7.0.1`                      |
| Node.js              | `20.13.1`                      |
| OS + version         | `macOS Sonoma 14.4` |

**Reproduction**

Try to use this openapi schema:

```json
"src__schemas__social_profiles__Create": {
  "properties": {
    "card_title": {
      "type": "string",
      "maxLength": 64,
      "minLength": 1,
      "title": "Card Title",
      "default": "Social Profile"
    },
    "template": {
      "type": "string",
      "maxLength": 64,
      "minLength": 1,
      "title": "Template"
    },
    "photo": {
      "anyOf": [
        {
          "type": "string",
          "maxLength": 2083,
          "minLength": 1,
          "format": "uri"
        },
        {
          "const": ""
        }
      ],
      "title": "Photo",
      "default": ""
    },
    "signature": {
      "anyOf": [
        {
          "type": "string",
          "maxLength": 2083,
          "minLength": 1,
          "format": "uri"
        },
        {
          "const": ""
        }
      ],
      "title": "Signature",
      "default": ""
    },
    "design": {
      "type": "object",
      "title": "Design",
      "default": {

      }
    },
    "details": {
      "type": "object",
      "title": "Details",
      "default": {

      }
    },
    "socials": {
      "type": "object",
      "title": "Socials",
      "default": {

      }
    },
    "meta": {
      "type": "object",
      "title": "Meta",
      "default": {

      }
    }
  },
  "type": "object",
  "required": [
    "template"
  ],
  "title": "Create",
  "description": "Social Profile schema for create."
}
```

You can see that default values are set, but the fields are not recognized as not required.

**Expected result**

Parameters that have default values have `| undefined` in type declaration, like
```ts
type SocialProfileCreate = {
    card_title: string;
    template: string;
    photo?: string | "";
    signature?: string | "";
    design?: Record<string, never>;
    details?: Record<string, never>;
    socials?: Record<string, never>;
    meta?: Record<string, never>;
}
```

**Checklist**

phk422:
fix: Correct handling of default parameter values in referenced component schema

## Changes
Fix #1741 

## How to Review

Tests added; tests should pass

## Checklist

- [x] Unit tests updated


phk422:
Ah, I usually use eslint + prettier for code styling, and I haven't used biome to enforce code style. There seems to be a bug with the biome plugin on my VSCode, causing multiple failures in code style checks.😭

phk422:
> Ah! I usually use eslint + prettier for code styling, and I haven't used biome to enforce code style. There seems to be a bug with the biome plugin on my VSCode, causing multiple failures in code style checks.😂

duncanbeevers:
I think this change is insufficient since `components/schemas` may still be referenced within request bodies.

Unless a field is explicitly marked `required` it should not need to be present in the request body; in such an instance the `default` value is only serving as a hint that a value _may_ be supplied by the server.

Consider this simple schema where a single component schema is used as both the request body and as the response.
* a buyer *must* specify a shirt color
* a buyer *may* specify a shirt quantity

Since only a single type definition will be created for `ShirtPurchase`, `quantity` must remain an optional property, _even though_ it will be present in the response body.

```yaml
paths:
  /purchase:
    post:
      requestBody:
        required: true
        content:
          'application/json':
            schema:
              $ref: '#/components/schemas/ShirtPurchase'
      responses:
        '200':
          description: OK
          content:
            'application/json':
              schema:
                $ref: '#/components/schemas/ShirtPurchase'
components:
  schemas:
    ShirtPurchase:
      type: object
      properties:
        color:
          type: string
          enum:
            - blue
            - yellow
        quantity:
          type: number
          minimum: 1
          default: 1
    required:
      - color
```

drwpow:
> Ah, I usually use eslint + prettier for code styling, and I haven't used biome to enforce code style. There seems to be a bug with the biome plugin on my VSCode, causing multiple failures in code style checks.😭

Ah sorry—yeah that was kind of a silent change. But I’ve been blown away with the bugfixes Biome had in lots of ESLint packages, and the speed is hard to beat. I’m actually neutral on Prettier vs Biome for formatting; that was more an experiment.

the `.vscode/settings.json` should be responsible for setting formatters, and `.vscode/extensions.json` should prompt you to install the correct extensions, but LMK if that’s not working as expected. I also added `format` commands to update formatting easily.

As I type all this out, this is probably all good info that I should put into the CONTRIBUTING.md guide 🙂

phk422:
> > Ah, I usually use eslint + prettier for code styling, and I haven't used biome to enforce code style. There seems to be a bug with the biome plugin on my VSCode, causing multiple failures in code style checks.😭
> 
> Ah sorry—yeah that was kind of a silent change. But I’ve been blown away with the bugfixes Biome had in lots of ESLint packages, and the speed is hard to beat. Plus the team is really collaborative and I’m finding contributing to Biome easier than ESLint! But formatting-wise I’m neutral on Prettier vs Biome for formatting; that was more an experiment.
> 
> the `.vscode/settings.json` should be responsible for setting formatters, and `.vscode/extensions.json` should prompt you to install the correct extensions, but LMK if that’s not working as expected. I also added `format` commands to update formatting easily.
> 
> As I type all this out, this is probably all good info that I should put into the CONTRIBUTING.md guide 🙂

Alright, I think I should learn more about Biome too!😊

phk422:
> I think this change is insufficient since `components/schemas` may still be referenced within request bodies.
> 
> Unless a field is explicitly marked `required` it should not need to be present in the request body; in such an instance the `default` value is only serving as a hint that a value _may_ be supplied by the server.
> 
> Consider this simple schema where a single component schema is used as both the request body and as the response.
> 
> * a buyer _must_ specify a shirt color
> * a buyer _may_ specify a shirt quantity
> 
> Since only a single type definition will be created for `ShirtPurchase`, `quantity` must remain an optional property, _even though_ it will be present in the response body.
> 
> ```yaml
> paths:
>   /purchase:
>     post:
>       requestBody:
>         required: true
>         content:
>           'application/json':
>             schema:
>               $ref: '#/components/schemas/ShirtPurchase'
>       responses:
>         '200':
>           description: OK
>           content:
>             'application/json':
>               schema:
>                 $ref: '#/components/schemas/ShirtPurchase'
> components:
>   schemas:
>     ShirtPurchase:
>       type: object
>       properties:
>         color:
>           type: string
>           enum:
>             - blue
>             - yellow
>         quantity:
>           type: number
>           minimum: 1
>           default: 1
>       required:
>         - color
> ```

Oh! It seems I did overlook this situation. I'm curious, what approach do you have for handling this kind of situation? @duncanbeevers @drwpow

duncanbeevers:
> what approach do you have for handling this kind of situation?

My suggestion is to not use the presence of `default` to determine whether a property should be optional, and *only* rely on a property's presence in `required`.

drwpow:
> I'm curious, what approach do you have for handling this kind of situation? @duncanbeevers @drwpow

+1 to @duncanbeevers’ comment. For fixing this, I usually start by creating a failing test (TDD), and [the one they supplied](https://github.com/openapi-ts/openapi-typescript/pull/1746#issuecomment-2210813947) seems sufficient!

From there, tracing/resolving nodes in 7.x is one of the major internal reasons for rewriting! In previous versions it was much harder to peer into nodes. But in the current version, `ctx.resolve($ref)` lets you look up any `$ref` to see its contents.

It does add a little bit of work (especially considering nested `$ref`s), but unless I’m misunderstanding you should only need **that top-level object** so I don’t _think_ it should be complicated (i.e. you don’t have to deeply-traverse all trees from the `requestBody`; just peer into that top-level `$ref`).

phk422:
I noticed that @duncanbeevers' proposed approach conflicts with #1681, so maybe we should consider again here. @drwpow @duncanbeevers

duncanbeevers:
> I noticed that @duncanbeevers' proposed approach conflicts with https://github.com/openapi-ts/openapi-typescript/pull/1681, so maybe we should consider again here.

That doesn't seem like a conflict. My proposal is to make these fields optional, which seems to align with what is happening in that PR as well. I'm happy to stand corrected if I'm misunderstanding.

phk422:
> > what approach do you have for handling this kind of situation?
> 
> My suggestion is to not use the presence of `default` to determine whether a property should be optional, and _only_ rely on a property's presence in `required`.

I discovered that this default behavior can be controlled through [defaultNonNullable](](https://github.com/openapi-ts/openapi-typescript/blob/69f0628b5ad22c3535c43653b70bd8fbc5f3887d/packages/openapi-typescript/src/types.ts#L635)). By setting it to false, the scenario mentioned in this PR can be accommodated. It seems designed this way to meet the needs of different users. If I am misunderstanding, please correct me.

phk422:
> > > what approach do you have for handling this kind of situation?
> > 
> > 
> > My suggestion is to not use the presence of `default` to determine whether a property should be optional, and _only_ rely on a property's presence in `required`.
> 
> I discovered that this default behavior can be controlled through [defaultNonNullable](https://github.com/openapi-ts/openapi-typescript/blob/69f0628b5ad22c3535c43653b70bd8fbc5f3887d/packages/openapi-typescript/src/types.ts#L635). By setting it to false, the scenario mentioned in this PR can be accommodated. It seems designed this way to meet the needs of different users.Or do we need to handle the requestBody separately? If I am misunderstanding, please correct me.

duncanbeevers:
@phk422 Thanks for calling out that `defaultNonNullable` option; I wasn't familiar with it.
Setting `defaultNonNullable: false` cleared up all the problems in our codebase around request bodies with default values.

I *think* that makes this PR the correct approach since it detects these `default` definitions in request bodies, and needn't concern itself with component schemas.

phk422:
Above, I think this pr can be merged,@drwpow, what do you think?

drwpow:
> Above, I think this pr can be merged,@drwpow, what do you think?

I think so! If you could add a `patch` changeset please (see Changeset comment above) I’d be happy to approve and merge! Thank you @phk422 and @duncanbeevers for discussing and digging into this further!

drwpow (approved):

duncanbeevers:
@Nikoms When I run the types-generator on this OpenAPI document i see the expected behavior.

* Running with `--default-non-nullable` results in a non-optional `priority: number`
* Running with no flags results in an optional `priority?: number`

```
$ npx openapi-typescript ./api.yaml | grep 'priority[?|:]'
      priority?: number;
      priority?: number;
$ npx openapi-typescript ./api.yaml --default-non-nullable | grep 'priority[?|:]'
      priority: number;
      priority: number;
```

Nikoms:
Ok then, I must have a problem with my computer or something, because I have this:

```
npx openapi-typescript https://docs-be.here.com/bundle/tour-planning-api-api-reference/page/api.yaml | grep 'priority[?:]'
            priority: number;
            priority: number;
```

philios33:
I can confirm this is still a problem.

```
npx openapi-typescript https://docs-be.here.com/bundle/tour-planning-api-api-reference/page/api.yaml | grep 'priority[?:]'
            priority: number;
            priority: number;
npx openapi-typescript https://docs-be.here.com/bundle/tour-planning-api-api-reference/page/api.yaml --default-non-nullable | grep 'priority[?:]'
            priority: number;
            priority: number;
npx openapi-typescript https://docs-be.here.com/bundle/tour-planning-api-api-reference/page/api.yaml --default-non-nullable false | grep 'priority[?:]'
            priority?: number;
            priority?: number;

```
My guess is that the option (which should default to false) somehow gets overridden somewhere to be true once a $ref is followed, and then explicitly setting the flag to be false in the command seems to force it to remain false properly.

philios33:
This option is now defaulting to true in v7, sorry I didn't realise that.  I'm guessing @duncanbeevers is on v6.

PR #1759, opened 2024-07-12
fix: export default exports correctly

phk422:
## Changes

When I was using the api, I found that some of the default exports inside were unusable

## How to Review

The default export method can be used correctly

## Checklist

- [ ] Unit tests updated


drwpow (approved):
👍 Good fix!

PR #1762, opened 2024-07-15
feat: allow form data through default body serializer

thatsprettyfaroutman:
I was trying to use openapi-fetch for file uploading, but by default that wasn't possible (using `multipart/form-data`). Looking at the source code I noticed that part of the solution for fixing it seemed to be missing as `defaultBodySerializer` never allows `FormData` to be passed through it:
```js
if (requestInit.body) {
  // Here, bodySerializer always returns a JSON string by default
  requestInit.body = bodySerializer(requestInit.body);
  // remove `Content-Type` if serialized body is FormData; browser will correctly set Content-Type & boundary expression
  if (requestInit.body instanceof FormData) {
    requestInit.headers.delete("Content-Type");
  }
}
```

I'm trying to fix that in this PR by allowing `FormData` to be passed through the `defaultBodySerializer` enabling file uploads with `multipart/form-data` by default.

## Changes

Allow `FormData` to be passed through `defaultBodySerializer` as `body` to enable file uploading by default.

## How to Review

Upload a file as part of `multipart/form-data`

## Checklist

- [ ] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/test/index.test.ts
+++ b/packages/openapi-fetch/test/index.test.ts
@@ -1352,6 +1352,33 @@ describe("client", () => {
         customProperty: "value",
       });
     });
+
+    it('multipart/form-data with a file', async () => {
+      const TEST_STRING = 'Hello this is text file string';
+      
+      const file = new Blob([TEST_STRING], { type: 'text/plain' });
+      const formData = new FormData();
+      formData.append('file', file);
+
+      const client = createClient<paths>({ baseUrl });
+       useMockRequestHandler({
+        baseUrl,
+        method: "post",
+        path: "/multipart-form-data-file-upload",
+        handler: async (data) =>{
+          // Get text from file and send it back
+          const formData = await data.request.formData();
+          const text = await (formData.get('file') as File).text()
+          return new HttpResponse(JSON.stringify({text}))
+        },
+      });
+      const {data} = await client.POST("/multipart-form-data-file-upload", {
+        // @ts-ignore // TODO: how to get this to accept FormData? 
+        body: formData,
+      })
+
+      expect(data?.text).toBe(TEST_STRING)
+    })
   });
 
   describe("responses", () => {
--- a/packages/openapi-fetch/test/index.test.ts
+++ b/packages/openapi-fetch/test/index.test.ts
@@ -1355,7 +1355,7 @@ describe("client", () => {
 
     it('multipart/form-data with a file', async () => {
       const TEST_STRING = 'Hello this is text file string';
-      
+
       const file = new Blob([TEST_STRING], { type: 'text/plain' });
       const formData = new FormData();
       formData.append('file', file);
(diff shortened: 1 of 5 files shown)


drwpow:
Thank you for this fix! I’d accept this change if you also **added a test** that was failing before your fix, and passes with.

Also be sure to add a **Changeset** for this that adds a `patch` to `openapi-fetch` so it can be released on merge!

packages/openapi-fetch/test/index.test.ts
```diff
@@ -1352,6 +1352,33 @@ describe("client", () => {
...
+        handler: async (data) => {
+          // Get text from file and send it back
+          const formData = await data.request.formData();
+          const text = await (formData.get("file") as File).text();
+          return new HttpResponse(JSON.stringify({ text }));
+        },
+      });
+      const { data } = await client.POST("/multipart-form-data-file-upload", {
+        // TODO: how to get this to accept FormData?
+        body: formData as unknown as string,
```
thatsprettyfaroutman, line 1377: How to get this to accept FormData?

thatsprettyfaroutman:
Thanks @drwpow! I added the test and a changeset entry. I'm having a small problem with typings though as the request body is always expected to be a `string` and I'm trying to send `FormData`.  I'm wondering how to tackle that, since I couldn't find the part where the body type is defined.

packages/openapi-fetch/test/index.test.ts
drwpow: Oh hm do we need to adjust the types here? Is there an improvement we could make in `packages/openapi-fetch/index.d.ts` for `body`?

drwpow:
Looks good! But I see we have a type error in the tests we should probably fix in `openapi-fetch/index.d.ts`.

packages/openapi-fetch/test/index.test.ts
drwpow: Oh this will be tricky. This library’s whole purpose is to assert types for `body` based on your OpenAPI schema. https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-fetch/src/index.d.ts#L90

We don’t want to _always_ allow FormData as that would mess up some people’s code. So we want to conditionally allow FormData in some instances where it’s possible (e.g. not for GET). So we’ll need to surgically add FormData as a union into request bodies, but only when it may be possible/desired.

Given the complexity, I’m OK with merging this PR without that just to unblock the runtime, and the types can be tackled separately (sidenote: that’s why the runtime and types are separated—the type complexity isn’t possible to express with normal runtime code)

drwpow: What are your thoughts? Do you want to try and solve the type errors? Or merge by requiring manual overrides?

thatsprettyfaroutman: I'm ok with merging now. I think letting FormData through the body serializer is a nice little addition in itself. Oh, I was wondering why the types are separated, that's very cool to know 👍

Issue #1651, opened 2024-05-01
`--properties-required-by-default` and `--default-non-nullable` flags do not work


cekpowell:
**Description**

The `--properties-required-by-default` and `--default-non-nullable` flags do not work. If you include these flag in the CLI, the execution hangs forever. From [this](https://github.com/drwpow/openapi-typescript/issues/1003) issue it also seems like the same thing is happening with `--path-params-as-types`. If you remove these flags, the execution works fine.

| Name                 | Version                      |
| :------------------- | :--------------------------- |
| `openapi-typescript` | `6.7.5`                      |
| Node.js              | `18.18.2`                      |
| OS + version         | `macOS 14.0` |

**Reproduction**

Run any generation command with `--properties-required-by-default`, `--default-non-nullable` or `--path-params-as-types` flags, and you will see the execution hangs.

**Checklist**

- [x] My OpenAPI schema passes the [Redocly validator](https://redocly.com/docs/cli/commands/lint/) (`npx @redocly/cli@latest lint`)
- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

drwpow:
Is anyone able to provide a reproduction of this issue? We have [tests](https://github.com/openapi-ts/openapi-typescript/blob/ff93eac96d424f6f901535b1f3645187d56beb59/packages/openapi-typescript/test/transform/schema-object/object.test.ts#L341) for [both](https://github.com/openapi-ts/openapi-typescript/blob/ff93eac96d424f6f901535b1f3645187d56beb59/packages/openapi-typescript/test/transform/schema-object/object.test.ts#L134) that test they’re working that are both passing.

rumyanar:
In 7.0.1 I'm experiencing a related issue - the option -properties-required-by-default has no effect. It doesn't hang the CLI as in the original comment.
I'm also experiencing https://github.com/openapi-ts/openapi-typescript/issues/1467 even the required fields are optional
```
 "DeviceDto": {
        "type": "object",
        "properties": {
          "deviceId": {
            "type": "string",
            "format": "uuid",
            "required": true
          },
          "name": {
            "type": "string",
          },
```

outputs:
```
DeviceDto: {
            /** Format: uuid */
            deviceId?: string;
            name?: string;
}
```

magnattic:
[...]

PR #1764, opened 2024-07-16
fix: --properties-required-by-default flag

michalfedyna:
## Changes

Fix --properties-required-by-default flag not working. #1651 

## How to Review

It was just simple typo.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/eight-pants-attack.md
+++ b/.changeset/eight-pants-attack.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Fix --properties-required-by-default flag not working
--- a/packages/openapi-typescript/bin/cli.js
+++ b/packages/openapi-typescript/bin/cli.js
@@ -101,7 +101,7 @@ async function generateSchema(schema, { redocly, silent = false }) {
       alphabetize: flags.alphabetize,
       arrayLength: flags.arrayLength,
       contentNever: flags.contentNever,
-      propertiesRequired: flags.propertiesRequired,
+      propertiesRequiredByDefault: flags.propertiesRequiredByDefault,
       defaultNonNullable: flags.defaultNonNullable,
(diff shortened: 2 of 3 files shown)


michalfedyna:
@drwpow any chances of fast approve and release?

drwpow:
Thanks for the fix; this looks good to me! I’d love to add a **test** for this so this doesn’t regress in the future. And also please add a `patch` changeset (see comment) so the version can be bumped and released. But otherwise, this looks good and would be happy to merge when both of those items are in.

michalfedyna:
@drwpow Changeset done. It will be hard to test that part. The problem comes from cli.js not being part of the build process and static analysis not being done on that file. I will take some time over the weekend and make it work if you wish. But for now is it possible to merge that change? It will be helpful for me at work.

drwpow:
So [there are CLI tests for flags already](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/test/cli.test.ts#L98). A lot of them had been moved into the Node API tests, but for this change, we could just do a simple `expect(…).toMatchFileSnapshot()` test to make sure it’s working as-expected. It would also just help prevent these regressions in the future!

michalfedyna:
@drwpow Should be done! You can run `diff ./examples/github-api.ts ./examples/github-api-required.ts` to check if flag works as you expect.

michalfedyna:
And I'm testing against `./example/github-api-required.ts`

drwpow:
Awesome, thank you so much! That works—just something that verifies a) this is the correct fix, and b) we’ll see a test failure if this regresses. That saves a ton of work in the long term; appreciate it!

drwpow (approved):

michalfedyna:
Great, thanks!

Issue #1754, opened 2024-07-08
Inject option dropped in v7


johnnymetz:
**Description**

Version 6 had an `inject` options, which allowed you to import arbitrary TS types at the beginning of the file, see docs: https://openapi-ts.pages.dev/6.x/node#options

This option no longer exists but I need it. Is there still a way to do this or has this support been dropped completely?

**Proposal**

_Outline the change to the library. If this is for the CLI, propose a flag name and/or workflow. If this is for Node API, propose how this would be consumed. Reference any prior art or similar examples._

Re-implement the `inject` option, see code: https://github.com/openapi-ts/openapi-typescript/pull/1363/files#diff-aa2bf9eceec4d8c746802aaaeaa09f72333c8a8983ef9e956fdc3c71c972bda7L117-L118

**Checklist**

- [ ] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

kerwanp:
Hey @johnnymetz, the feature has been recently released (thanks to @phk422 #1766 !). The documentation: https://openapi-ts.dev/node#options

I'm closing the issue has this should be resolved, feel free to open a new one if necessary!

PR #1766, opened 2024-07-17
feat: Add the inject option

phk422:
## Changes

#1754 - Add the inject option

## How to Review

All the tests should pass

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/little-pets-perform.md
+++ b/.changeset/little-pets-perform.md
@@ -1,0 +1,6 @@
+---
+"openapi-typescript-docs": minor
+"openapi-typescript": minor
+---
+
+feat: Add the inject option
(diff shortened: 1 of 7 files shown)


.changeset/little-pets-perform.md
```diff
@@ -0,0 +1,6 @@
+---
+"openapi-typescript-docs": minor
```
drwpow, line 2: Just a note: We don’t need to version the docs; they always get released on merge to `main`. Should make sure these are ignored from changesets.

drwpow (approved):
This feature makes sense, and I think would be useful to many people. Thank you!

PR #1772, opened 2024-07-19
fix: add explicit import of performance to fix bazel compatibility issues

ashwin153:
## Changes

- Running openapi-typescript through Bazel results in a `ReferenceError: performance is not defined` error
- This PR adds an explicit `import { performance } from "node:perf_hooks"` whenever `performance` is used in the code.
- I've tested that this fixes my Bazel compability issues.

## How to Review

- Make sure I didn't miss any usages of `performance`.
- Let me know if there's a reason why the import was omitted.

## Checklist

- [ ] Unit tests updated

### DIFF
--- a/.changeset/chilled-radios-remain.md
+++ b/.changeset/chilled-radios-remain.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+fix: add explicit import of performance to fix bazel compatibility issues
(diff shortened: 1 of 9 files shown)


drwpow (approved):
Thank you for the fix! While normally I’d recommend raising this upstream with Bazel, this is unopinionated enough it’s safe to merge.

If you add a `patch` changeset we can release it on merge!

drwpow:
Looks great, thank you!

PR #1782, opened 2024-07-24
Improve querySerializer

petr-like:
## Changes

Added filtering after forming query parameters in querySerialize.

## How to Review

Currently, if there is any key in queryParams with an empty object, the result of the generation will include a double ampersand.

For example:
```javascript
{
      a: 1,
      b: {},
      c: 3,
}
```
will return => example.com?a=1&&c=3
Because Serialize creates an array with an empty string.

After the improvement, it will be example.com?a=1&c=3

## Checklist

- [ ] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -382,7 +382,7 @@ export function createQuerySerializer(options) {
         search.push(serializePrimitiveParam(name, value, options));
       }
     }
-    return search.join("&");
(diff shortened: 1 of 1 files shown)


drwpow (changes requested):
Thank you for opening this! But we’ll need to add some test cases ensuring that this fixes problems. I see that it doesn’t fail any tests, but we’ll need to add some too.

Issue #1781, opened 2024-07-24
Problem using --array-length when minItems equals to maxItems


Couc:
Hello everyone :)
**Description**

When using --array-length flag and minItems, maxItems with identical values in schema definition, the generated interface is [].

| Name                 | Version                      |
| :------------------- | :--------------------------- |
| `openapi-typescript` | `>=7.0.0`                      |
| Node.js              | `20.15.0`                      |
| OS + version         | `Windows 10` |

**Reproduction**

Run openapi-typescript with --array-length flag set to true on yaml :
```yaml
openapi: 3.1.0
info:
  title: Webhook Example
  summary: My lovely API
  version: 1.0.0
  description: some description
  license:
    name: MIT
    identifier: MIT
    url: https://opensource.org/licenses/MIT
jsonSchemaDialect: https://spec.openapis.org/oas/3.1/dialect/base
tags:
  - name: Gameplay
    description: gameplay tag
servers:
  - url: https://redocly-example.com/{var}
    variables:
      var:
        enum:
          - a
          - b
        default: b
security: []
paths:
  /pets/items:
    get:
      operationId: getTest
      summary: Get Test! Foo! Etc!
      responses:
        200:
          description: '200'
          content:
            application/json:
              schema:
                type: array
                items:
                  type: string
                minItems: 3
                maxItems: 3
        400:
          description: An error response
```

**Expected Result**
The part "application/json" should be typed as "[string, string, string]" in : 
```ts
responses: {
            /** @description 200 */
            200: {
                headers: {
                    [name: string]: unknown;
                };
                content: {
                    "application/json": [string, string, string];
                };
            };
```
**Current Result**
```ts
responses: {
            /** @description 200 */
            200: {
                headers: {
                    [name: string]: unknown;
                };
                content: {
                    "application/json": [];
                };
            };
```
**Checklist**

- [X] My OpenAPI schema passes the [Redocly validator](https://redocly.com/docs/cli/commands/lint/) (`npx @redocly/cli@latest lint`)
- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

PartyWumpus:
I've got the same issue.

PR #1784, opened 2024-07-26
Fix: Correct handling of identical minItems and maxItems in array schemas when arrayLength option is true

yoshi2no:
close: #1781

## Changes
- Corrects handling of array schemas with identical `minItems` and `maxItems` when arrayLength option is true
- Adds test case

## How to Review
The added test should pass.

## Checklist

- [x] Unit tests updated


kerwanp (approved):
Looks good to me :rocket:

drwpow (approved):
Thank you!

PR #1792, opened 2024-07-31
style: fix formatting after CI commit

gzm0:
Seems like 8e5fa3a24cf217b730d35a7f825ba0509856a1bb broke the formatting.

### DIFF
--- a/packages/openapi-typescript-helpers/package.json
+++ b/packages/openapi-typescript-helpers/package.json
@@ -23,12 +23,7 @@
     },
     "./*": "./*"
   },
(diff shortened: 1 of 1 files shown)


drwpow:
Thanks for fixing, but we just need to ignore JSON from linting/formatting. Otherwise npm/pnpm and Biome will be fighting over the formatting of these files

drwpow:
Fixed in #1797; just ignored from Biome. Whenever `pnpm` touches this it’ll reformat it anyway; no sense in linting.

PR #1795, opened 2024-08-01
Automatically stale and close issues

kerwanp:
## Introduction

This is a pull request as well as an enhancement proposal for the repository.

There are issues created in 2021 that are still opened, some of them might be relevant but not enough to receive any updates. To avoid accumulating issues, I think it would be interesting to automatically stale and close issues that does not receive any update.

## Changes

This pull requests adds a new automated workflow to automatically mark issues as stale and close them using the [Stale Github Action](https://github.com/actions/stale).

It works as the following:

1. After 90 days without activity, the issue is marked as stale with the following message

`"This issue is stale because it has been open for 90 days with no activity. If there is no activity in the next 7 days, the issue will be closed."`

2. After 97 days without activity, the issue is closed with the following message

`"This issue was closed because it has been inactive for 7 days since being marked as stale. Please open a new issue if you believe you are encountering a related problem."`

> It is possible to label issues with `Issue accepted` to ensure that they will never be closed automatically.

### DIFF
--- a/.github/workflows/issue-stale.yml
+++ b/.github/workflows/issue-stale.yml
@@ -1,0 +1,23 @@
+name: Close inactive issues
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "30 1 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-issue-message: "This issue is stale because it has been open for 90 days with no activity. If there is no activity in the next 7 days, the issue will be closed."
+          close-issue-message: "This issue was closed because it has been inactive for 7 days since being marked as stale. Please open a new issue if you believe you are encountering a related problem."
+          ascending: false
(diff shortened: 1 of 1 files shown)


kerwanp:
@drwpow what do you think?

drwpow:
I’m in favor of this. I think we’re at a point now where the number of issues opened do need to get cleaned up. While sometimes we may close an issue that may still be present, I think it will help the **quality** of issues as the stale ones usually contain no reproductions and contain too much missing info for it to be a quick, actionable fix.

> It is possible to label issues with `Issue accepted` to ensure that they will never be closed automatically.

I really like this. There will definitely be some issues where they may go stale, not from lack of interest or missing info, but just because they are particularly challenging problems.

drwpow (approved):
LGTM! Worth trying.

With these GitHub Actions tweaks, it sometimes takes several tries (or at least, it does for me—you may be better than I am 😄). Either way I don’t mind approving followups/tweaks to this

PR #1810, opened 2024-08-02
test: remove unused nanoid import in index.bench.js

gzm0:
It is not declared in package.json, so running `npx vite bench` fails.

### DIFF
--- a/packages/openapi-fetch/test/index.bench.js
+++ b/packages/openapi-fetch/test/index.bench.js
@@ -2,7 +2,6 @@ import axios from "axios";
 import { http, HttpResponse } from "msw";
 import { setupServer } from "msw/node";
 import { Fetcher } from "openapi-typescript-fetch";
-import { nanoid } from "nanoid";
 import superagent from "superagent";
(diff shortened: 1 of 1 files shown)


drwpow:
Also a note—I’ve found the benchmarking to be wildly inconsistent 😓. It’s still in beta for Vitest, and I think we should have another alternative.

Or, put another way, the **absolute** numbers fluctuate wildly. But the **relative** measurements have some consistency (i.e. `openapi-fetch` being XX% faster than another package stays consistent-_ish_).

It’s been on my list to come up with more reliable benchmarks—would love your thoughts if there’s a better way to profile performance.

Also like I said before, anecdotally I haven’t had good experiences with performance regression tests (or, at least, not without an entire team at a company dedicated to maintaining them, and we don’t have that for this OSS project). CI (especially free CI) just fluctuates too widely with resources to be reliable. But I’m strongly in favor of manually-run benchmarking, that can be run when we want to see if something moves the needle up or down. And I think there’s improvements to be made in this area here, and I welcome your input! 🙏

Issue #1805, opened 2024-08-02
openapi-react-query: Add the ability to bring queryClient as argument


iffa:
Noticed that this wasn't possible yet and was a TODO in the code too.

kerwanp:
Temporarily, it is possible to wrap inside a `QueryClientProvider`. But in the future we should be able to pass the queryClient as an argument.

@tanstack/react-query
```tsx
const query = useQuery({
  queryKey: [],
  queryFn: () => {},
}, queryClient);
```

openapi-react-query
```tsx
const query = $api.useQuery('get', '/api/posts', {}, queryClient);
```

Feel free to open a pull request! This should be pretty straight forward.

kerwanp:
Closing this issue as this is resolved in #1814 thanks to @yoshi2no !

PR #1814, opened 2024-08-04
feat(openapi-react-query): Add the ability to bring queryClient as argument

yoshi2no:
ref: #1805

## Changes
Added the ability to pass `queryClient` as an argument in the `openapi-react-query`. 
This enhancement allows users to directly pass queryClient to the `useQuery`, `useMutation`, `useSuspenseQuery` hooks.

The sample code in [the issue](https://github.com/openapi-ts/openapi-typescript/issues/1805) is as follows:
```ts
const query = $api.useQuery('get', '/api/posts', {}, queryClient);
```

In my understanding, `$api.useQuery` currently accepts method, path, options, and queryOptions, and with the addition of queryClient, the correct format should be:
```ts
const query = $api.useQuery('get', '/api/posts', {}, {}, queryClient);
```

If I am mistaken in my understanding, please let me know. Your feedback would be greatly appreciated.

## How to Review
I don’t think it is necessary to add tests for this change, but what do you think?

## Checklist

- [x] Unit tests updated


kerwanp:
Hey! That looks really great, thanks a lot for your contrib!
Could you write some unit tests to ensure that we properly use the provided queryClient?

Then it will only needs the changesets and it will be ready to merge!

yoshi2no:
- Added test cases
  - The tests ensure that everything works correctly without using the `QueryClientProvider`. This is based on the approach used in react-query. 
    - [useQuery](https://github.com/TanStack/query/blob/1a1708bbfd7230a52d1fce92d768a38c6a33e22d/packages/react-query/src/__tests__/useQuery.test.tsx#L6023)
    - [useMutation](https://github.com/TanStack/query/blob/1a1708bbfd7230a52d1fce92d768a38c6a33e22d/packages/react-query/src/__tests__/useMutation.test.tsx#L1092)
- Added a minor changeset

kerwanp (approved):
Looks good to me! Really nice idea to check for react-query actual tests for that.

One more thing left, add yourself to the contributors list in `docs/scripts/update-contributors.js`

Issue #1633, opened 2024-04-26
Set baseUrl per request


RobinClowers:
**Description**

My app has an API that is per tenant, so I end up having to create a client every time I need to call the API, since the baseUrl varies.

**Proposal**

Ideally it would be as simple as:
```typescript
client.GET("/api/region", {
  baseUrl: "https://api.example.com",
})
```

**Checklist**

- [x] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

Gruak:
I also need this for the same reason.
It would also be nice to be able to set a baseUrl getter per client that would read the baseUrl from an external store on request or be able to set the baseUrl through a middleware.

drwpow:
This is an easy add and wouldn’t disrupt anything. Would welcome a PR for this 🙂

RobinClowers:
@Gruak I forgot about this, thanks for making that change! @drwpow can we get a release with the change soon?

PR #1817, opened 2024-08-04
feat(openapi-fetch): baseUrl per request

Gruak:
## Changes

ref: #1633

Now `baseUrl` can be set per request and it should take precedence over `baseUrl` client option:
```javascript
client.GET("/endpoint", {
  baseUrl: "https://api.example.com",
})
```

## How to Review

The tests should pass.

## Checklist

- [x] Unit tests updated


packages/openapi-fetch/test/index.test.ts
```diff
@@ -767,6 +767,29 @@ describe("client", () => {
       expect(getRequestUrl().href).toBe(toAbsoluteURL("/self"));
     });
 
+    it("baseUrl per request", async () => {
```
kerwanp, line 770: Is there a reason to have this in `describe('headers')`?

packages/openapi-fetch/test/index.test.ts
```diff
@@ -1272,6 +1295,33 @@ describe("client", () => {
         expect(req.headers.get("onFetch")).toBe("exists");
         expect(req.headers.get("onRequest")).toBe("exists");
       });
+
+      it("baseUrl can be overriden", async () => {
```
kerwanp, line 1299: It seems that this test is a duplicate of https://github.com/openapi-ts/openapi-typescript/pull/1817/files#diff-de37e4843cc1c72ff2441d5eed04f070dde7f4ad98e8b2c446bca1243c571d7dR770

The logic is the same for `{ getRequest }` https://github.com/openapi-ts/openapi-typescript/blob/231d5140a56a1922eb738bda6b42efe60d2af1ac/packages/openapi-fetch/test/fixtures/mock-server.ts#L78

.changeset/happy-singers-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+"openapi-fetch": minor
+---
+
+feat: set baseUrl per request
```
kerwanp, line 5: Could be a bit more explicit as this will be used in the changelog:
```suggestion
Allow specifying baseUrl per request
```

kerwanp (changes requested):
Great work!
Looks good to me, just two small comments. Happy to discuss about it if needed!

packages/openapi-fetch/test/index.test.ts
Gruak: Are you sure you're talking about the [second test](https://github.com/openapi-ts/openapi-typescript/pull/1817/files#diff-de37e4843cc1c72ff2441d5eed04f070dde7f4ad98e8b2c446bca1243c571d7dR1299) I wrote for the middleware? `{ getRequest }` is not even used there. It's purely a client-side test (doesn't really involve the mock server) which verifies that the overridden baseURL is correctly passed to the middleware and also checks that the trailing slash is removed.

.changeset/happy-singers-fry.md
Gruak: Sure, good point.

packages/openapi-fetch/test/index.test.ts
kerwanp: As the first test already ensure that the baseUrl is correct on the server (mock server) the second test is kind of a duplicate imo. But as the second one also test that specific parameters (in this case only baseUrl) are passed to the middleware before being used I'm ok to merge this as it is!

kerwanp (approved):

.changeset/happy-singers-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+"openapi-fetch": minor
```
drwpow, line 2: Just a note for next time: we’re using `minor` patches only for breaking changes while we’re on `0.x` on `openapi-fetch`. `patch` includes feature additions and bugfixes before `1.0`.

PR #1825, opened 2024-08-07
Allow falsy values (except undefined) as a valid body

goce-cz:
## Changes

Resolves https://github.com/openapi-ts/openapi-typescript/issues/1695

This PR ensures that all falsy values except `undefined` i.e. `null`, `false`, `0` and `""` are treated as a valid body. This means:
- going through `bodySerializer`, when specified
- sending the body to the server

## How to Review

- Check the added unit tests which describe the desired behavior.
- The change in implementation itself is ridiculously simple.

## Checklist

- [x] Unit tests updated
  - I didn't find any section describing the behavior. 

### DIFF
--- a/.changeset/rich-crews-repeat.md
+++ b/.changeset/rich-crews-repeat.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Allow falsy values (except undefined) as a valid body
--- a/packages/openapi-fetch/test/index.test.ts
+++ b/packages/openapi-fetch/test/index.test.ts
@@ -1,6 +1,8 @@
 import { HttpResponse, type StrictResponse } from "msw";
 import { afterAll, beforeAll, describe, expect, expectTypeOf, it } from "vitest";
 import createClient, {
+  type BodySerializer,
+  type FetchOptions,
   type MethodResponse,
   type Middleware,
   type MiddlewareCallbackParams,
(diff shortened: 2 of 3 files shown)


packages/openapi-fetch/test/index.test.ts
```diff
@@ -1275,6 +1277,186 @@ describe("client", () => {
...
+    const BODY_ACCEPTING_METHODS = [["PUT"], ["POST"], ["DELETE"], ["OPTIONS"], ["PATCH"]] as const;
+    const ALL_METHODS = [...BODY_ACCEPTING_METHODS, ["GET"], ["HEAD"]] as const;
+
+    const fireRequestAndGetBodyInformation = async (options: {
+      bodySerializer?: BodySerializer<unknown>;
+      method: (typeof ALL_METHODS)[number][number];
+      fetchOptions: FetchOptions<any>;
+    }) => {
+      const client = createClient<any>({ baseUrl, bodySerializer: options.bodySerializer });
+      const { getRequest } = useMockRequestHandler({
```
goce-cz, line 1290: I assume this is piling one mock over the other. I didn't find any "cleanup" procedure in the existing tests, so piling several more mocks on top 🙂

packages/openapi-fetch/test/index.test.ts
```diff
@@ -1275,6 +1277,186 @@ describe("client", () => {
...
+      });
+      await client[options.method]("/blogposts-optional", options.fetchOptions as any);
+
+      const request = getRequest();
+      const bodyText = await request.text();
+
+      return { bodyUsed: request.bodyUsed, bodyText };
+    };
+
+    it.each(ALL_METHODS)("missing body (with body serializer) - %s", async (method) => {
```
kerwanp, line 1304: I think we should definitely use more often `it.each` to have cleaner tests while testing more cases. But in this case isn't it overkill to test every method each time as the logic tested is only about falsy values?

What about a single test `it.each([{ body: undefined, except: false }, { body: 1, except: true }, ...])`

packages/openapi-fetch/test/index.test.ts
kerwanp: All good, the cleanup is right here! https://github.com/openapi-ts/openapi-typescript/blob/449d7e86369e8aa7788571ae67e00e6869e0d06f/packages/openapi-fetch/test/index.test.ts#L21

goce-cz: Ha, missed that. Thanks 👍

kerwanp:
Thanks for the contrib! I just added some comments regarding the tests.
I'm really not experienced in writing tests so I'll let @drwpow review this.

But imo, a unit test must test a single feature, in this case `falsy` bodies. Testing that `GET` and `POST` have different behavior deserve its own test (that we don't have).

packages/openapi-fetch/test/index.test.ts
```diff
@@ -1275,6 +1277,186 @@ describe("client", () => {
     });
   });
 
+  describe("body serialization", () => {
```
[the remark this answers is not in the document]
goce-cz: I originally had them there, but then saw that the whole test group is within "TypeScript checks" and these have nothing to do with TypeScript.

packages/openapi-fetch/test/index.test.ts
goce-cz: > But in this case isn't it overkill to test every method each time as the logic tested is only about falsy values?

There are differences in how GET and HEAD are used in some of the underlying implementation (not in this lib, hopefully) - some clients and servers disallow bodies for them. This includes the tools used in these tests. Having to think about these differences inspired me to rather have a comprehensive test of all methods. If you guys don't see it valuable, I'm OK with dropping it.


> What about a single test it.each([{ body: undefined, except: false }, { body: 1, except: true }, ...])

I usually prefer the assertions to be as explicit as possible, hence the explicit `expect().toBe()`, but I don't mind pulling the expectations into the test arguments, if that's what you guys are used to. Matter of taste really.

goce-cz:
> Testing that GET and POST have different behavior deserve its own test (that we don't have).

The fact that I'm running the tests over each method independently doesn't have anything to do with testing method behavior. It's just to ensure that the feature I am testing works consistently across all methods. See my reply to your comment above for more clarity.

packages/openapi-fetch/test/index.test.ts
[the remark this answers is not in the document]
kerwanp: You are right and it seems that most of the tests in this group are not even about types :thinking:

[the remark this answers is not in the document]
drwpow: That’s a good callout. We probably need to clean up our tests by moving TypeScript tests to actual [Vitest type tests](https://vitest.dev/guide/testing-types). This test file has also grown in size (which is a great problem to have) and can probably be broken up. All that to say, not too prescriptive about where this test goes; if it exists _somewhere_ I’m happy 🙂

drwpow (approved):
Thank you for the fix! These are great tests.

⚠️ We need a changeset to merge (please see comment)!

Non-blocking suggestion: agree with @kerwanp on some minor suggestions for test cleanup. We could deduplicate a little bit. Agree with your points @goce-cz about being explicit, but a lot of the tests have quite a lot of shared overlap, and very minor differences, that are easy to miss when modifying, and a few could be combined with some more clever iterating as you have

goce-cz:
I rebased and added the change-set. Hopefully correctly.

.changeset/rich-crews-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+"openapi-typescript": patch
```
drwpow, line 2: Whoops! Didn’t catch this was for the wrong package. Should be `openapi-fetch`.

goce-cz: Ohh... crap. Sorry. I'll issue a fix PR.

goce-cz: I see it has been already released. 😢

drwpow: Yours was released correctly! 🙂 https://github.com/openapi-ts/openapi-typescript/pull/1889

The `openapi-typescript` release was for an actual patch from another PR: https://github.com/openapi-ts/openapi-typescript/pull/1890

PR #1826, opened 2024-08-07
Do not set content-type on body-less requests

### LINKED ISSUE #1694: Empty body should not result in Content-Type header being set

goce-cz:
## Changes

Resolves #1694

**Important:** This PR touches the same code as #1825 and in-fact includes the same fix by itself. I recommend merging #1825 first.

This PR prevents Content-Type header from being set on requests with no body. This is required to ensure compatibility with strict server impls. such as [fastify](https://fastify.dev/).

## How to Review

- Consider checking the included tests and docs to understand the desired behavior.
- The change in the implementation is rather simple.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/docs/openapi-fetch/api.md
+++ b/docs/openapi-fetch/api.md
@@ -181,6 +181,18 @@ const { data, error } = await client.PUT("/submit", {
 });
 ```
 
+::: tip
+
+For convenience, `openapi-fetch` sets `Content-Type` to `application/json` automatically
+for any request that provides value for the `body` parameter. When the `bodySerializer` returns an instance of `FormData`,
+`Content-Type` is omitted, allowing the browser to set it automatically with the correct message part boundary.
+
+You can also set `Content-Type` manually through `headers` object either in the fetch options,
+or when instantiating the client. Setting `Content-Type` to `null` will omit the header, however the native fetch API
+will likely set it to `text/plain` in this case.
+
+:::
+
 ## Path serialization
 
 openapi-fetch supports path serialization as [outlined in the 3.1 spec](https://swagger.io/docs/specification/serialization/#path). This happens automatically, based on the specific format in your OpenAPI schema:
--- a/docs/openapi-fetch/api.md
+++ b/docs/openapi-fetch/api.md
@@ -188,8 +188,7 @@ for any request that provides value for the `body` parameter. When the `bodySeri
 `Content-Type` is omitted, allowing the browser to set it automatically with the correct message part boundary.
 
 You can also set `Content-Type` manually through `headers` object either in the fetch options,
-or when instantiating the client. Setting `Content-Type` to `null` will omit the header, however the native fetch API
-will likely set it to `text/plain` in this case.
+or when instantiating the client.
 
 :::
 
--- a/packages/openapi-fetch/test/index.test.ts
+++ b/packages/openapi-fetch/test/index.test.ts
@@ -981,9 +981,10 @@ describe("client", () => {
             },
           });
           // the fetch implementation won't allow sending a body without content-type,
-          // so it invents one up and sends it, hopefully this will be consistent across
-          // local environments and won't make the tests flaky
-          expect(contentType).toBe("text/plain;charset=UTF-8");
+          // and it defaults to `text/plain;charset=UTF-8`, however the actual default value
+          // is irrelevant and might be flaky across different fetch implementations
+          // for us, it's important that it's not `application/json`
+          expect(contentType).not.toBe("application/json");
         },
       );
 
--- a/packages/openapi-fetch/test/index.test.ts
+++ b/packages/openapi-fetch/test/index.test.ts
@@ -4,6 +4,8 @@ import createClient, {
   type BodySerializer,
   type FetchOptions,
   type MethodResponse,
+  type FetchOptions,
+  type HeadersOptions,
   type Middleware,
   type MiddlewareCallbackParams,
   type QuerySerializerOptions,
@@ -11,7 +13,7 @@ import createClient, {
   type PathBasedClient,
   createPathBasedClient,
 } from "../src/index.js";
-import { server, baseUrl, useMockRequestHandler, toAbsoluteURL } from "./fixtures/mock-server.js";
+import { baseUrl, server, toAbsoluteURL, useMockRequestHandler } from "./fixtures/mock-server.js";
 import type { paths } from "./fixtures/api.js";
 
 beforeAll(() => {
@@ -819,12 +821,7 @@ describe("client", () => {
         await client.GET("/self");
 
         // assert default headers were passed
-        expect(getRequest().headers).toEqual(
-          new Headers({
-            ...headers, // assert new header got passed
-            "Content-Type": "application/json", //  probably doesn’t need to get tested, but this was simpler than writing lots of code to ignore these
-          }),
-        );
+        expect(getRequest().headers).toEqual(new Headers(headers));
       });
 
       it("can be overridden", async () => {
@@ -850,7 +847,6 @@ describe("client", () => {
         expect(getRequest().headers).toEqual(
           new Headers({
             "Cache-Control": "no-cache",
-            "Content-Type": "application/json",
           }),
         );
       });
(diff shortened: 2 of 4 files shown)


docs/openapi-fetch/api.md
```diff
@@ -143,6 +143,18 @@ const { data, error } = await client.PUT("/submit", {
...
 
+::: tip
+
+For convenience, `openapi-fetch` sets `Content-Type` to `application/json` automatically
+for any request that provides value for the `body` parameter. When the `bodySerializer` returns an instance of `FormData`,
+`Content-Type` is omitted, allowing the browser to set it automatically with the correct message part boundary.
+
+You can also set `Content-Type` manually through `headers` object either in the fetch options,
+or when instantiating the client. Setting `Content-Type` to `null` will omit the header, however the native fetch API
+will likely set it to `text/plain` in this case.
```
goce-cz, line 154: Not sure we really need to mention the `null` thing anymore, when this behaves properly now. 🤷‍♂️ 

Opinions?

packages/openapi-fetch/test/index.test.ts
```diff
@@ -859,6 +855,138 @@ describe("client", () => {
...
+          },
+        });
+
+        expect(contentType).toBe(null);
+      });
+
+      const BODIES = [{ prop: "a" }, {}, "", "str", null, false, 0, 1, new Date("2024-08-07T09:52:00.836Z")] as const;
+      const METHOD_BODY_COMBINATIONS = BODY_ACCEPTING_METHODS.flatMap(([method]) =>
+        BODIES.map((body) => [method, body] as const),
+      );
```
goce-cz, line 903: @kerwanp Guessing from https://github.com/openapi-ts/openapi-typescript/pull/1825#discussion_r1706738232 you won't like this 🤣 

Happy to ditch the methods, but I'd rather keep the various bodies as this was broken until #1825.

kerwanp: Aha I'm not against this when behavior have to be different between cases, the pattern you use looks good to me! The methods constants could be moved at the top of the file to be reused in other tests.

packages/openapi-fetch/test/index.test.ts
```diff
@@ -859,6 +855,138 @@ describe("client", () => {
       });
     });
 
+    describe("content-type", () => {
+      const BODY_ACCEPTING_METHODS = [["PUT"], ["POST"], ["DELETE"], ["OPTIONS"], ["PATCH"]] as const;
```
kerwanp, line 859: nit: Is there a reason to have two dims arrays here?

kerwanp (commented):
Great work! Same thing as the other pull request, I'll let @drwpow review this

docs/openapi-fetch/api.md
drwpow: Yeah just remove it 🔪

drwpow:
See Martin’s comments/questions on tests. The implementation seems to solve the original issue. But in order to approve/merge we’ll need a `patch` changeset (see comment). Thanks for contributing!

packages/openapi-fetch/test/index.test.ts
```diff
@@ -859,6 +855,138 @@ describe("client", () => {
...
+            },
+            method,
+            fetchOptions: {
+              body,
+            },
+          });
+          // the fetch implementation won't allow sending a body without content-type,
+          // so it invents one up and sends it, hopefully this will be consistent across
+          // local environments and won't make the tests flaky
+          expect(contentType).toBe("text/plain;charset=UTF-8");
```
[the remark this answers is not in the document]
goce-cz: Yeah, that sounds better. Good idea.

packages/openapi-fetch/test/index.test.ts
goce-cz: Until they are actually reused in other tests, they IMO make more sense when they are collocated with the test itself. It's just less scrolling to understand how they are used. Anyway this is just my opinion I've grown over the years, nothing I would fight for.

I guess we are in agreement that this file already does too much and deserves some larger refactoring and splitting.  Let's maybe open a separate issue and discuss a bit what the desired structure could look like?

packages/openapi-fetch/test/index.test.ts
goce-cz: This is what the [each method](https://vitest.dev/api/#test-each) officially expects. It's a list of param sets where each set can consist of multiple params. The API supports a flat array, where `[x]` is treated as `[[x]]`, but I don't really fancy these magical interfaces and prefer to be explicit. Imagine what happens when `x` is actually an array?

goce-cz:
> But in order to approve/merge we’ll need a patch changeset (see comment).

I'll address this once I'm back from my vacation. Thanks!

docs/openapi-fetch/api.md
goce-cz: BTW Just a thought... I wonder whether allowing the `bodySerializer` to return `{contentType: "...", body: ...}` wouldn't be the best way to allow transport customization 🤔

goce-cz:
@drwpow should this really be a `patch` release? IIRC, in the issue, you mentioned that you would release it as a breaking change, just to be sure.

goce-cz:
> minor (“breaking”)

Unfamiliar combination 😄 , but here you go.

drwpow:
> > minor (“breaking”)
> 
> Unfamiliar combination 😄 , but here you go.

Thanks! This is actually following standard semver. Pre-1.0, the minor almost always signals breaking changes. Technically you can choose to either put features in minor releases or patches, but I choose the latter so a minor predictably means _some_ breaking change.

goce-cz:
> Pre-1.0, the minor almost always signals breaking changes.

This didn't make sense to me, until I stopped looking at (and releasing 🤦 ) a wrong package. (openapi-typescript v7).

Rebased and hopefully ready for merging.

docs/openapi-fetch/api.md
```diff
@@ -181,6 +181,17 @@ const { data, error } = await client.PUT("/submit", {
...
 ```
 
+::: tip
+
+For convenience, `openapi-fetch` sets `Content-Type` to `application/json` automatically
+for any request that provides value for the `body` parameter. When the `bodySerializer` returns an instance of `FormData`,
+`Content-Type` is omitted, allowing the browser to set it automatically with the correct message part boundary.
+
+You can also set `Content-Type` manually through `headers` object either in the fetch options,
+or when instantiating the client.
```
drwpow, line 191: I forget if I said this already, but thanks for adding this! Little docs improvements go a long way and are always welcome ❤️

drwpow (approved):
Great addition and tests, thank you!

bkis:
The documentation states

> For convenience, openapi-fetch sets Content-Type to application/json automatically for any request that provides value for the body parameter. When the bodySerializer returns an instance of FormData, Content-Type is omitted, allowing the browser to set it automatically with the correct message part boundary.
>
> You can also set Content-Type manually through headers object either in the fetch options, or when instantiating the client.

It seems like I actually can't set the header manually anymore. I have a fetch options object

```ts
{
    headers: {
      'Content-Type': 'application/x-www-form-urlencoded',
    },
    bodySerializer: (body: Record<string, string | number | boolean | undefined | null>) => {
      return new URLSearchParams(
        Object.entries(body).map(([key, value]) => [String(key), String(value)])
      ).toString();
    },
  }
```

...but the `Content-Type` header is omitted. This happens since these changes were made. When returning a `FormData` object in the `bodySerializer`, the API I target doesn't understand the request.  
It **does** work with the above options **including** the header, though. How can I force this header to be set?

PR #1827, opened 2024-08-07
fix(openapi-fetch): don't rely on DOM types

### LINKED ISSUE #1823: [openapi-fetch] HeadersInit support in Node.JS
**Description**

When using this library in a backend (e.g. for testing HTTP servers), it is reasonable to assume Node might be used (it works fine during runtime since Node supports the Fetch API). Currently, the generics used by this library rely on the `HeadersInit` type which is available in some TS environments but not others (the `lib` option in the TSConfig file seems to determine this; it looks like it is available if `lib` has `'DOM'` in it). The `@types/node` package declares the `Headers` type, but not `HeadersInit`. The `HeadersInit` type seemingly becomes unresolved if `@types/node` is used but not the `'DOM'` lib.

**Proposal**

I think this problem could be solved by removing the dependency of this package on the `'DOM'` lib by altering the [tsconfig file](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-fetch/tsconfig.json). The `HeadersInit` type could be defined as follows:

```typescript
type HeadersInit = NonNullable<
  ConstructorParameters<typeof Headers>[0]
>;
```

Then again, the `Headers` type and other fetch types do need to come from somewhere. I'm not sure what would be most appropriate; maybe they could be bundled with the declaration files somehow.

**Checklist**

- [ ] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

### LINKED ISSUE #1824: Invalid types using @tsconfig/node20
It looks like extending the tsconfig.json file from @tsconfig/node20 is causing type errors, particularly with the second argument (options) on routes and with parameters like params, body, headers, etc., not appearing in VSCode IntelliSense. However, when you remove the lib: ["es2023"] setting, everything works as expected.

Reference #1749 

Packages versions:
"openapi-fetch": "^0.10.5",
"openapi-typescript": "^7.3.0",
"typescript": "^5.5.4"

Generator script:
"generate:client": "openapi-typescript https://petstore3.swagger.io/api/v3/openapi.yaml -o ./src/lib/client.d.ts",

ngraef:
## Changes

This PR changes typing to avoid the global `HeadersInit`, which is declared in the DOM lib and therefore typically not included in Node projects.

Resolves #1823 
Resolves #1824

## How to Review

## Checklist

- [ ] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -24,7 +24,7 @@ export interface ClientOptions extends Omit<RequestInit, "headers"> {
 }
 
 export type HeadersOptions =
-  | HeadersInit
+  | Required<RequestInit>["headers"]
   | Record<string, string | number | boolean | (string | number | boolean)[] | null | undefined>;
 
 export type QuerySerializer<T> = (
--- a/.changeset/smooth-actors-tickle.md
+++ b/.changeset/smooth-actors-tickle.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Fix typing for environments without DOM lib


packages/openapi-fetch/src/index.d.ts
```diff
@@ -24,7 +24,7 @@ export interface ClientOptions extends Omit<RequestInit, "headers"> {
 }
 
 export type HeadersOptions =
-  | HeadersInit
+  | Required<RequestInit>["headers"]
```
ngraef, line 27: This change pulls a reference to `HeadersInit` from `RequestInit` rather than relying on the global declaration, which only exists in the DOM lib. In projects with the DOM lib, this reference will resolve to `HeadersInit` declared in `lib.dom.d.ts`. In projects with Node types and no DOM lib, this reference will resolve to `HeadersInit` from `undici-types/fetch.d.ts` as referenced from `@types/node/globals.d.ts`.

drwpow (approved):
Thank you for fixing! I agree with this logic. And here, existing tests will work. Thank you for improving!

PR #1833, opened 2024-08-09
fix type checking when `strictNullChecks` is disabled

### LINKED ISSUE #1778: params always required
The following snippet of code should be valid, as per the README:

```
import createClient from "openapi-fetch";
import type { paths } from "./petstore"; // npx openapi-typescript https://petstore3.swagger.io/api/v3/openapi.yaml -o petstore.d.ts

const client = createClient<paths>();

client.POST("/store/order", {
    body: {
        id: 0,
    },
})
```

However, it is failing with:
```
error TS2345: Argument of type '{ body: {}; }' is not assignable to parameter of type '{ params: { query?: never; header?: never; path?: never; cookie?: never; }; } & { body?: { id?: number; petId?: number; quantity?: number; shipDate?: string; status?: "placed" | "approved" | "delivered"; complete?: boolean; }; } & { ...; } & Omit<...> & { ...; }'.
  Property 'params' is missing in type '{ body: {}; }' but required in type '{ params: { query?: never; header?: never; path?: never; cookie?: never; }; }'.

  6 client.POST("/store/order", {
                                ~
  7     body: {
    ~~~~~~~~~~~
...
  9     },
    ~~~~~~
 10 })
  ~

  node_modules/openapi-fetch/dist/index.d.ts:87:9
    87     : { params: T["parameters"] }
               ~~~~~~
    'params' is declared here.
```

If I change it to:
```
import createClient from "openapi-fetch";
import type { paths } from "./petstore"; // npx openapi-typescript https://petstore3.swagger.io/api/v3/openapi.yaml -o petstore.d.ts

const client = createClient<paths>();

client.POST("/store/order", {
    params: {},
    body: {
        id: 0,
    },
})
```

the error goes away. 

Is this a recent regression? I don't remember it doing this before. I'm using typescript `5.5.4`, openapi-fetch `0.10.2`, and openapi-typscript `7.1.0`.

my `tsconfig.json`:
```
{
    "compilerOptions": {
        "module": "ESNext",
        "moduleResolution": "node",
        "noImplicitAny": true,
    },
    "files": [
        "test.ts"
    ]
}
```

**Checklist**

ngraef:
## Changes

This PR refactors some types to remove usage of `undefined extends T` conditional types. It also adds a new test suite that is type checked with `strictNullChecks` disabled to allow for increased support of that configuration.

Resolves #1778 

## How to Review

See context in [this comment on the issue](https://github.com/openapi-ts/openapi-typescript/issues/1778#issuecomment-2276217668). Using `undefined extends T` only behaves as expected when `strictNullChecks` is enabled. While this configuration is highly recommended for consuming projects, it is not the default, and we should design types that also work without it when possible.

#1778 describes a case where `params` is required on calls to routes with no parameters defined or no required parameters in the spec. At least two other cases are caused by the same underlying issue:
1. If all parameter types (query, header, path, and cookie) are defined on a route, `params` is not required, regardless of whether any of the parameters are required in the spec.
2. The `body` property is not required on calls to routes with a required request body defined in the spec.

This PR fixes the behavior of those cases.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/lazy-dancers-push.md
+++ b/.changeset/lazy-dancers-push.md
@@ -1,0 +1,7 @@
+---
+"openapi-typescript-helpers": patch
+"openapi-react-query": patch
+"openapi-fetch": patch
+---
+
+Fix identification of required properties when `strictNullChecks` is disabled
--- a/packages/openapi-fetch/package.json
+++ b/packages/openapi-fetch/package.json
@@ -59,6 +59,7 @@
     "test": "pnpm run \"/^test:/\"",
     "test:js": "vitest run",
     "test:ts": "tsc --noEmit",
+    "test:ts-no-strict": "tsc --noEmit -p test/no-strict-null-checks/tsconfig.json",
     "test-e2e": "playwright test",
     "e2e-vite-build": "vite build test/fixtures/e2e",
     "e2e-vite-start": "vite preview test/fixtures/e2e",
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -1,12 +1,13 @@
 import type {
   ErrorResponse,
   FilterKeys,
-  HasRequiredKeys,
   HttpMethod,
+  IsOperationRequestBodyOptional,
   MediaType,
   OperationRequestBodyContent,
   PathsWithMethod,
   ResponseObjectMap,
+  RequiredKeysOf,
   SuccessResponse,
 } from "openapi-typescript-helpers";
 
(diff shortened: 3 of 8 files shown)


.changeset/lazy-dancers-push.md
```diff
@@ -0,0 +1,7 @@
+---
+"openapi-typescript-helpers": patch
+"openapi-react-query": patch
+"openapi-fetch": patch
```
ngraef, line 4: I view this as a bug fix that implements the expected behavior. However, it does have the potential to break type checking in projects with incorrect code, so I can change this to `minor` if we want to be safe.

drwpow (approved):
Great fix, and great tests, thank you!

drwpow:
Shoot looks like another PR merged before this caused some conflicts. Once the conflicts are resolved, happy to merge & release.

PR #1837, opened 2024-08-10
Allow client option for custom dispatcher into fetch requests (e.g. to disable certificate validation) #1631

mellster2012:
## Changes

[Allow client option for custom dispatcher into fetch requests (e.g. to disable certificate validation)](https://github.com/openapi-ts/openapi-typescript/issues/1631)

## How to Review

This has been discussed before under a previous PR #1636. The branches were quite out of sync so it was easier to start from scratch.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/selfish-items-jump.md
+++ b/.changeset/selfish-items-jump.md
@@ -1,5 +1,0 @@
----
-"openapi-metadata": minor
----
-
-Handle array types and fix ApiProperty decorator type
--- a/packages/openapi-metadata/CHANGELOG.md
+++ b/packages/openapi-metadata/CHANGELOG.md
@@ -1,5 +1,11 @@
 # openapi-metadata
 
+## 0.1.0
+
+### Minor Changes
+
+- [#2001](https://github.com/openapi-ts/openapi-typescript/pull/2001) [`e864bec`](https://github.com/openapi-ts/openapi-typescript/commit/e864bec76cef70b16020983174c3288a09a7d250) Thanks [@kerwanp](https://github.com/kerwanp)! - Handle array types and fix ApiProperty decorator type
+
 ## 0.0.0
 
 ### Patch Changes
--- a/packages/openapi-metadata/package.json
+++ b/packages/openapi-metadata/package.json
@@ -1,7 +1,7 @@
 {
   "name": "openapi-metadata",
   "description": "Auto-Generate OpenAPI specifications from Typescript decorators",
-  "version": "0.0.0",
+  "version": "0.1.0",
   "author": {
     "name": "Martin PAUCOT",
     "email": "contact@martin-paucot.Fr"


mellster2012:
@drwpow This is the redone PR which was previously at https://github.com/openapi-ts/openapi-typescript/pull/1636

packages/openapi-fetch/src/index.d.ts
```diff
@@ -10,10 +10,14 @@ import type {
   SuccessResponse,
 } from "openapi-typescript-helpers";
 
+import type { Dispatcher } from "undici";
```
drwpow, line 13: Can we redeclare this type so we don’t have an `undici` dependency (even type-wise)? The type signatures can match, but I don’t want people who install `openapi-fetch` to also have undici install as well.

packages/openapi-fetch/package.json
```diff
@@ -77,6 +77,7 @@
     "openapi-typescript-codegen": "^0.25.0",
     "openapi-typescript-fetch": "^2.0.0",
     "superagent": "^9.0.2",
+    "undici": "^6.14.1",
```
drwpow, line 80: See other comment—unfortunately this is a `dependency` because of how it’s used, and I’d like to avoid that. Let’s just copy the `Dispatcher` type locally to prevent this.

drwpow (changes requested):
I’m still in favor of adding this feature, and this is getting there, but we’re missing documentation! We need to answer:

1. When would users use this option? What usecases/prior art? (We chatted about this in another issue, so just consolidating that into docs so others can find that info)
2. How would they opt in? Do they need to use undici? Will other custom fetch clients work?
3. What is some example code they can copy/paste to get started with this feature?

All of this should be clear to users how to use this function.

Also be sure to add a `patch` changeset to `openapi-fetch` (yes this is a feature add, but we’re following the convention of minor changes are breaking in 0.x, and this isn’t breaking, I believe)!

packages/openapi-fetch/test/index.test.ts
```diff
@@ -768,6 +769,33 @@ describe("client", () => {
...
+        method: "get",
+        path: "/self",
+        status: 200,
+        body: { message: "OK" },
+      });
+
+      await client.GET("/self");
+
+      // assert baseUrl and path mesh as expected
+      expect(getRequestUrl().href).toBe(toAbsoluteURL("/self"));
```
drwpow, line 791: Is this test testing what we think it is? I don’t see how this uses `dispatch` (or maybe I’m missing it?)

mellster2012:
I'll work on the improvements/suggestions as soon as time permits, including the documentation.

mellster2012:
The Dispatcher type is quite large and contains other (nested) undici custom types, and including all these extra types doesn't provide much benefit for the bloat. Also the fetch implementation here by default does not provide and optional RequestInit, rather an expanded requestInit is hidden within the CustomRequest type, which seems already like a deviation.

My suggestion would be to revert back to providing the optional addition of custom RequestInit properties into the ClientOptions, e.g. RequestInitExtensions of type Record<string, unknown>, so any number of custom properties can be added, and pass them to the fetch call as 2nd parameter as RequestInit when present (like it's currently proposed in this PR), with the user assuming the control and responsibility of using the correct custom fetch implementation with it.

@drwpow @kerwanp let me know what you think - thanks!

drwpow:
> My suggestion would be to revert back to providing the optional addition of custom RequestInit properties into the ClientOptions, e.g. RequestInitExtensions of type Record<string, unknown>, so any number of custom properties can be added, and pass them to the fetch call as 2nd parameter as RequestInit when present (like it's currently proposed in this PR), with the user assuming the control and responsibility of using the correct custom fetch implementation with it.
> 
> @drwpow @kerwanp let me know what you think - thanks!

I’m on board with trying that. Currently the problem with using the 2nd parameter approach is we found out that the `fetch()` implementation is different across different browsers and Node 😓. Specifically, most implementations want EITHER ✅ `fetch(Request)` OR ✅ `fetch(string, RequestInit object)`, but NEVER :x: `fetch(string, Request)` or :x: `fetch(Request, object)`.

Fortunately we [added tests to catch this](https://github.com/openapi-ts/openapi-typescript/pull/1719) that does catch some gamebreaking regressions from happening again, but taking a 2nd look over everything I’m not confident we have all the tests in place (will work on adding that). All that said, for now, we’re somewhat stuck with the single-param fetch API of `fetch(Request)` as long as we have our current middleware API that uses Request/Response for performance and minimalism. But to that end, we can adjust the middleware API too and improve that with, say, a proper RequestInit object. But one decision affects the other, and vice-versa.

Sorry for the extra context; this just overlaps several other APIs/behavior, and past regressions, so I wanted to spend a little more thought on it.

> Also the fetch implementation here by default does not provide an optional RequestInit, rather an expanded requestInit is hidden within the CustomRequest type, which seems already like a deviation.

To previous point, the CustomRequest type extends the built-in Request class, and though it does allow for additional properties, is still a valid implementation of the fetch spec that works in Node.js and all browsers, and is tested and working. So any changes to the `fetch()` call will just need a LOT of testing.

mellster2012:
Sounds, good, starting a fresh PR.

mellster2012:
https://github.com/openapi-ts/openapi-typescript/pull/2020

PR #1842, opened 2024-08-12
fix: allow use of `PathBasedClient` with generated `paths`

### LINKED ISSUE #1840: PathBasedClient cannot be (directly) used with generated paths
**Description**

Typescript fails:

```
│ test/index.test.ts(1888,58): error TS2344: Type 'paths' does not satisfy the constraint 'Record<string, Record<HttpMethod…
│   Index signature for type 'string' is missing in type 'paths'.
```

**Reproduction**

```ts
import type { PathBasedClient } from "../src/index.js";
import type { paths } from "./fixtures/api.js";

PathBasedClient<paths> // this line fails to compile
```
**Expected result**

Compiles

**Checklist**

gzm0:
To achieve this, we remove unnecessary type bounds.

This fixes #1840.

## Changes

- Remove unnecessary type bounds.
- Add tests for client types.

## How to Review

- Double check the removal of type bounds does not break anything.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/test/index.test.ts
+++ b/packages/openapi-fetch/test/index.test.ts
@@ -5,6 +5,8 @@ import createClient, {
   type Middleware,
   type MiddlewareCallbackParams,
   type QuerySerializerOptions,
+  type Client,
+  type PathBasedClient,
   createPathBasedClient,
 } from "../src/index.js";
 import { server, baseUrl, useMockRequestHandler, toAbsoluteURL } from "./fixtures/mock-server.js";
@@ -140,6 +142,11 @@ describe("client", () => {
         expectTypeOf(result.error).extract<{ code: number }>().toEqualTypeOf<{ code: number; message: string }>();
         expectTypeOf(result.error).exclude<{ code: number }>().toEqualTypeOf<never>();
       }
+    });
+
+    it("provides a Client type", () => {
+      const client = createClient<paths>({ baseUrl });
+      expectTypeOf(client).toEqualTypeOf<Client<paths>>();
     });
 
     describe("params", () => {
@@ -1875,6 +1882,11 @@ describe("client", () => {
   });
 
   describe("path based client", () => {
+    it("provides a PathBasedClient type", () => {
+      const client = createPathBasedClient<paths>({ baseUrl });
+      expectTypeOf(client).toEqualTypeOf<PathBasedClient<paths>>();
+    });
+
     it("performs a call without params", async () => {
       const client = createPathBasedClient<paths>({ baseUrl });
 
--- a/.changeset/little-knives-design.md
+++ b/.changeset/little-knives-design.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+fix: allow use of `PathBasedClient` with generated `paths`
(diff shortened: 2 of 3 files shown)


packages/openapi-fetch/test/index.test.ts
```diff
@@ -142,6 +144,11 @@ describe("client", () => {
       }
     });
 
+    it("provides a Client type", () => {
+      const client = createClient<paths>({ baseUrl });
+      expectTypeOf(client).toEqualTypeOf<Client<paths>>();
```
drwpow, line 149: These tests are great, thank you. I’d like to have more of these in general. And since these can run inside a `test-d.ts` test, I’ve been thinking it would be good to separate the type vs runtime tests (and have more of these granular type assertions).

What are your thoughts on splitting runtime vs type tests? And could these be good additions to get that rolling?

drwpow:
Great work! In favor of this merging, but see comment on type vs runtime tests

packages/openapi-fetch/test/index.test.ts
gzm0: This is an excellent question.

Libraries I've been working on so far (notably Scala.js) *had* to split runtime / type tests because of infrastructure restrictions (or more specifically test cases that should compile and test cases that shouldn't compile).

Personally, I often found that annoying. Because you end up with test cases like (pseudocode):

```
assert(max([1]) == 1)
assert(max([1,2]) == 2)
```

And somewhere completely different:

```
assertFailsCompile(max([]))
```

What we can do with TS is much nicer IMO:

```
assert(max([1]) == 1)
assert(max([1,2]) == 2)
assertThrows(
  // @ts-expect-error
  max([])
)
```

That all being said: IMO it does make sense to split tests by some axis (e.g. *the functionality* offered, independent of whether the functionality is offered by the type system or the runtime system or a combination thereof).

So in this case, it is true we are testing the `Client` type function and its inference (w/o inference issues, this test would not be necessary, one could simply read the return type of `createClient`). So this could / should indeed go into a different "describe" (and/or even a different file).

Now, along which axes to split tests (exported elements, "features", ...) is not easy. However, it is also something that can be done very ad-hoc, so there is no need to "get it right" the first time.

Splitting what is in `index.test.js` into multiple files is IMO definitely a good idea (merely based on the LOC count of `index.test.js`).

In summary:
- Splitting tests into different categories / units / ... is IMO a good thing.
- I'm not sure typing-time / run-time is the best splitting axis though.

drwpow: > In summary:
>
> - Splitting tests into different categories / units / ... is IMO a good thing.
> - I'm not sure typing-time / run-time is the best splitting axis though.

You’ve convinced me 🙂. Let’s split up the tests into more meaningful divisions then. I think the test suite “feels” complete currently because it’s just too big a file. But in reality is missing several cases still.

Can be a followup obviously! Thanks for weighing in.

gzm0:
For posterity: I have been working around this using an intersection type:

```ts
// BAD!!! See explanation below.
type Client = PathBasedClient<paths & Record<string, Record<string, string>>>;
```

However, this breaks inference on usage-site for wrong paths: They get accepted by the typechecker rather than flagged as non-existent.

A correct workaround is:

```ts
type Client = ResultType<typeof createPathBasedClient<paths>>;
```

Issue #1818, opened 2024-08-05
Better Benchmarking


gzm0:
Also a note—I’ve found the benchmarking to be wildly inconsistent 😓. It’s still in beta for Vitest, and I think we should have another alternative.

Or, put another way, the **absolute** numbers fluctuate wildly. But the **relative** measurements have some consistency (i.e. `openapi-fetch` being XX% faster than another package stays consistent-_ish_).

It’s been on my list to come up with more reliable benchmarks—would love your thoughts if there’s a better way to profile performance.

Also like I said before, anecdotally I haven’t had good experiences with performance regression tests (or, at least, not without an entire team at a company dedicated to maintaining them, and we don’t have that for this OSS project). CI (especially free CI) just fluctuates too widely with resources to be reliable. But I’m strongly in favor of manually-run benchmarking, that can be run when we want to see if something moves the needle up or down. And I think there’s improvements to be made in this area here, and I welcome your input! 🙏

_Originally posted by @drwpow in https://github.com/openapi-ts/openapi-typescript/issues/1810#issuecomment-2265690346_

gzm0:
First of all, my only real-life experience with benchmarking is on [Scala.js](https://github.com/scala-js/scala-js) (Scala to JavaScript transpiler, I'm one of the maintainers).

So far, we have performed 3 types for benchmarks (I'll compare them to what we probably want here below):
- Execution speed of generated code (https://github.com/sjrd/scalajs-benchmarks).
- Execution speed of the transpilation pipeline (in the core repo, [example](https://github.com/scala-js/scala-js/pull/4981#issuecomment-2090847203)).
- Memory usage of the transpilation pipeline ([example](https://github.com/scala-js/scala-js/pull/4981#issue-2275857455)).

We execute these benchmarks on-demand manually (as in: if we have a hunch it might affect performance, not even on release). We have not automated infrastructure to run them whatsoever.

The first benchmark bullet is probably the one most similar to what we'd want here, of course also the one I haven't been working on at all :shrug: .

[...]

PR #1850, opened 2024-08-14
Docs: Update benchmarks & sizes

drwpow:
## Changes

Updates the benchmarks & footprints for `openapi-typescript-fetch`. Sorry @ajaishankar! It wasn’t intentional; should have updated this sooner 😅 (PRs are also always welcome to correct out-of-date info too)

Also I don’t have access to the original machine I benchmarked on anymore, so all the numbers are different. Rather than re-run them, we’ll wait for the improvements from #1818 to make them more reliable. But `openapi-typescript-fetch` may pull out to be the lead here 🚀 !

## How to Review

N/A

## Checklist

N/A


ajaishankar:
No worries, thanks @drwpow!

Without the benchmark would not even have known that there was so much overhead targeting older versions of Javascript.

Again I can't say enough how awesome openapi-typescript is!

At work we have a code generated client library from a vendor that is 20MB (1.4MB gzip)!

Needless to say `openapi-typescript` handled their huge swagger spec with ease, and all that bloat disappeared at compile time! 😉

Issue #1853, opened 2024-08-14
Help wanted: add size-limit check in CI


drwpow:
`openapi-fetch` as it’s grown in features has grown in size! It was originally `2 kb` when it started out, but with middleware support and other features is now closer to `6 kb` (technically `5.8 kb` but who’s counting?—wait we need to!).

Adding something like [Size Limit](https://evilmartians.com/chronicles/size-limit-make-the-web-lighter) (or something else) in CI to make sure the bundle size doesn’t **increase** would be helpful. The goal of this check isn’t to prevent any features or bugfixes if they add to the weight at all, rather, it just helps prevent **unintentional size increases.** We can always decide to increase the package weight if it provides value for the weight.

jeiea:
Can this be closed?

drwpow:
Yes! I had thought this closed; thanks for the ping

PR #1859, opened 2024-08-17
feat(ci): add size-limit and apply it to openapi-fetch

yoshi2no:
## Changes

ref: #1853

- added a workflow named "Size Limit" using [size-limit-action](https://github.com/andresz1/size-limit-action) to check the size of `openapi-fetch` on the main branch.
- added a size-limit section to the root package.json and defined the limit.
  -  set `brotli` to false because I believe it’s better to define the limit based on simply minified values.
  -  I’ve set the limit to `6.5 kB` for now, but I would appreciate your input on what the ideal size should be.
- (optional) added an npm script called `size` to allow checking the size locally. This is optional, so feel free to remove it if it’s not needed.

```
  Size limit: 6.5 kB
  Size:       5.96 kB with all dependencies and minified
```

## How to Review

- Ensure that the GitHub Actions bot comments on the PR with the results of size-limit.

## Checklist
N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "format": "pnpm run -r --parallel --aggregate-output format",
     "test": "pnpm run -r --parallel --aggregate-output test",
     "test-e2e": "pnpm run -r --parallel --aggregate-output test-e2e",
+    "size": "pnpm run build && size-limit",
     "version": "pnpm run build && changeset version && pnpm i"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


yoshi2no:
Currently, there are no size-limit-related dependencies in the main branch, which seems to be causing an error 👀

package.json
```diff
@@ -13,17 +13,27 @@
...
     "prettier": "^3.3.2",
+    "size-limit": "^11.1.4",
     "typescript": "^5.4.5",
     "vitest": "^2.0.5"
-  }
+  },
+  "size-limit": [
+    {
+      "path": "packages/openapi-fetch/dist/index.min.js",
+      "limit": "6.5 kB",
```
yoshi2no, line 35: This should be replaced with a more appropriate value.

drwpow (approved):
Awesome, thank you so much for adding!

Yes I understand these checks do get weird in PRs. But since that check isn’t blocking (currently) it’s fine for now. Thanks again!

PR #1863, opened 2024-08-19
feat: Proposal of an approach to support readOnly / writeOnly properties

siketyan:
## Changes

refs #604

To support readOnly / writeOnly properties, I tried to generate a magic object type `{$read: T} | {$write: T}` and resolve the type with `Readable<T>` and `Writable<T>` helpers. This approach is inspired from [ColumnType API](https://kysely-org.github.io/kysely-apidoc/types/ColumnType.html) in Kysely, a TypeScript-first query builder library.

Since this pull request changes the generated type, it will be a breaking change for some users, I marked this as an experimental feature and require users to opt-in the feature if needed.

This pull request is working, but still just a PoC and some improvements may be needed. I would like request you all for any ideas or concerns about this approach.

## How to Review

- The generated types doesn't change anyway if `experimentalVisibility` is turned off
- `readOnly` or `writeOnly` properties can be transformed into a magic `{$read: T}` `{$write: T}` type correctly
- Path parameters, query parameters, request body, and response body are resolved correctly with visibility

TS Playground for reviewing type utilities: https://www.typescriptlang.org/play/?#code/C4TwDgpgBAShCGATA8gOwDYgDwBUB8UAvFAN5QAkATgogFxQ5QC+A3AFBuiRQDqlAlsAhpMuAsTLkA7gKH1GrDl2hwkIkAGkI2fEVJQA2hqj9UUANbaA9gDMGAXXlH7UCAA8hqRAGdYNdVgArqjmqFZSqAQA-FDG9KgQAG4QlMwGliC2Duyc4NB8gsIYmtpiemRGJmYZWTiODM6uHhBevgVCAcGh4ZFQMXFQCcmpTOnWdnU5yn5I8ABG6BBlxMgAtoJYbFD6laYW4w5OGi7unj4zKMVYpjYpsNGwUEcnza1+AMZWlIhY3sACqAA5gAaKBdMIRB6qRDzRa4ZwEZ7MYFbXiyIqYLQ6PBsPBTPJowSwpa6FbrYCbbYkVHbXbVA51JGnFrndoY7A3O48B48J4NY5NM6+OCfb6-f6mEFgkIQ3oxdrE+HHRH8+zsbZMFHbaHqLFiXH47jkbzvAAWEFW8HKNJMdFIVBo9D+AMBim221Q8FWECdEqB6vdiAgJoEYGA-CsqCivpdAe2YHg3m8Ui+dpI0nRMclikUbAAFNT3YMvT6oAByGxWKxlrXuhNJlPfehl+CtmtsJhQbzwcPeGz8YOE4CK41mi3wPAAShyBZt-DtLbm73bRc93ublerKM73d7-cH0JHJvNlqn7CAA

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/package.json
+++ b/packages/openapi-fetch/package.json
@@ -54,7 +54,7 @@
     "build:cjs": "esbuild --bundle src/index.js --format=cjs --outfile=dist/cjs/index.cjs && cp dist/index.d.ts dist/cjs/index.d.cts",
     "format": "biome format . --write",
     "lint": "biome check .",
-    "generate-types": "openapi-typescript -c test/redocly.yaml",
+    "generate-types": "openapi-typescript -c test/redocly.yaml --experimental-visibility",
     "pretest": "pnpm run generate-types",
     "test": "pnpm run \"/^test:/\"",
     "test:js": "vitest run",
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -9,6 +9,7 @@ import type {
   ResponseObjectMap,
   RequiredKeysOf,
   SuccessResponse,
+  Writable,
 } from "openapi-typescript-helpers";
 
 /** Options for each client instance */
@@ -29,7 +30,7 @@ export type HeadersOptions =
   | Record<string, string | number | boolean | (string | number | boolean)[] | null | undefined>;
 
 export type QuerySerializer<T> = (
-  query: T extends { parameters: any } ? NonNullable<T["parameters"]["query"]> : Record<string, unknown>,
+  query: T extends { parameters: any } ? Writable<NonNullable<T["parameters"]["query"]>> : Record<string, unknown>,
 ) => string;
 
 /** @see https://swagger.io/docs/specification/serialization/#query */
@@ -84,8 +85,8 @@ export type ParamsOption<T> = T extends {
   parameters: any;
 }
   ? RequiredKeysOf<T["parameters"]> extends never
-    ? { params?: T["parameters"] }
-    : { params: T["parameters"] }
+    ? { params?: Writable<T["parameters"]> }
+    : { params: Writable<T["parameters"]> }
   : DefaultParamsOption;
 
 export type RequestBodyOption<T> = OperationRequestBodyContent<T> extends never
--- a/packages/openapi-fetch/test/fixtures/api.d.ts
+++ b/packages/openapi-fetch/test/fixtures/api.d.ts
@@ -859,9 +859,15 @@ export type webhooks = Record<string, never>;
 export interface components {
     schemas: {
         Post: {
+            id: {
+                $read: string;
+            };
             title: string;
             body: string;
             publish_date?: number;
+            password: {
+                $write: string;
+            };
         };
         StringArray: string[];
         User: {
--- a/packages/openapi-fetch/test/fixtures/api.yaml
+++ b/packages/openapi-fetch/test/fixtures/api.yaml
@@ -495,15 +495,23 @@ components:
     Post:
       type: object
       properties:
+        id:
+          type: string
+          readOnly: true
         title:
           type: string
         body:
           type: string
         publish_date:
           type: number
+        password:
+          type: string
+          writeOnly: true
       required:
+        - id
         - title
         - body
+        - password
     StringArray:
       type: array
       items:
--- a/packages/openapi-fetch/test/index.test.ts
+++ b/packages/openapi-fetch/test/index.test.ts
@@ -114,6 +114,7 @@ describe("client", () => {
               updated_at: number;
             }
           | {
+              id: string;
               title: string;
               body: string;
               publish_date?: number;
@@ -721,6 +722,7 @@ describe("client", () => {
             title: "",
             publish_date: 3,
             body: "",
+            password: "",
           },
         });
       });
--- a/packages/openapi-react-query/package.json
+++ b/packages/openapi-react-query/package.json
@@ -54,7 +54,7 @@
     "dev": "tsc -p tsconfig.build.json --watch",
     "format": "biome format . --write",
     "lint": "biome check .",
-    "generate-types": "openapi-typescript test/fixtures/api.yaml -o test/fixtures/api.d.ts",
+    "generate-types": "openapi-typescript test/fixtures/api.yaml -o test/fixtures/api.d.ts --experimental-visibility",
     "pretest": "pnpm run generate-types",
     "test": "pnpm run \"/^test:/\"",
     "test:js": "vitest run",
--- a/packages/openapi-react-query/test/fixtures/api.d.ts
+++ b/packages/openapi-react-query/test/fixtures/api.d.ts
@@ -819,9 +819,15 @@ export type webhooks = Record<string, never>;
 export interface components {
     schemas: {
         Post: {
+            id: {
+                $read: string;
+            };
             title: string;
             body: string;
             publish_date?: number;
+            password: {
+                $write: string;
+            };
         };
         StringArray: string[];
         User: {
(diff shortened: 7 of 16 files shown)


drwpow:
Thanks so much for investigating this! 🙏 I think you’re on the right track in supporting this. I would honestly love for this to Just Work™ without a flag, and perhaps we can explore a solution that makes it invisible/backwards-compatible for users. If we find that, then we could just ship this. But yes, any breaking changes in behavior would have to be behind flags.

That said, I’m pretty strongly against the current implementation detail of turning types into objects `{ $read: T }`. Even users of `openapi-fetch` use the generated types directly in other parts of their application. As soon as we mess with those types, the original file is unusable. 
So I’d like to avoid that at all cost. Also, generating 2 schemas—one for internal usage, one for `openapi-fetch`—is something else I’d like to avoid at all cost. The one schema should work for both.

This isn’t a correct solution, but I’m interested if there are just additional types we can generate alongside the existing ones ([playground](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgQewLYAc0guAZ2QG8AoZC5AhACwgzgIC4TzL2AFNAsFs9gZWAATFjyigA5gG42g9mGBgANhDFgJIGXPkUARmmEBPdZu272WAK57lwAjQD6wuJAD8LEFYx7osi5RYjAQA7mhQolQaUv4ByI5QEHDCAPIgyibIAEQiWbEBjiESkGkZLFlBBKHhwnk67AC++cgNpK2kEAAeOFBgyGBGWCgASkmp6UZcPMgAvMgpGEoAPDro2Lj4YAQA2lnUdAwEWQC6u1NgJwA0q5g4eIS7+-SMJ2fcF6dZhcUQpUYnyAA9IDkCA0JB+nRkKAAG7QAgoNAweJFJS-Cb1SgAH2yCTGfyyyBxX1RJQmhOByCwUDQejgtiMyDgyF8YEgUGQITgjLAaGQwj5YBo9iZVl5DEUCDgygypAAfLIOt1wn0BkNkAB1H5-c6zeaLMArdhrO6bHZ7WjPI6fc5XG7re5bR6Ww6vLK2z545IE46YijEr3jDKEgOk9HBoEg6m0+kZJksiBs6Cc7n9PkCyEiuBizCuYBSmVGeXSIA)):

CleanShot 2024-08-29 at 08 59 44@2x

This this approach:

- ✅ Preserves the original types 
- ✅ Adds the `readOnly` and `writeOnly` support additively,
- ❌ However, adds additional properties that shouldn’t be read by default when accessing the original type (the underscore _helps_ but may not be enough to ruin some setups)

---

I think probably a better solution overall is just **generating an entirely-new `interface` altogether** for these additions, that `openapi-fetch` would just use instead. In other words, keep the original `paths` and `components` 100% untouched, and generate something like `readOnlyPaths`/`readOnlyComponents`/etc. that adds things like `Omit<components["schemas"]["Post"], "password">`. 

Then in `openapi-fetch` it does 2 traversals: first looks in `readOnlyPaths` to see if a key matches, then just normal `paths` if not.

Generating additional types is not a breaking change. And the beauty of generating types is they don’t incur any runtime weight (yes, we will run into TS inference performance at scale, however, for this change just don’t even take that into consideration). So it’s “free” in a sense to generate additional structures that `openapi-fetch` can read, without disrupting the core types, if that makes sense.

---

Anyways, all that said, again, this is on the right path! But I would like to explore solutions that somehow sidestep touching the original `paths`/`components`/etc. and add types additively that only `openapi-fetch` know about. But also without generating a completely-new 2nd schema.

If we can meet that goal, no flag needed 😎

siketyan:
@drwpow 

Thank you for your feedback 🙏 

I agree that the approach without any breaking changes nor flags is better at all. This was the biggest problem of my approach.

> I think probably a better solution overall is just **generating an entirely-new `interface` altogether** for these additions, that `openapi-fetch` would just use instead. In other words, keep the original `paths` and `components` 100% untouched, and generate something like `readOnlyPaths`/`readOnlyComponents`/etc. that adds things like `Omit<components["schemas"]["Post"], "password">`.
>
> Then in openapi-fetch it does 2 traversals: first looks in readOnlyPaths to see if a key matches, then just normal paths if not.

AFAIK we need to specify the generated paths interface to use it in `openapi-fetch`:

```ts
import type {paths} from './$schema.js'

const client = createClient<paths>({ baseUrl: "https://myapi.dev/v1/" });
```

So we cannot traverse 2 options here without changing the users' code like:

```ts
import type {pathsWithVisibility} from './$schema.js';

const client = createClient<pathsWithVisibility>({ baseUrl: "https://myapi.dev/v1/" });
```

Also we need to produce 2x or more amount of the code to support the both of usages with or without visibility, for each components, paths, and operations 🤯. This affects all users regardless the user need the visibility or not. If we can switch the behaviour of the types generation, the size increase will not occur.

Are there any solution to avoid breaking changes and flags without this problem?

drwpow:
> So we cannot traverse 2 options here without changing the users' code like:
> 
> ```ts
> import type {pathsWithVisibility} from './$schema.js';
> 
> const client = createClient<pathsWithVisibility>({ baseUrl: "https://myapi.dev/v1/" });
> ```
> 
> Also we need to produce 2x or more amount of the code to support the both of usages with or without visibility, for each components, paths, and operations 🤯. This affects all users regardless the user need the visibility or not. If we can switch the behaviour of the types generation, the size increase will not occur.

Yeah this is tough, and no straightfoward solution. I will say that **don’t worry about the size/amount of types generated.** Types are “free”—they all get baked out during the build and don’t impact runtime. That’s the whole philosophy of this library.

This doesn’t directly answer your questions, but for the openapi-fetch and openapi-typescript library:

**Hard requirements**
- Default type generation must not change for openapi-typescript (without a breaking version)
- openapi-typescript schema objects must be usable directly as simple types (no generics needed)
- openapi-fetch must be usable with default openapi-typescript types (otherwise there’s a danger your fetch client is using different typings from other parts of your app)
    - Note: emphasis on _default_. There are some openapi-typescript flags that break openapi-fetch and that’s OK; not all opt-in flags are supported

_Note: some of the hard requirements here may make it seem like this feature is impossible to add. That’s not the case! But it is a strong indicator that this feature would warrant a breaking change for both. I was trying to see if there are creative solutions to adding this without a major version bump, but maybe not! And that’s OK._

**Soft goals** (nice-to-haves)
- openapi-fetch imports the same `paths` interface
- openapi-fetch adds support for readOnly/writeOnly by just changing its internals rather than its API

Note that both of these things can be conceded as a last resort. As long as there’s justification for this being the only way.

Again, this feature seems like it should be default behavior because it’s a part of the spec. And I’ll admit this part of the spec just wasn’t front-of-mind while developing 7.x and the initial version of openapi-fetch. This seems like it is deeper-reaching than originally-anticipated, which is also probably why it’s taken so long to add support.

drwpow:
> Also we need to produce 2x or more amount of the code to support the both of usages with or without visibility, for each components, paths, and operations 🤯.

Also, just a thought 🤔 we may not need “2x” the generated types. Not if we generated the “raw” interfaces with the spread objects everywhere (`{$read: T}`), and the original `components` interface  just has the `Readable<T>` and `Writeable<T>` wrappers everywhere referencing the raw types. It behaves the same. And should be lightweight as it’s just referencing the bigger object.

Now for openapi-fetch, this would still require that alternate `paths` like you mentioned. There’s probably not, but what if there’s a way to detect the presence of those `Readable<T>` and `Writeable<T>` generics in `paths` and unwrap them somehow? My gut says they’re already resolved, and behave like the flattened types, but there may be something clever we could do.

Still, even if not, and openapi-fetch does require a separate generic out of necessity, I don’t mind that. We could make a semi-breaking change requiring this just for more correct OpenAPI behavior.

This is basically your PR and approach 99% as you originally authored it, just with some very slight tweaks.

drwpow:
Closing in favor of #2549

PR #1864, opened 2024-08-19
feat(openapi-react-query): Pass abort signal to fetch

zsugabubus:
## Changes

See changeset.

## How to Review

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/signal.md
+++ b/.changeset/signal.md
@@ -1,0 +1,5 @@
+---
+"openapi-react-query": patch
+---
+
+Pass down signal to fetch function this way `useQuery` and `useSuspenseQuery` can cancel queries.
--- a/packages/openapi-react-query/src/index.ts
+++ b/packages/openapi-react-query/src/index.ts
@@ -70,10 +70,10 @@ export default function createClient<Paths extends {}, Media extends MediaType =
       return useQuery(
         {
           queryKey: [method, path, init],
-          queryFn: async () => {
+          queryFn: async ({ signal }) => {
             const mth = method.toUpperCase() as keyof typeof client;
             const fn = client[mth] as ClientMethod<Paths, typeof method, Media>;
-            const { data, error } = await fn(path, init as any); // TODO: find a way to avoid as any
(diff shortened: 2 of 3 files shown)


kerwanp (approved):
Clean update and great test! 👌

Might be a stupid question but should it be aswell added to other `useSuspenseQuery` and `useMutation`? @zsugabubus

zsugabubus:
@kerwanp As far as I know mutations cannot be cancelled, `mutationFn` does not have a `signal` parameter, it just gets the variables and I neither can find an API for it. However, it turns out `useSuspenseQuery` can be cancelled, I updated the code, thanks for the question, at first I was too lazy because I just copy pasted the same code and it did not work.

.changeset/signal.md
```diff
@@ -0,0 +1,5 @@
+---
+"openapi-react-query": minor
```
drwpow, line 2: Yes this adds a feature, but for pre-1.0 we’re only releasing breaking changes in minor releases

```suggestion
"openapi-react-query": patch
```

SteinarVrenne:
This is great to prevent memory leaks, but is there a way to circumvent this behavior in case you want the fetch to continue even if the component has been unmounted?

zsugabubus:
My first thought was that this kind of stuff should be configured via Tanstack Query but I couldn't find such setting. So you have to build your own `queryFn`. You can either create a wrapper around `queryOptions().queryFn` or just call `client.GET` directly. A slightly less correct solution would be to clear `request.signal` in a middleware.

But are you 100% sure you want to go down this road? Maybe `useMutation` or `queryClient.fetchQuery` would be more appropriate for your use-case?

SteinarVrenne:
Thanks for the quick reply! 

Some background to why I am asking: I am currently migrating to this library, and the old behaviour was to fire and forget a fetch call. This fetch call is very expensive and not cancellable, so it locked down the API until it finished (but of course, no one took the answer). This happened because of a route change so the component unmounted. It was also not cached, so overall not good. 

It can definitely be solved using the options you provided, or using the fetchClient, but I wish maybe an override of this feature was available in the options :)

Issue #1785, opened 2024-07-28
`SuccessResponseJSON` and `ErrorResponseJSON` always return `never`


piousdeer:
**Description**

I think `SuccessResponseJSON` and `ErrorResponseJSON` weren't updated along with their non-`JSON` counterparts in v7 and they should've been something like this:

```diff
- export type JSONLike<T> = FilterKeys<T, `${string}/json`>;
+ export type JSONLike = `${string}/json`;

- export type SuccessResponseJSON<PathMethod> = JSONLike<SuccessResponse<ResponseObjectMap<PathMethod>>>;
+ export type SuccessResponseJSON<PathMethod> = SuccessResponse<ResponseObjectMap<PathMethod>, JSONLike>;
```

| Name                 | Version                      |
| :------------------- | :--------------------------- |
| `openapi-typescript` | `7.1.0`                      |
| Node.js              | `v21.7.2`                      |
| OS + version         | `NixOS` |

**Reproduction**

```ts
import type { SuccessResponseJSON } from "openapi-typescript-helpers";

type x = SuccessResponseJSON<paths["/a/valid/path"]["get"]>; // `never`

type y = SuccessResponse<ResponseObjectMap<paths["/a/valid/path"]["get"]>>; // works
```

**Checklist**

- [ ] My OpenAPI schema passes the [Redocly validator](https://redocly.com/docs/cli/commands/lint/) (`npx @redocly/cli@latest lint`)
- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

drwpow:
That’s a good suggestion, and we’d welcome a PR for those. Those are just helper utils that people can use or not use, and we make no guarantees about breaking API changes. And if they can be improved like this, PRs are always welcome

PR #1871, opened 2024-08-21
fix(openapi-typescript-helpers): fix SuccessResponseJSON and ErrorResponseJSON helpers

brunolca:
## Changes

fix SuccessResponseJSON and ErrorResponseJSON helpers

## How to Review

Should fix issue  #1785 

I tested it with the following code:

```typescript
import type { ErrorResponseJSON, SuccessResponseJSON } from "./index.d.ts";
import type { components, paths } from "./v1.d.ts";

type Expect<T extends true> = T;
type Equal<X, Y> = (<T>() => T extends X ? 1 : 2) extends <T>() => T extends Y ? 1 : 2 ? true : false;

type A = ErrorResponseJSON<paths["/fact"]["get"]>;

type B = number;

type Tests = [
  Expect<Equal<ErrorResponseJSON<paths["/fact"]["get"]>, components["schemas"]["Error"]>>,
  Expect<Equal<SuccessResponseJSON<paths["/breeds"]["get"]>, components["schemas"]["Breed"][]>>,
];
```

## Checklist

- [ ] Unit tests updated

I could add a test.d.ts to test the types but i thought it could (on breaking changes) break the global test script (used in ci...)

### DIFF
--- a/.changeset/silly-apes-know.md
+++ b/.changeset/silly-apes-know.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript-helpers": patch
+---
+
+fix SuccessResponseJSON and ErrorResponseJSON helpers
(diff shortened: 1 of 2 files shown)


drwpow (approved):
Thank you!

PR #1876, opened 2024-08-23
feat(openapi-typescript): Optional Export Root Type Aliases

BradHacker:
## Changes

Adds the CLI option `--root-types` to optionally export all of the types from `components` for convenience.

image

A dependency on `change-case` was added given the need to account for various casing in the OpenAPI component schema name (e.g. `'some-type'`) not always being valid for TypeScript type identifier. (If this dependency can be avoided, I'm open to suggestions). Possible edge cases for this breaking are *open for discussion*, so please let me know of all the edge cases you can think of!

Relevant Issues:
- https://github.com/openapi-ts/openapi-typescript/issues/1414
- https://github.com/openapi-ts/openapi-typescript/issues/1598

Previous PR's:
- https://github.com/openapi-ts/openapi-typescript/pull/1260

## How to Review

Keep in mind this is my first ever time working with the TypeScript Compiler API and AST so any recommendations for a way to write this in a cleaner way is more than welcome.

I've added a GitHub example called `github-api-root-types.ts` and the `update:examples` should generate this. Tried to stick to what seems to be standard practices here. If there are other preferred methods of testing/demonstrating this feature, please let me know!

## Checklist

- [x] Unit tests updated


BradHacker:
Not sure why the `openapi-react-query` package is failing to build in CI as `openapi-fetch` is building properly

drwpow:
> Not sure why the `openapi-react-query` package is failing to build in CI as `openapi-fetch` is building properly

GitHub Actions has had some minor timeouts/failures in the last week. Nothing major, but it seems a few PRs failed for no reason. Can’t complain about free CI though!

docs/cli.md
```diff
@@ -104,21 +104,22 @@ The following flags are supported in the CLI:
 | `--help`                           |       |          | Display inline help message and exit                                                                                |
 | `--version`                        |       |          | Display this library’s version and exit                                                                             |
 | `--output [location]`              | `-o`  | (stdout) | Where should the output file be saved?                                                                              |
-| `--redocly [location]`               |       |          | Path to a `redocly.yaml` file (see [Multiple schemas](#multiple-schemas))                                           |
+| `--redocly [location]`             |       |          | Path to a `redocly.yaml` file (see [Multiple schemas](#multiple-schemas))                                           |
```
drwpow, line 107: ❤️ Thanks for fixing little things like this! It means a lot

packages/openapi-typescript/test/transform/components-object.test.ts
```diff
@@ -464,6 +464,37 @@ describe("transformComponentsObject", () => {
...
+        };
+    };
+    responses: never;
+    parameters: never;
+    requestBodies: never;
+    headers: never;
+    pathItems: never;
+}
+export type SomeTypeSchema = components['schemas']['SomeType'];`,
+        options: { ...DEFAULT_OPTIONS, rootTypes: true },
```
drwpow, line 495: So this is a great start to testing this, and this is exactly what we need! But we will need a lot more tests to handle all of the scenarios [from my original requirements](https://github.com/openapi-ts/openapi-typescript/pull/1260#issuecomment-1658694431):

1. **Conflicts**: test that `#/components/schemas/schemaFoo` and `#/components/schemas/Schemafoo` don’t conflict (and many combinations including hyphens and invalid JS characters like leading numbers!). We’ll need to handle:
  - Case sensitivity (this is handled by the large object, and this is intentional)
  - Invalid characters: `-` and `.` are the most common. Beyond that IMO not too important.
  - Leading numbers (e.g. `1schema`)
2. **ALL of components object, not just `#/components/schemas`:
  - `#/components/schemas`
  - `#/components/responses`
  - `#/components/parameters`
  - `#/components/requestBodies`

We will need all of this to happen in the same PR, because it’s very likely that supporting all this could lead to refactoring and churn based on the implementation. As-is I like the path you’re on! But we need to have all of these present and tested before we merge this.

drwpow (changes requested):

docs/cli.md
BradHacker: Your welcome! Gotta keep things consistent 🙂

packages/openapi-typescript/test/transform/components-object.test.ts
BradHacker: Addressed point 1 with [d40a7fd](https://github.com/openapi-ts/openapi-typescript/pull/1876/commits/d40a7fd9e852d61c62c68c116b8886ea4bf9ed84). All conflicting names are suffixed by a `_2`, `_3`, etc. as per the original discussion in the previous PR you linked. This behavior is up for discussion. Not sure this is a great developer experience, but at the same time don't feel it warrants an error given the OpenAPI spec permits these things. Given that, figured this should be the compliant portion and work around the limitations of JS to comply to the OpenAPI spec as closely as possible. Maybe possible to just print warnings about conflicts and automatic suffixing if detected as a happy compromise?

Filled out the test to cover the other components properties in [7edf333](https://github.com/openapi-ts/openapi-typescript/pull/1876/commits/7edf3336fff3929da0a2d78583de99da357ef522) and tested for conflicts across them in [f64e0e6](https://github.com/openapi-ts/openapi-typescript/pull/1876/commits/f64e0e6b98199654a12cc2596600345fad1fceab)

packages/openapi-typescript/test/transform/components-object.test.ts
```diff
@@ -464,6 +464,241 @@ describe("transformComponentsObject", () => {
...
+export type SchemaError = components['schemas']['Error'];
+export type ResponseOk = components['responses']['OK'];
+export type ResponseNoContent = components['responses']['NoContent'];
+export type ResponseErrorResponse = components['responses']['ErrorResponse'];
+export type ResponseSomeType = components['responses']['SomeType'];
+export type ParameterSearch = components['parameters']['Search'];
+export type RequestBodyUploadUser = components['requestBodies']['UploadUser'];
+export type HeaderAuth = components['headers']['Auth'];
+export type PathItemUploadUser = components['pathItems']['UploadUser'];`,
+options: { ...DEFAULT_OPTIONS, rootTypes: true },
```
drwpow, line 699: 💯

packages/openapi-typescript/test/transform/components-object.test.ts
drwpow: > Given that, figured this should be the compliant portion and work around the limitations of JS to comply to the OpenAPI spec as closely as possible.

+1

> Maybe possible to just print warnings about conflicts and automatic suffixing if detected as a happy compromise?

Yeah I’m open to that, but don’t feel strongly. I think developers will realize if they’re importing the wrong things quickly; the main thing to handle was just types disappearing / not exporting from `rootTypes` at all would have people raising bug reports.

drwpow:
> A dependency on `change-case` was added given the need to account for various casing in the OpenAPI component schema name (e.g. `'some-type'`) not always being valid for TypeScript type identifier. (If this dependency can be avoided, I'm open to suggestions).

Nope! This is fine. Any time a lightweight dependency is added that serves a need is OK. Better to have well-tested code handling a common usecase than write code off-the-shelf.

drwpow (approved):
Tests looks great, and this meets all the requirements! 🎉 Thanks for adding; a lot of people will love using this feature.

Be sure to add a **minor changeset** (see comment) so we can merge & release this!

BradHacker:
> Tests looks great, and this meets all the requirements! 🎉 Thanks for adding; a lot of people will love using this feature.
> 
> Be sure to add a **minor changeset** (see comment) so we can merge & release this!

Added a minor changeset with a brief summary of the changes. Wanted this feature for a project and I've already been using this update locally. It cleans up the usage so much when trying to add types to my code! Glad to help everyone out 😄

PR #1893, opened 2024-08-31
fix(openapi-typescript): correct generated type when using `prefixItems`

### LINKED ISSUE #1892: Incorrect type when using `prefixItems`
**Description**

Using `prefixItems` generates incorrect type.

| `openapi-typescript` | `7.3.3`                      |
| Node.js              | `20.17.0`                      |

**Reproduction**

The following schema
```yaml
type: array
items:
  type: number
prefixItems:
  - number
  - number
```
generates incorrect type:
```ts
[number, number];
```

**Expected result**

This should generate open-tuple type according to JSON Schema 2020-12 draft[^1]:
```ts
[number, number, ...number[]];
```
and closed-tuple type should be generated[^2][^3] by
```yaml
type: array
items: false
prefixItems:
  - number
  - number
```
or
```yaml
type: array
unevaluatedItems: false
prefixItems:
  - number
  - number
```

[^1]: https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#name-prefixitems
[^2]: https://json-schema.org/draft/2020-12/release-notes#closed-tuple
[^3]: https://json-schema.org/understanding-json-schema/reference/array#unevaluateditems

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

zyoshoka:
## Changes

Closes #1892

## How to Review

Note that this PR also adds support for `unevaluatedItems`, which was supported in OAS 3.1.0.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/test/transform/schema-object/array.test.ts
+++ b/packages/openapi-typescript/test/transform/schema-object/array.test.ts
@@ -35,11 +35,77 @@ describe("transformSchemaObject > array", () => {
       },
     ],
     [
-      "tuple > prefixItems",
+      "tuple > prefixItems (open tuple using items)",
       {
         given: {
           type: "array",
           items: { type: "number" },
+          prefixItems: [{ type: "number" }, { type: "number" }, { type: "number" }],
+        },
+        want: `[
+    number,
+    number,
+    number,
+    ...number[]
+]`,
+        // options: DEFAULT_OPTIONS,
+      },
+    ],
+    [
+      "tuple > prefixItems (closed tuple using unevaluatedItems)",
+      {
+        given: {
+          type: "array",
+          unevaluatedItems: { type: "number" },
+          prefixItems: [{ type: "number" }, { type: "number" }, { type: "number" }],
+        },
+        want: `[
+    number,
+    number,
+    number,
+    ...number[]
+]`,
+        // options: DEFAULT_OPTIONS,
+      },
+    ],
+    [
+      "tuple > prefixItems (open tuple not specified)",
+      {
+        given: {
+          type: "array",
+          prefixItems: [{ type: "number" }, { type: "number" }, { type: "number" }],
+        },
+        want: `[
+    number,
+    number,
+    number,
+    ...unknown[]
+]`,
+        // options: DEFAULT_OPTIONS,
+      },
+    ],
+    [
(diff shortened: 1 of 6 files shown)


drwpow (changes requested):
Thank you for opening this PR, and fixing the generation. I agree that this is probably the more “correct” solution to what was originally implemented. But this would be a breaking change for people, because the original implementation (whether intentionally or not) essentially ignored `items`.

We’ll accept the PR as-implemented, but **we’ll need to make this an opt-in feature flag, for now.** In the next breaking release we can look into changing the default behavior to match this (which I strongly think we will). But I don’t want to release a new major version now just for this change and this change alone.

Would you be willing to put this behind an opt-in flag?

packages/openapi-typescript/test/transform/schema-object/array.test.ts
```diff
@@ -179,7 +245,7 @@ describe("transformSchemaObject > array", () => {
       {
         given: {
           type: "array",
-          items: { type: "number" },
+          items: false,
```
drwpow, line 248: This specifically is the breaking change I’m referring to—we’ll need to ensure that the original test case of `items: { type: "number" }` stays tested, and it generates the same output. While we’re on version 7.x this can’t change (and we can revisit it when it’s time to release 8.x).

In general, modifying an original test is a good hint that it may be a breaking change.

packages/openapi-typescript/test/transform/schema-object/array.test.ts
```diff
@@ -45,6 +45,72 @@ describe("transformSchemaObject > array", () => {
         want: `[
     number,
     number,
+    number,
+    ...number[]
```
drwpow, line 49: This other modification to an existing test may also _technically_ be a breaking change, but I think because it’s purely additive, could be considered a bugfix. I would be OK with this going in as-is.

duncanbeevers:
Part of this PR is superseded by https://github.com/openapi-ts/openapi-typescript/pull/2148
I did not include support for `unevaluatedItems`, so this PR is likely worth revisiting once that work lands in order add that support.

Issue #754, opened 2021-09-27
Support `patternProperties` in `components.schemas`


KevinPoole:
I have a `component.schema` of the following form:

```
      "SomeSchemaName": {
        "title": "SomeSchemaName",
        "type": "object",
        "properties": {
          "somePropertyName": {
            "type": "object",
            "additionalProperties": false,
            "patternProperties": {
              "^[0-9]+$": {
                "$ref": "#/components/schemas/OtherSchemaName"
              }
            }
          }
        }
      },
```

Of particular importance is the fact that I am utilizing `patternProperties`. This is a valid usage according to the following issue:

- https://github.com/OAI/OpenAPI-Specification/issues/687

Anyone know if this is something that would be expected to work right now? If not, any ideas on what it would take to support this?

drwpow:
Interesting. What would be the desired TypeScript generated types for this?

alexserov:
> Interesting. What would be the desired TypeScript generated types for this?

Having an interface that allows fields with arbitrary keys but specifies types would be extremely helpful.

For example, we have the following schema: [MobilityData/gbfs-json-schema/gbfs.json](https://github.com/MobilityData/gbfs-json-schema/blob/master/v2.3/gbfs.json#L26-L74).
Current version of the `openapi-typescript` produces the following output:
```TS
   interface GbfsV2_3 {
      //...
      data: { [key: string]: unknown };
    };
```
Instead of unknown we could have the following:
```TS
   interface GbfsV2_3 {
      //...
      data: {
        [key: string]: {
          //...
          feeds?: {
           //...
           };
         };
      };
    };
```
This will allow us extract the type of the `data` field like follows:
```TS
type DataType = GbfsV2_3['data'];
type DataFieldType = DataType[keyof DataType];
```

---

If there are multiple patterns in the schema, `openapi-typescript` can generate type unions. Not too strict, but better than `unknown`.

MarioHi1:
Any update on this? Without this spec, it is not truly 3.1 compatible.

drwpow:
I’d love to support patternProperties, but I’m not sure this would work as intended in static TypeScript types. Mixing static object keys with generic signatures like `[key: string]` already leads to some rough edges, but also TS isn’t capable of truly implementing RegEx without some runtime component.

I’m not saying it’s necessarily impossible; I’m just saying I’d love a PoC that shows how this could be handled in TS and would welcome PRs adding this!

gzm0:
[...]

PR #1901, opened 2024-09-04
RFC: feat: support pattern properties

gzm0:
## Changes

Add support for `patternProperties` in addition to `additionalProperties` (#754).

## How to Review

Overall approach so far only.

## Checklist

- [ ] Unit tests updated


gzm0:
@drwpow something like this is what I have been thinking. Of course it will not validate the regexes, but it will at least generate a union type that represents the schema.

drwpow:
Yeah this is a tricky one but I’m open to adding whatever support we can here. Better than simply ignoring it altogether. I’m in favor of this approach as it shouldn’t break backwards compatibility for most folks

drwpow:
Sorry—just realized I went a while without checking in on this. I’m happy with the direction of this PR, and we’ve discussed a lot of things outside of here 🙂. If we just add a couple tests for this I think this direction is good to ship!

gzm0:
Obsoleted by #2288.

PR #1936, opened 2024-10-01
fix(openapi-fetch): Prevent orphan ampersands in query from empty arrays

BlakeSzabo:
## Changes

Fixes a case where multiple empty arrays are passed as query params will create multiple orphaned ampersands in the resultant request URL

Closes https://github.com/openapi-ts/openapi-typescript/issues/1935

## How to Review

I decided to effectively ignore empty arrays in the querySerializer, the issue was they were returning `""` from `seralizeArrayParam` which was then caught up in the `.join("&")`. However, we could just as easily `.filter` out the empty strings and catch for the general case.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/strong-boats-drive.md
+++ b/.changeset/strong-boats-drive.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Fix multiple empty arrays in query params appending extra ampersands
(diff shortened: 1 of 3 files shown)


drwpow (approved):
Wonderful, thank you!! 💯 No feedback, great tests 🙂

PR #1937, opened 2024-10-03
fix(openapi-fetch): Return union of possible responses (data & error)

DjordyKoert:
## Changes
fixes https://github.com/openapi-ts/openapi-typescript/issues/1723, fixes https://github.com/openapi-ts/openapi-typescript/issues/1933, fixes https://github.com/openapi-ts/openapi-typescript/issues/1820

This change ensures that a union of possible types from the response content is returned.

## How to Review

The behaviour of the new type can be seen in the newly added `types.test.ts` file. This also helps ensure possible breaking changes in the future are caught by the failing test(s) (these types are exported after all and can be used by users of this package).

## Checklist

- [x] Unit tests updated


packages/openapi-fetch/test/never-response/never-response.test.ts
```diff
@@ -0,0 +1,143 @@
...
+      params: { path: { id: 123 } },
+    });
+
+    assertType<components["schemas"]["Post"] | undefined>(data);
+
+    // assert correct URL was called
+    expect(actualPathname).toBe("/posts/123");
+
+    // assert 204 to be transformed to empty object
+    expect(data).toEqual({});
```
DjordyKoert, line 62: Side-note: 

I do not fully agree with this behaviour at https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-fetch/src/index.js#L155. 
1. This makes it impossible to have proper types for 204 codes. 
2. If users of this package DO return some data from 204's (this shouldn't be the case of course, but if their openapi docs DO specify this then it should still return the data from the response instead of transforming it like this)

This PR ensures that a documented 204 response (`content?: never`) will return a `data` as a possible `undefined`. BUT this will now **never** be the case because of this behaviour.

Would love to hear some feedback and if it could be considered to remove this 😄

packages/openapi-fetch/test/common/response.test.ts
```diff
@@ -49,8 +49,8 @@ describe("response", () => {
         {},
       );
 
-      assertType<undefined>(result.data);
-      // @ts-expect-error: FIXME when #1723 is resolved; this shouldn’t throw an error
+      //@ts-expect-error impossible to determine data type for invalid path
+      assertType<never>(result.data);
```
drwpow, line 53: 💯

packages/openapi-fetch/test/never-response/never-response.test.ts
drwpow: Yes we’re open to revisiting that behavior! Agree that at this point, that does more harm than good. That was an early behavior that just never received much questioning until now.

I’d be open to either of the following:

1. Keep this PR as-is, release as a `patch`, then release that `204` change as a `minor` (breaking) change, OR
2. Include that change in this PR, and release everything as a `minor` (breaking) change

Either way it would constitute a breaking change, but IMO a welcome one. No strong preference on #1 or #2, but #1 would be easier to release faster

packages/openapi-fetch/test/types.test.ts
```diff
@@ -0,0 +1,206 @@
...
+        204: { content: never };
+        206: { content: { "text/plain": `206 text/plain` } };
+        404: { content: { "text/plain": `404 text/plain` } };
+        500: { content: { "application/json": { error: `500 application/json` } } };
+      }
+
+      test("returns all possible responses", () => {
+        type Response = GetResponseContent<MixedResponses>;
+        assertType<Response>({ data: "200 application/json" });
+        // @ts-expect-error It picks literal over string
```
drwpow, line 23: Minor feedback (here and below): if we could scope the `// @ts-expect-error` a little better by breaking apart the next line into as many lines as possible, that ensures we’re ignoring the correct error as much as we can (otherwise, with multiple type errors going on, we may be throwing a different error than intended)

drwpow (approved):
This is a FANTASTIC PR, thank you! Great tests that ensure it’s working.

Will wait to merge/release depending on whether or not you’d like to tackle the `204` breaking behavior change as part of this PR, or in a followup. Either works.

packages/openapi-fetch/test/never-response/never-response.test.ts
DjordyKoert: My preference would also be to keep this PR as-is aswell. I will make a new PR to fix the `204` thing soon 😄 

After that I will have a look at what @zkulbeda suggested for exposing the `status` to help narrowing the `data` type even more 😎

drwpow: That works! Also noticed there’s a tiny conflict with the openapi-typescript-helpers; can merge once that’s resolved.

Also feel free to make an adjustment there if it causes a regression; that recent change was just an arbitrary QoL improvement and didn’t fix any type issues like your PR does

DjordyKoert: Great! Fixed the feedback you gave so everything should be good now 😄

drwpow (approved):
Awesome work. Thanks again! 🙂 Will release this now.

PR #1943, opened 2024-10-10
Security schemes support

FreeAoi:
## Changes

_What does this PR change?

This pull request try to closes [Add support for security and securitySchemes](https://github.com/openapi-ts/openapi-typescript/issues/922)

## How to Review

## Checklist

- [x] Unit tests updated


packages/openapi-typescript/src/types.ts
```diff
@@ -543,44 +543,54 @@ export interface XMLObject extends Extensable {
...
-      type: "mutualTLS";
-    }
-  | {
-      /** REQUIRED. Tye type of the security scheme. */
-      type: "oauth2";
-      /** REQUIRED. An object containing configuration information for the flow types supported. */
-      flows: OAuthFlowsObject;
-    }
-  | {
-      /** REQUIRED. Tye type of the security scheme. */
```
drwpow, line 578: Question: why this change here? My understanding was that other properties are required or not based on the `type`; this changes them so that they’re all potentially nullable, which feels like a regression?

drwpow (changes requested):
Thanks for adding! Aside from a minor question, **we’ll need dedicated tests for this!** I see that technically one test did update, but that was incidental. We’ll need to explicitly test that different security schemas generate as-expected.

Note that the `examples/` do NOT count as dedicated tests. Those are large snapshot tests where we don’t look carefully at the contents; we just a) make sure they generate without errors, and b) don’t generate invalid TS. Beyond that, their contents are largely-ignored.

Requesting changes only so we can add dedicated security schema tests, but again, generally **in favor of this merging.** Thanks for taking the time to contribute!

MrFox131:
First of all, thank you for your time @FreeAoi .
Are there any est on how soon it will be released?

drwpow:
Closing this PR because there hasn’t been any movement for a little while, but again, we’d gladly accept PRs with this exact implementation **so long as tests are added!** For anyone else following, feel free to open another PR with this approach but with tests added

Rendez:
> First of all, thank you for your time @FreeAoi . Are there any estimations on how soon it will be released?

+1

PR #1945, opened 2024-10-10
add MethodResponse for openapi-react-query

freshgiammi:
## Changes

Following suit from https://github.com/openapi-ts/openapi-typescript/pull/1831, this PR adds the ability to get the return type of an endpoint from a `OpenapiQueryClient` client.

## How to Review

The rationale implements the same behaviours as the referenced PR, only changing expected type for the incoming client to be compliant with the `OpenapiQueryClient` interface (which doesn't extend `Client`).

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-react-query/src/index.ts
+++ b/packages/openapi-react-query/src/index.ts
@@ -11,7 +11,7 @@ import {
   useQuery,
   useSuspenseQuery,
 } from "@tanstack/react-query";
-import type { ClientMethod, FetchResponse, MaybeOptionalInit, Client as FetchClient } from "openapi-fetch";
+import type { ClientMethod, FetchResponse, MaybeOptionalInit, Client as FetchClient, ClientPathsWithMethod } from "openapi-fetch";
 import type { HttpMethod, MediaType, PathsWithMethod, RequiredKeysOf } from "openapi-typescript-helpers";
 
 type InitWithUnknowns<Init> = Init & { [key: string]: unknown };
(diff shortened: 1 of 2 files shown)


kerwanp:
Hi @freshgiammi, thanks for willing to contribute!

Is there any reason why this needs to be in openapi-react-query as it requires an openapi-fetch client anyway?

freshgiammi:
Hi @kerwanp, the issue is that `OpenapiQueryClient` doesn't extend `FetchClient`, which means two things:
- it can't be provided to the original `MethodResponse` included in `openapi-fetch`
- it can't be passed to `ClientPathsWithMethod` in order to infer the `Path` type

I've pushed an improvement that solves the second point, by directly creating the `PathsWithMethod` type (which also makes auto-completion work correctly)

```tsx
Path extends CreatedClient extends OpenapiQueryClient<infer Paths, infer _Media>
	? PathsWithMethod<Paths, Method>
	: never,
```

Doing this only means we only need `FetchResponse` from `openapi-fetch` to correctly resolve the return type. However I guess we could work around this too.

As for 
> Is there any reason why this needs to be in openapi-react-query as it requires an openapi-fetch client anyway?

Considering this can only be used for react-query clients, it doesn't make much sense to have it in the `openapi-fetch` package considering you'd have to bring in `OpenapiQueryClient` and the whole type chain (and users wouldn't use this, unless they use `openapi-react-query` too)

drwpow (approved):
This looks fine for me to merge, thank you! I’m going to add a changeset for you so you’ll get credit for this.

PR #1948, opened 2024-10-14
fix(openapi-typescript-helpers): fix SuccessResponseJSON, ErrorResponseJSON and RequestBodyJSON

piousdeer:
## Changes

Since #1871,
1. SuccessResponseJSON and ErrorResponseJSON no longer return `never`, but may return a non-application/json response.
2. RequestBodyJSON returns `{ "application/json": T }` instead of `T`.

This fixes these issues.

## Checklist

- [ ] Unit tests updated


drwpow (approved):
Looks great; thanks!

PR #1950, opened 2024-10-16
feat(openapi-react-query): use `queryOptions` helper

freshgiammi:
## Changes

The current state of queryOptions doesn't take advantage of the `queryOptions` helper, which has the ability of linking the returned data shape to the queryKey.
This is useful when using `queryClient.getQueryData` and especially handy for optimistic updates managed via `queryClient.setQueryData`.

The proposed implementation only consumes the helper to build the queryOptions property, and doesn't touch the way options are passed to `useQuery/useSuspenseQuery` to avoid narrowing the allowed options to the ones provided by the helper.

## How to Review
When consuming `queryClient.getQueryData` by passing a queryKey coming from a queryOptions object, see that the returned type isn't `unknown` anymore, but `TData | undefined`.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-react-query/src/index.ts
+++ b/packages/openapi-react-query/src/index.ts
@@ -11,14 +11,24 @@ import {
   useQuery,
   useSuspenseQuery,
   queryOptions,
-  type DataTag,
-  type DefinedInitialDataOptions,
-  type UndefinedInitialDataOptions,
 } from "@tanstack/react-query";
 import type { ClientMethod, FetchResponse, MaybeOptionalInit, Client as FetchClient } from "openapi-fetch";
 import type { HttpMethod, MediaType, PathsWithMethod, RequiredKeysOf } from "openapi-typescript-helpers";
 
 type InitWithUnknowns<Init> = Init & { [key: string]: unknown };
+
+// biome-ignore lint/suspicious/noRedeclare: <https://stackoverflow.com/questions/52760509/typescript-returntype-of-overloaded-function>
+type OverloadedReturnType<T> = T extends { (...args: any[]): infer R; (...args: any[]): infer R }
+  ? R
+  : T extends (...args: any[]) => infer R
+    ? R
+    : any;
+
+type OverloadedParameters<T> = T extends { (...args: infer P1): any; (...args: infer P2): any }
+  ? P1
+  : T extends (...args: infer P) => any
+    ? P
+    : never;
 
 export type QueryKey<
   Paths extends Record<string, Record<HttpMethod, {}>>,
@@ -26,14 +36,14 @@ export type QueryKey<
   Path extends PathsWithMethod<Paths, Method>,
 > = readonly [Method, Path, MaybeOptionalInit<Paths[Path], Method>];
 
-export type QueryOptionsObject<Paths extends Record<string, Record<HttpMethod, {}>>, Media extends MediaType> = <
+export type QueryOptionsFunction<Paths extends Record<string, Record<HttpMethod, {}>>, Media extends MediaType> = <
   Method extends HttpMethod,
   Path extends PathsWithMethod<Paths, Method>,
   Init extends MaybeOptionalInit<Paths[Path], Method>,
   Response extends Required<FetchResponse<Paths[Path][Method], Init, Media>>, // note: Required is used to avoid repeating NonNullable in UseQuery types
   Options extends Omit<
     UseQueryOptions<Response["data"], Response["error"], Response["data"], QueryKey<Paths, Method, Path>>,
-    "queryKey" | "queryFn" | "initialData"
+    "queryKey" | "queryFn"
   >,
 >(
   method: Method,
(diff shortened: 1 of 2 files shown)


zsugabubus (commented):
What do you think about simply changing return type of `QueryOptionsFunction` to be `queryKey: DataTag<...>`, just like what `queryOptions` does internally? I think it would cause less complications than this overloaded stuff.

freshgiammi:
Probably a better idea, seeing that [complexity is growing with SkipToken](https://github.com/TanStack/query/blob/main/packages/react-query/src/queryOptions.ts) (even though I'm not too sure of the implications).
You want to take care of this your PR and have https://github.com/openapi-ts/openapi-typescript/pull/1952 superseed this?

freshgiammi:
@zsugabubus welp it looks like we're out of luck. Either we ask for the symbol to be exported, or we have to rely on the overloads... unless you have a better idea, I'm not sure of alternatives here.

zsugabubus:
The below code works fine for me. I guess we don't need to actually assign anything, `dataTagSymbol` is used by the type system only and no code will ever read this variable. I could be more precise with the typing instead of an `as any` but I think that would too verbose with little added value.

<details>
<summary>diff against my PR</summary>

```diff
 packages/openapi-react-query/src/index.ts        |  6 ++++--
 packages/openapi-react-query/test/index.test.tsx | 11 +++++++++++
 2 files changed, 15 insertions(+), 2 deletions(-)

diff --git a/packages/openapi-react-query/src/index.ts b/packages/openapi-react-query/src/index.ts
index 97d83a6..01f848b 100644
--- a/packages/openapi-react-query/src/index.ts
+++ b/packages/openapi-react-query/src/index.ts
@@ -10,2 +10,3 @@ import {
   type SkipToken,
+  type DataTag,
   useMutation,
@@ -43,3 +44,3 @@ export type QueryOptionsFunction<Paths extends Record<string, Record<HttpMethod,
     UseQueryOptions<Response["data"], Response["error"], Response["data"], QueryKey<Paths, Method, Path>>,
-    "queryFn"
+    "queryFn" | "queryKey"
   > & {
@@ -49,2 +50,3 @@ export type QueryOptionsFunction<Paths extends Record<string, Record<HttpMethod,
     >;
+    queryKey: DataTag<QueryKey<Paths, Method, Path>, Response["data"]>;
   }
@@ -124,3 +126,3 @@ export default function createClient<Paths extends {}, Media extends MediaType =
   const queryOptions: QueryOptionsFunction<Paths, Media> = (method, path, ...[init, options]) => ({
-    queryKey: [method, path, init as InitWithUnknowns<typeof init>] as const,
+    queryKey: [method, path, init as InitWithUnknowns<typeof init>] as const as any,
     queryFn,
diff --git a/packages/openapi-react-query/test/index.test.tsx b/packages/openapi-react-query/test/index.test.tsx
index 97550d4..06ebfa7 100644
--- a/packages/openapi-react-query/test/index.test.tsx
+++ b/packages/openapi-react-query/test/index.test.tsx
@@ -233,2 +233,13 @@ describe("client", () => {
     });
+
+    it("returns queryKey that allows data to be inferred", async () => {
+      const fetchClient = createFetchClient<paths>();
+      const client = createClient(fetchClient);
+
+      const { queryKey } = client.queryOptions("get", "/string-array");
+
+      const data = queryClient.getQueryData(queryKey);
+
+      expectTypeOf(data).toEqualTypeOf<string[] | undefined>();
+    });
   });
```

</details>

kerwanp:
I think it is a good idea to rely on tanstack-query helper types as much as possible. When I originally wrote this library I went straight to the point but now when using more complex features the types are not correct (queryOptions, callbacks, etc).

drwpow:
Sorry for the delay on this. We’ve merged some other PRs that have caused this PR to fall out of line with `main`.

Would love if you could rebase. It seems like this PR could still be useful?

freshgiammi:
@drwpow no problem, I've had a couple of busy months as well and couldn't keep this updated. Just rebased, let me know if there's anything else I can do 👍

abhiaagarwal:
Status on this? I use optimistic updates a lot, this would be a wonderful QOL improvement :)

irsooti:
What's missing here? It's a huge ergonomic feature to have

freshgiammi:
Hello, I've rebased the PR and changed the approach used here, considering the challenge of using the `queryOptions` helper due to function overloads. I've updated the PR description accordingly, but tldr; we're manually tagging the `queryKey` (which is what the helper does as well) to make sure the data/error types for the query are attached to the `queryKey`.

This PR has been open for quite a while, can we get this reviewed?

cc @drwpow @kerwanp

drwpow:
Thanks for your patience, we’re getting caught up on reviews. Could you add a `patch` changeset (see comment) and we can get it released? We’re using patch on 0.x to signal any sort of backwards-compatible change and this seems like it is. With a changeset, happy to approve & release.

Also, in case you missed it, we will be [stopping development on this package](https://github.com/openapi-ts/openapi-typescript/discussions/2559) and returning to just working on `openapi-typescript` to make better use of our limited resources. So we‘re going to release in-flight PRs like this one, but just know after maybe a few more releases (~3 months or so) this package will be frozen.

freshgiammi:
Changeset added!

---
No worries! I didn't see the discussion, and I'm sad that fetch and react-query packages are being deprecated.
I definitely fall into the category of trivial users (_tanstack-router with multiple non-ts api services_), but I completely understand and agree with the points you raised. 

Given that both libraries are relatively small, we should be able to re-implement them should the need arise. I'm looking forward to the guide and to openapi-typescript v8 (and especially validation, hopefully with something that supports [standard-schema](https://standardschema.dev/))!

Still, thank you for these libraries: they've been such a breeze to work with compared to other solutions we've tried.

PR #1958, opened 2024-10-23
Add a flag to skip Schema prefix when rendering root types

sultaniman:
## Changes

This PR adds another cli flag `--root-types-no-prefix` which complements `--root-types`.
If set it will render root types without `Schema` prefix.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/bin/cli.js
+++ b/packages/openapi-typescript/bin/cli.js
@@ -11,26 +11,27 @@ const HELP = `Usage
   $ openapi-typescript [input] [options]
 
 Options
-  --help                     Display this
-  --version                  Display the version
-  --redocly [path], -c       Specify path to Redocly config (default: redocly.yaml)
-  --output, -o               Specify output file (if not specified in redocly.yaml)
-  --enum                     Export true TS enums instead of unions
-  --enum-values              Export enum values as arrays
-  --dedupe-enums             Dedupe enum types when \`--enum=true\` is set
-  --check                    Check that the generated types are up-to-date. (default: false)
-  --export-type, -t          Export top-level \`type\` instead of \`interface\`
-  --immutable                Generate readonly types
-  --additional-properties    Treat schema objects as if \`additionalProperties: true\` is set
-  --empty-objects-unknown    Generate \`unknown\` instead of \`Record<string, never>\` for empty objects
-  --default-non-nullable     Set to \`false\` to ignore default values when generating non-nullable types
+  --help                                      Display this
+  --version                                   Display the version
+  --redocly [path], -c                        Specify path to Redocly config (default: redocly.yaml)
+  --output, -o                                Specify output file (if not specified in redocly.yaml)
+  --enum                                      Export true TS enums instead of unions
+  --enum-values                               Export enum values as arrays
+  --dedupe-enums                              Dedupe enum types when \`--enum=true\` is set
+  --check                                     Check that the generated types are up-to-date. (default: false)
+  --export-type, -t                           Export top-level \`type\` instead of \`interface\`
+  --immutable                                 Generate readonly types
+  --additional-properties                     Treat schema objects as if \`additionalProperties: true\` is set
+  --empty-objects-unknown                     Generate \`unknown\` instead of \`Record<string, never>\` for empty objects
+  --default-non-nullable                      Set to \`false\` to ignore default values when generating non-nullable types
   --properties-required-by-default
-                             Treat schema objects as if \`required\` is set to all properties by default
-  --array-length             Generate tuples using array minItems / maxItems
-  --path-params-as-types     Convert paths to template literal types
-  --alphabetize              Sort object keys alphabetically
-  --exclude-deprecated       Exclude deprecated types
-  --root-types (optional)    Export schemas types at root level
+                                              Treat schema objects as if \`required\` is set to all properties by default
+  --array-length                              Generate tuples using array minItems / maxItems
+  --path-params-as-types                      Convert paths to template literal types
+  --alphabetize                               Sort object keys alphabetically
+  --exclude-deprecated                        Exclude deprecated types
+  --root-types (optional)                     Export schemas types at root level
+  --root-types-no-schema-prefix (optional)    Do not add "Schema" prefix to types at the root level (should be used with --root-types)
 `;
 
 const OUTPUT_FILE = "FILE";
@@ -56,6 +57,9 @@ if (args.includes("-it")) {
 if (args.includes("--redoc")) {
   errorAndExit(`The --redoc config flag has been renamed to "--redocly" (or -c as shorthand).`);
 }
+if (args.includes("--root-types-no-schema-prefix") && !args.includes("--root-types")) {
+  console.warn("--root-types-no-schema-prefix has no effect without --root-types flag");
+}
 
 const flags = parser(args, {
   boolean: [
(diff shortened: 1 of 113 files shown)


packages/openapi-typescript/bin/cli.js
```diff
@@ -11,26 +11,27 @@ const HELP = `Usage
...
+  --enum-values                               Export enum values as arrays
+  --dedupe-enums                              Dedupe enum types when \`--enum=true\` is set
+  --check                                     Check that the generated types are up-to-date. (default: false)
+  --export-type, -t                           Export top-level \`type\` instead of \`interface\`
+  --immutable                                 Generate readonly types
+  --additional-properties                     Treat schema objects as if \`additionalProperties: true\` is set
+  --empty-objects-unknown                     Generate \`unknown\` instead of \`Record<string, never>\` for empty objects
+  --default-non-nullable                      Set to \`false\` to ignore default values when generating non-nullable types
   --properties-required-by-default
-                             Treat schema objects as if \`required\` is set to all properties by default
```
drwpow, line 28: Minor feedback: I’d like to keep the current column split, and simply wrap to the next line like this. Reason is many people have ~80char terminal windows, and if the columns are too spaced out it’s hard to read. The current column sizing is arbitrary and can be adjusted a little bit, but IMO the adjustment here spaces them out too much

drwpow (commented):
In favor of this in general! I think we need some tests here just to ensure it works as expected, but that test can be minimal since it’s an opt-in flag.

Also agree with the “off/on” approach here, since the prefixes MAY be `Schema` but also may be `Response` or `Request`, etc. I don’t think it would make sense to have a flag to manually set that prefix, since that prefix isn’t hardcoded to begin with.

But with that in mind, a test that **tests what happens when you have a naming conflict between `#/components/schemas` and `#/components/responses` would be great (not _required_, but it would really help validate that all bases are being covered)

sultaniman:
Hey @drwpow thanks for review, I will add tests and update the PR with requested changes some time next week.

packages/openapi-typescript/bin/cli.js
sultaniman: Done.

sultaniman:
@drwpow I will add another test which tests when there is a parameter with the same name. I added a test case for the new flag.

drwpow (approved):
Sorry for the delay. This looks FANTASTIC. Thank you!

sultaniman:
That's awesome, thanks @drwpow!

fluctus:
@sultaniman @drwpow First, great job on adding the new --root-types-no-prefix flag! It’s a really useful addition.

However, I've noticed that when the --root-types-no-prefix flag is used alongside the --enum flag, it generates a root type with the same name as the enum, leading to naming conflicts.

Given that enums already serve as types in TypeScript, it seems unnecessary to generate a separate root type in these cases. Skipping the root type when both flags are used would help avoid conflicts and streamline the output.

It would be great if this could be addressed, whether by skipping the root type or considering other potential solutions to prevent such conflicts. Looking forward to seeing how this can be improved!

sultaniman:
Hey @fluctus @drwpow I think I can adjust with additional check if `--enum` flag is passed then maybe skipping prefix trimming, but probably we first need to discuss the desired behavior.

fluctus:
👋 Good point @sultaniman. Regarding the desired behavior, I think it would be a bit odd if this feature doesn't work when the `--enum` flag is used. A better solution might be to make both flags work together without conflicts.

One option could be to skip generating the root type **only for enums** when both the `--root-types-no-prefix` and `--enum` flags are used. This would avoid naming conflicts while keeping the intended behavior intact. If there's another way to handle both flags together without issues, that would work too.

drwpow:
Yeah with some of these flags, they’re not always perfectly compatible with one another. We could either throw an error that these flags don’t work together. Or do what @fluctus suggested by modifying the behavior of `--root-types-no-prefix`. As long as we **DON’T** modify the behavior of the `--enum` flag in any way I think that’s a fine solution.

sultaniman:
alright then as @fluctus suggested it would mean if `--enum` is passed then we skip `--root-types` only for enums?

fluctus:
@sultaniman if @drwpow agrees with this solution, would you be handling the fix?

sultaniman:
Hey @fluctus sorry for being slow, I was sick, will try to carve out some time this week for it.

fluctus:
@sultaniman No worries, hope you're feeling better! Thanks for taking the time to look into this.

sultaniman:
@fluctus can you please provide some minimal OpenAPI spec example with enum definitions you have?

fluctus:
@sultaniman Sure, this is a sample spec that generates types with the same name for enums:
```
openapi: 3.0.3
info:
  title: My API
  version: 1.0.0
components:
  schemas:
    NonEnumSchema:
      type: string
    EnumSchema:
      type: string
      enum:
        - VALUE1
        - VALUE2
```
Without `--root-types-no-prefix`:
```
export type paths = Record<string, never>;
export interface components {
    schemas: {
        NonEnumSchema: string;
        /** @enum {string} */
        EnumSchema: EnumSchema;
    };
    responses: never;
    parameters: never;
    requestBodies: never;
    headers: never;
    pathItems: never;
}
export type SchemaNonEnumSchema = components['schemas']['NonEnumSchema'];
export type SchemaEnumSchema = components['schemas']['EnumSchema'];
export type $defs = Record<string, never>;
export enum EnumSchema {
    VALUE1 = "VALUE1",
    VALUE2 = "VALUE2"
}
export type operations = Record<string, never>;
```
With `--root-types-no-prefix`:
```
export type paths = Record<string, never>;
export interface components {
    schemas: {
        NonEnumSchema: string;
        /** @enum {string} */
        EnumSchema: EnumSchema;
    };
    responses: never;
    parameters: never;
    requestBodies: never;
    headers: never;
    pathItems: never;
}
export type NonEnumSchema = components['schemas']['NonEnumSchema'];
export type EnumSchema = components['schemas']['EnumSchema'];
export type $defs = Record<string, never>;
export enum EnumSchema {
    VALUE1 = "VALUE1",
    VALUE2 = "VALUE2"
}
export type operations = Record<string, never>;
```
Which results in these TS errors:
`TS2502: EnumSchema is referenced directly or indirectly in its own type annotation.`
`TS2456: Type alias EnumSchema circularly references itself.`
`TS2567: Enum declarations can only merge with namespace or other enum declarations.`

sultaniman:
@fluctus thank you very much for providing an example I will use it in the tests.

theo-staizen:
Hi @sultaniman @fluctus I have also run into the "TS2567: Enum declarations can only merge with namespace..." error, because I was running both `--root-types-no-prefix` and `--enums`. Is there no way to fix this?

fluctus:
@theo-staizen yes, @sultaniman is working on a fix that will work with both flags.

sultaniman:
Hey guys, sorry for delays I wanted to work on this last week but was unable to do so. I will do it in the coming weekend and the following week.

theo-staizen:
hi @sultaniman, sorry to bother you, but have you had a chance to work on this at all? is there anything we can do to help?

sultaniman:
Good evening @drwpow @theo-staizen @fluctus sorry for being so slow had many personal things going on recently. I linked the [PR](https://github.com/openapi-ts/openapi-typescript/pull/2206) which disables `--root-types-no-schema-prefix` if used with `--enum` flag.

Issue #1940, opened 2024-10-04
`"nullable": true` is not respected for empty objects.


BillyBobFry:
**Description**

When a JSON API spec has `"nullable": true` for an empty object, the resulting TS type does not include `null`.

**Reproduction**

Copy the following into a JSON API spec:
```
    "NullableEmptyObject": {
        "nullable": true,
        "properties": {},
        "title": "NullableEmptyObject",
        "type": "object"
      },
      "NullableObject": {
        "nullable": true,
        "properties": {
          "name": {
            "type": "string"
          }
        },
        "title": "NullableObject",
        "type": "object"
      },
      "NullableString": {
        "nullable": true,
        "title": "NullableString",
        "type": "string"
      },
```

The generated TS types are
```
        /** NullableEmptyObject */
        NullableEmptyObject: Record<string, never>;
        /** NullableObject */
        NullableObject: {
            name?: string;
        } | null;
        /** NullableString */
        NullableString: string | null;
```

**Expected result**

`NullableEmptyObject` should be typed as `Record<string, never> | null`.

**Checklist**

- [x] My OpenAPI schema passes the [Redocly validator](https://redocly.com/docs/cli/commands/lint/) (`npx @redocly/cli@latest lint`)
- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

I could open a PR in a few weeks, but I'm not able to do so right now!

gduliscouet-ubitransport:
It is a duplicate of https://github.com/openapi-ts/openapi-typescript/issues/1821

PR #1959, opened 2024-10-23
Fix handling of nullable schemas

### LINKED ISSUE #1821: [openapi-typescript] Incorrect type generated for nullable objects
Thanks everyone for the new updates of `openapi-typescript`! I have found an issue which I believe might be related to the new major version (7). 

**Description**

In version `7.3.0` and OpenAPI version `3.0.3`, combining `type: object` and `nullable: true` does not produce the expected union type.

**Reproduction**

Run the CLI with default options on the following schema:

```yaml
openapi: "3.0.3"
info: { title: "Test", version: "0" }
servers:
  - url: "example.com"
components:
  schemas:
    obj1:
      type: object
      nullable: true
paths: {}
```

**Expected result**

```typescript
schemas: {
  obj1: Record<string, never> | null;
};
```

**Actual Result**

```typescript
schemas: {
  obj1: Record<string, never>;
};
```

**Checklist**

The example given above works fine with `type: string` and `nullable: true`. It also works in `openapi-typescript` version `6.7.6`.

mikebolt:
## Changes

Allow all types except UNKNOWN to be nullable.
Fixes #1821 and #1940 .

**This could be considered a breaking change.** Whenever users are processing a request in a server or processing a response in the client, there could be a type error due to the unhandled null case. However, this is a good change because it helps TypeScript prevent bugs.

## How to Review

The actual code changes are limited to the end of the transformSchemaObjectWithComposition function in schema-object.ts.

I have added a new test containing the schema in issue #1940 . You can check that this output looks good.

I have also changed the expected values in the various github and digitalocean .ts test files (although they were possibly overwritten by the update:examples script). Wherever I added ` | null` to a type, its corresponding schema is nullable. This is something you can double-check, although there are a few hundred cases.

## Checklist

- [ Yes ] Unit tests updated
- [ N/A ] `docs/` updated (if necessary)
- [ Yes ] `pnpm run update:examples` run (only applicable for openapi-typescript)

### DIFF
--- a/packages/openapi-typescript/examples/digital-ocean-api.ts
+++ b/packages/openapi-typescript/examples/digital-ocean-api.ts
@@ -11634,7 +11634,7 @@ export interface components {
              * @description The Droplet that the floating IP has been assigned to. When you query a floating IP, if it is assigned to a Droplet, the entire Droplet object will be returned. If it is not assigned, the value will be null.
              * @example null
              */
-            droplet?: Record<string, never> | components["schemas"]["droplet"];
+            droplet?: (Record<string, never> | null) | components["schemas"]["droplet"];
             /**
              * @description A boolean value indicating whether or not the floating IP has pending actions preventing new ones from being submitted.
              * @example true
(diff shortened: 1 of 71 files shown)


drwpow:
Thanks for adding this! While I agree—this _may_ be breaking, I think it’s a sensible enough change, and people have asked for this, that I think it’s fine to introduce. Be sure to add a `minor` changeset patch so the version gets bumped! (see changeset comment)

The only change I’d request is you have a test for `nullable: true`, but as I understand it, that’s the “old” syntax. And today, `oneOf: [SchemaFoo, { type: "null" }]` is preferred. If we can add tests for that, too, and make sure that works the same way (you’ll see other tests of this nature in there), then I’d be happy to approve & ship.

duncanbeevers:
This PR is superseded by https://github.com/openapi-ts/openapi-typescript/pull/2059

PR #1964, opened 2024-10-25
Chore: Update deps

drwpow:
## Changes

Updates deps to fix artificial security issues

## How to Review

- CI should pass

## Checklist

N/A

### DIFF
--- a/packages/openapi-react-query/package.json
+++ b/packages/openapi-react-query/package.json
@@ -65,17 +65,18 @@
     "openapi-typescript-helpers": "workspace:^"
   },
   "devDependencies": {
+    "@tanstack/react-query": "^5.59.16",
     "@testing-library/react": "^16.0.1",
     "@types/react": "18.3.1",
-    "@vitejs/plugin-react": "^4.3.1",
+    "@vitejs/plugin-react": "^4.3.3",
     "del-cli": "^5.1.0",
     "esbuild": "^0.20.2",
     "execa": "^8.0.1",
-    "msw": "^2.4.1",
+    "msw": "^2.5.1",
     "openapi-fetch": "workspace:^",
     "openapi-typescript": "workspace:^",
     "react": "^18.3.1",
-    "react-error-boundary": "^4.0.13"
+    "react-error-boundary": "^4.1.2"
   },
   "peerDependencies": {
     "@tanstack/react-query": "^5.0.0",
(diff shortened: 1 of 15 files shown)


packages/openapi-react-query/src/index.ts
```diff
@@ -121,7 +121,11 @@ export default function createClient<Paths extends {}, Media extends MediaType =
     useQuery: (method, path, ...[init, options, queryClient]) =>
       useQuery(queryOptions(method, path, init as InitWithUnknowns<typeof init>, options), queryClient),
     useSuspenseQuery: (method, path, ...[init, options, queryClient]) =>
-      useSuspenseQuery(queryOptions(method, path, init as InitWithUnknowns<typeof init>, options), queryClient),
+      useSuspenseQuery(
+        // @ts-expect-error TODO: fix minor type mismatch between useQuery and useSuspenseQuery
```
drwpow, line 125: cc @kerwanp there’s an error on the newer version of `@tanstack/react-query` here, but it looks like a superficial one:

```
Argument of type 'UseQueryOptions<Response["data"], Response["error"], Response["data"], QueryKey<Paths, Method, Path>>' is not assignable to parameter of type 'UseSuspenseQueryOptions<Response["data"], Response["error"], Response["data"], QueryKey<Paths, Method, Path>>'.
  Types of property 'queryFn' are incompatible.
    Type 'unique symbol | QueryFunction<Response["data"], QueryKey<Paths, Method, Path>, never> | undefined' is not assignable to type 'QueryFunction<Response["data"], QueryKey<Paths, Method, Path>, never> | undefined'.
      Type 'typeof skipToken' is not assignable to type 'QueryFunction<Response["data"], QueryKey<Paths, Method, Path>, never>'.ts(2345)
```

I don’t think it actually affects runtime too much. And the solution may be hairy / didn’t want to destroy the inference happening. So just ignored it for now.

packages/openapi-react-query/package.json
```diff
@@ -65,17 +65,18 @@
     "openapi-typescript-helpers": "workspace:^"
   },
   "devDependencies": {
+    "@tanstack/react-query": "^5.59.16",
```
drwpow, line 68: Also we weren’t loading a specific version of `@tanstack/react-query` for tests, etc but we should be. Added as a `devDep` (users still control this package, but we still have to pin a version for our own local tests, etc.)

PR #1970, opened 2024-10-25
feat: narrow type based on status code

DjordyKoert:
## Changes

The various client methods (`GET`, `POST`, `PUT` e.t.c.) now have an additional `status` property. 

This property allows users to narrow the type of their `data`/`error` based on their documented response status codes.

## How to Review

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -114,10 +114,10 @@ export type FetchResponse<
   } & (S extends OkStatus
     ? {
         data: ParseAsResponse<GetResponseContent<ResponseObjectMap<T>, Media, S>, Options>;
-        error: never;
+        error?: never;
       }
     : {
-        data: never;
+        data?: never;
         error: GetResponseContent<ResponseObjectMap<T>, Media, S>;
       });
 }[TStatus];
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -9,6 +9,10 @@ import type {
   ResponseObjectMap,
   RequiredKeysOf,
   SuccessResponse,
+  GetResponseContent,
+  ErrorStatus,
+  OkStatus,
+  OpenApiStatusToHttpStatus,
 } from "openapi-typescript-helpers";
 
 /** Options for each client instance */
@@ -98,17 +102,25 @@ export type RequestBodyOption<T> = OperationRequestBodyContent<T> extends never
 
 export type FetchOptions<T> = RequestOptions<T> & Omit<RequestInit, "body" | "headers">;
 
-export type FetchResponse<T extends Record<string | number, any>, Options, Media extends MediaType> =
-  | {
-      data: ParseAsResponse<SuccessResponse<ResponseObjectMap<T>, Media>, Options>;
-      error?: never;
-      response: Response;
-    }
-  | {
-      data?: never;
-      error: ErrorResponse<ResponseObjectMap<T>, Media>;
-      response: Response;
-    };
+export type FetchResponse<
+  T extends Record<string | number, any>,
+  Options,
+  Media extends MediaType,
+  TStatus extends keyof ResponseObjectMap<T> = keyof ResponseObjectMap<T>,
+> = {
+  [S in TStatus]: {
+    response: Response;
+    status: OpenApiStatusToHttpStatus<S>;
+  } & (S extends OkStatus
+    ? {
+        data: ParseAsResponse<GetResponseContent<ResponseObjectMap<T>, Media, S>, Options>;
+        error: never;
+      }
+    : {
+        data: never;
+        error: GetResponseContent<ResponseObjectMap<T>, Media, S>;
+      });
+}[TStatus];
 
 export type RequestOptions<T> = ParamsOption<T> &
   RequestBodyOption<T> & {
--- a/packages/openapi-fetch/test/common/response.test.ts
+++ b/packages/openapi-fetch/test/common/response.test.ts
@@ -65,7 +65,7 @@ describe("response", () => {
       }
     });
 
-    test("returns union for mismatched errors", async () => {
+    test("returns union for mismatched errors", async () => {
       const client = createObservedClient<paths>();
       const result = await client.GET("/mismatched-errors");
       if (result.data) {
--- a/packages/openapi-typescript-helpers/index.d.ts
+++ b/packages/openapi-typescript-helpers/index.d.ts
@@ -7,6 +7,19 @@ export type OkStatus = 200 | 201 | 202 | 203 | 204 | 206 | 207 | "2XX";
 // biome-ignore format: keep on one line
 export type ErrorStatus = 500 | 501 | 502 | 503 | 504 | 505 | 506 | 507 | 508 | 510 | 511 | '5XX' | 400 | 401 | 402 | 403 | 404 | 405 | 406 | 407 | 408 | 409 | 410 | 411 | 412 | 413 | 414 | 415 | 416 | 417 | 418 | 420 | 421 | 422 | 423 | 424 | 425 | 426 | 427 | 428 | 429 | 430 | 431 | 444 | 450 | 451 | 497 | 498 | 499 | '4XX' | "default";
 
+/**
+ * 'default' returns every status code
+ * '2XX' returns all 2XX status codes
+ * '4XX' returns all 4XX status codes
+ * '5XX' returns all 5XX status codes
+ */
+export type OpenApiStatusToHttpStatus<Status> = 
+  Status extends number ? Status :
+  Status extends "default" ? number :
+  Status extends "2XX" ? Exclude<OkStatus, string> :
+  Status extends "4XX" | "5XX" ? Exclude<ErrorStatus, string> :
+  never;
+
 /** Get a union of OK Statuses */
 export type OKStatusUnion<T> = FilterKeys<T, OkStatus>;
 
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -152,16 +152,16 @@ export default function createClient(clientOptions) {
 
     // handle empty content
     if (response.status === 204 || response.headers.get("Content-Length") === "0") {
-      return response.ok ? { data: undefined, response } : { error: undefined, response };
+      return response.ok ? { data: undefined, response, status: response.status  } : { error: undefined, response, status: response.status };
     }
 
     // parse response (falling back to .text() when necessary)
     if (response.ok) {
       // if "stream", skip parsing entirely
       if (parseAs === "stream") {
-        return { data: response.body, response };
+        return { data: response.body, response, status: response.status };
       }
-      return { data: await response[parseAs](), response };
+      return { data: await response[parseAs](), response, status: response.status };
     }
 
     // handle errors
@@ -171,7 +171,7 @@ export default function createClient(clientOptions) {
     } catch {
       // noop
     }
-    return { error, response };
+    return { error, response, status: response.status };
   }
 
   return {
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -152,7 +152,9 @@ export default function createClient(clientOptions) {
 
     // handle empty content
     if (response.status === 204 || response.headers.get("Content-Length") === "0") {
-      return response.ok ? { data: undefined, response, status: response.status  } : { error: undefined, response, status: response.status };
+      return response.ok
+        ? { data: undefined, response, status: response.status }
+        : { error: undefined, response, status: response.status };
     }
 
     // parse response (falling back to .text() when necessary)
--- a/packages/openapi-fetch/test/never-response/never-response.test.ts
+++ b/packages/openapi-fetch/test/never-response/never-response.test.ts
@@ -140,4 +140,36 @@ describe("GET", () => {
     expect(data).toBeUndefined();
     expect(error).toBe("Unauthorized");
   });
+
+  test("type narrowing on status", async () => {
+    const mockData = {
+      id: 123,
+      title: "My Post",
+    };
+
+    let actualPathname = "";
+    const client = createObservedClient<paths>({}, async (req) => {
+      actualPathname = new URL(req.url).pathname;
+      return Response.json(mockData);
+    });
+
+    const { data, error, response, status } = await client.GET("/posts/{id}", {
+      params: { path: { id: 123 } },
+    });
+
+    if (status === 200) {
+      // @ts-expect-error FIXME: The '200' will never be undefined
+      assertType<typeof mockData>(data);
+    } else if (status === 204) {
+      assertType<undefined>(data);
+    } else if (status === 400) {
+      assertType<components["schemas"]["Error"]>(error);
+    } else if (status === 500) {
+      // @ts-expect-error FIXME: The 'default' response should be excluded
+      assertType<undefined>(error);
+    } else {
+      // @ts-expect-error FIXME: The '500' has already been caught so the undefined should be excluded
+      assertType<components["schemas"]["Error"]>(error);
+    }
+  });
 });


drwpow:
I LOVE this! Please let us know if you need any help with the tests or the current direction. This would be amazing to land

DjordyKoert:
@drwpow I think this is ready for a review now :)

.changeset/rare-grapes-explode.md
```diff
@@ -0,0 +1,7 @@
+---
+"openapi-typescript-helpers": minor
```
drwpow, line 2: Since these are all < 1.0, we want to only bump `minor` versions on breaking changes. We ship bugfixes and features on patches while on 0.x.

```suggestion
"openapi-typescript-helpers": patch
```

.changeset/rare-grapes-explode.md
```diff
@@ -0,0 +1,7 @@
...
+"openapi-react-query": minor
```
drwpow, line 3:
```suggestion
"openapi-react-query": patch
```

.changeset/rare-grapes-explode.md
```diff
@@ -0,0 +1,7 @@
...
+"openapi-fetch": minor
```
drwpow, line 4:
```suggestion
"openapi-fetch": patch
```

docs/openapi-fetch/index.md
```diff
@@ -158,17 +161,18 @@ The `POST()` request required a `body` object that provided all necessary [reque
 
 ### Response
 
-All methods return an object with **data**, **error**, and **response**.
+All methods return an object with **data**, **error**, **status** and **response**.
```
drwpow, line 164: Great docs updates! Thank you

packages/openapi-fetch/test/common/response.test.ts
```diff
@@ -20,22 +20,38 @@ describe("response", () => {
       // 2. assert data is not undefined inside condition block
       if (result.data) {
         assertType<NonNullable<Resource[]>>(result.data);
-        assertType<undefined>(result.error);
+        // @ts-expect-error FIXME: This is a limitation within Typescript
```
drwpow, line 23: I love these tests, and this is a great start! But I fear that we’ll have to fix this before merging (it’ll be much more difficult to fix in a followup, when other work has built off the inference). I don’t really care if the type is `undefined` or `never`—that’s a detail—but TS should be able to infer the other property otherwise there could be issues!

packages/openapi-fetch/test/common/response.test.ts
```diff
@@ -20,22 +20,38 @@ describe("response", () => {
...
       // 2b. inverse should work, too
       if (!result.error) {
         assertType<NonNullable<Resource[]>>(result.data);
-        assertType<undefined>(result.error);
+        assertType<never>(result.error);
+      }
+
+      if (result.status === 200) {
+        assertType<NonNullable<Resource[]>>(result.data);
+        assertType<never>(result.error);
```
drwpow, line 34: These are good tests! I would want to see the other common codes tested here, too—`201`, and `400`–`405`, just for completeness (and if they’re not in the schema, we should add them!)

packages/openapi-fetch/test/common/response.test.ts
```diff
@@ -20,22 +20,38 @@ describe("response", () => {
...
+      }
+
+      if (result.status === 500) {
+        assertType<never>(result.data);
+        assertType<Error>(result.error);
+      }
+
+      // @ts-expect-error 204 is not defined in the schema
+      if (result.status === 204) {
```
drwpow, line 43: This is a REALLY good test I want to keep—testing what happens when a status code is impossible for a given endpoint! But we should test that `204` works somewhere, even if it’s a different test / different endpoint (doesn’t matter where, as long as it’s somewhere, but soft recommendation to just add `204` here since we have all the other checks, and test for the “missing” case in its own test).

drwpow (changes requested):
This is looking great! But the minor type regression to testing `if (result.data)` vs `if (result.error)` could be a breaking change, and could be a type error here. The status narrowing seems to work brilliantly, but we’ll need to make sure users already using this library can still infer types the same way.

If we can find a way to not have the `FIXME` in the type tests, I think we’ve got it.

packages/openapi-fetch/src/index.d.ts
```diff
@@ -98,17 +100,14 @@ export type RequestBodyOption<T> = OperationRequestBodyContent<T> extends never
...
-      data?: never;
-      error: ErrorResponse<ResponseObjectMap<T>, Media>;
-      response: Response;
-    };
+export type FetchResponse<T extends Record<string | number, any>, Options, Media extends MediaType> = {
+  [S in keyof ResponseObjectMap<T>]: {
+    response: Response;
+    status: OpenApiStatusToHttpStatus<S, keyof ResponseObjectMap<T>>;
+    data: S extends OkStatus ? ParseAsResponse<GetResponseContent<ResponseObjectMap<T>, Media, S>, Options> : never;
+    error: S extends ErrorStatus ? GetResponseContent<ResponseObjectMap<T>, Media, S> : never;
```
gzm0, line 108: This is super nice! I like how it removes "specialness" from data versus error.

IIUC, at this point, the only difference between data and error is parseAs. (totally unrelated to this PR).

packages/openapi-typescript-helpers/index.d.ts
```diff
@@ -7,6 +7,22 @@ export type OkStatus = 200 | 201 | 202 | 203 | 204 | 206 | 207 | "2XX";
 // biome-ignore format: keep on one line
 export type ErrorStatus = 500 | 501 | 502 | 503 | 504 | 505 | 506 | 507 | 508 | 510 | 511 | '5XX' | 400 | 401 | 402 | 403 | 404 | 405 | 406 | 407 | 408 | 409 | 410 | 411 | 412 | 413 | 414 | 415 | 416 | 417 | 418 | 420 | 421 | 422 | 423 | 424 | 425 | 426 | 427 | 428 | 429 | 430 | 431 | 444 | 450 | 451 | 497 | 498 | 499 | '4XX' | "default";
 
+/**
+ * 'default' returns every not explicitly defined status code
+ * '2XX' returns all 2XX status codes
+ * '4XX' returns all 4XX status codes
+ * '5XX' returns all 5XX status codes
```
gzm0, line 14: Shouldn't we remove the explicitly defined status codes from the wildcards? From the OpenAPI Spec (https://swagger.io/specification/#responses-object)

> If a response is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code.

So for example, if we have: "2xx" -> `A` and "204" -> `B`, with different, "204" should resolve to `B` and I think currently it would resolve to `A | B`.

packages/openapi-typescript-helpers/index.d.ts
```diff
@@ -7,6 +7,22 @@ export type OkStatus = 200 | 201 | 202 | 203 | 204 | 206 | 207 | "2XX";
...
+ */
+export type OpenApiStatusToHttpStatus<Status, AllStatuses> = Status extends number
+  ? Status
+  : Status extends "default"
+    ? Exclude<Exclude<OkStatus | ErrorStatus, string>, AllStatuses>
+    : Status extends "2XX"
+      ? Exclude<OkStatus, string>
+      : Status extends "4XX" | "5XX"
```
gzm0, line 22: It might help readability if you split `OkStatus` / `ErrorStatus` into numerics / wildcards. For example:

```ts
export type ErrorStatus = NumericErrorStatus | WildcardErrorStatus | "default";
type NumericErrorStatus = 500 | ...
type WildcardErrorStatus = '5XX' | '4XX'; 

export type OpenApiStatusToHttpStatus<Status, AllStatuses> = Status extends number
  ? Status
  : Status extends "default"
    ? Exclude<NumericOKStatus | NumericErrorStatus, AllStatuses>
    : Status extends WildcardOKStatus
      ? NumericOKStatus
      : Status extends WildcardErrorStatus
     ? NumericErrorStatus
     : never
```

(totally optional)

packages/openapi-fetch/test/types.test.ts
```diff
@@ -280,4 +286,179 @@ describe("types", () => {
       assertType<Response>({ error: "default application/json" });
     });
   });
+
+  describe("OpenApiStatusToHttpStatus", () => {
```
gzm0, line 290: It is super nice to see unit tests for this helper!

I was wondering: IIUC the second argument (`AllStatuses`) comes directly from the response object keys. In this case, it could contain wildcards (e.g. `"2XX"`). Would it make sense to add tests for these as well? (I have tried pointing out places below where I think this might affect the tests).

@drwpow self note to maintainers: we should probably set up the infrastructure for `openapi-typescript-helpers` to have tests :)

packages/openapi-fetch/test/types.test.ts
```diff
@@ -280,4 +286,179 @@ describe("types", () => {
...
+      assertType<Status>(
+        // @ts-expect-error 204 has been manually defined
+        204,
+      );
+      assertType<Status>(201);
+      assertType<Status>(504);
+    });
+
+    test("returns 200 likes response", () => {
+      type Status = OpenApiStatusToHttpStatus<"2XX", 200 | 204 | 206 | 404 | 500 | "default">;
```
gzm0, line 323: It feels to me like this test case is "degenerate" at least if I understand the usage in this PR correctly:

I would expect the `Status` parameter to always be a subtype of `AllStatuses` (which in this case it isn't).

For this test case specifically, I would expect that the explicitly / numerically defined status codes are *not* part of the end result (precedence, as I pointed out in the doc comment of `OpenApiStatusToHttpStatus`).

packages/openapi-fetch/test/types.test.ts
```diff
@@ -280,4 +286,179 @@ describe("types", () => {
...
+      assertType<OpenApiStatusToHttpStatus<204, number>>(
+        // @ts-expect-error 200 is not a valid
+        200,
+      );
+
+      assertType<OpenApiStatusToHttpStatus<404, 200 | 204 | 206 | 404 | 500 | "default">>(404);
+    });
+
+    test("returns default response", () => {
+      type Status = OpenApiStatusToHttpStatus<"default", 200 | 204 | 206 | 404 | 500 | "default">;
```
gzm0, line 309: Would it make sense add a test case where there also is a wildcard (e.g. `"2XX"`) in the second argument?

packages/openapi-typescript-helpers/index.d.ts
```diff
@@ -7,6 +7,22 @@ export type OkStatus = 200 | 201 | 202 | 203 | 204 | 206 | 207 | "2XX";
 // biome-ignore format: keep on one line
 export type ErrorStatus = 500 | 501 | 502 | 503 | 504 | 505 | 506 | 507 | 508 | 510 | 511 | '5XX' | 400 | 401 | 402 | 403 | 404 | 405 | 406 | 407 | 408 | 409 | 410 | 411 | 412 | 413 | 414 | 415 | 416 | 417 | 418 | 420 | 421 | 422 | 423 | 424 | 425 | 426 | 427 | 428 | 429 | 430 | 431 | 444 | 450 | 451 | 497 | 498 | 499 | '4XX' | "default";
 
+/**
+ * 'default' returns every not explicitly defined status code
+ * '2XX' returns all 2XX status codes
+ * '4XX' returns all 4XX status codes
+ * '5XX' returns all 5XX status codes
+ */
```
gzm0, line 15: Note: IIUC this implementation has the limitation that status codes unknown to `openapi-fetch` have to be explicitly defined in the OpenAPI spec so they can be typed on the client side.

For example, `"2XX"` in the OpenAPI spec will not match `209`:

```
if (status === 209) {
  // `data` will be never
}
```

IMO this is a reasonable limitation / design decision.

Relevant excerpt from [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-6)

>    HTTP status codes are extensible.  HTTP clients are not required to
   understand the meaning of all registered status codes, though such
   understanding is obviously desirable.  However, a client MUST
   understand the class of any status code, as indicated by the first
   digit, and treat an unrecognized status code as being equivalent to
   the x00 status code of that class, with the exception that a
   recipient MUST NOT cache a response with an unrecognized status code.

>   For example, if an unrecognized status code of 471 is received by a
   client, the client can assume that there was something wrong with its
   request and treat the response as if it had received a 400 (Bad
   Request) status code.  The response message will usually contain a
   representation that explains the status.

gzm0 (commented):
This is super cool!

I've left some comments mostly regarding http status wildcard precedence.

packages/openapi-fetch/src/index.d.ts
DjordyKoert: I already got a PR semi-ready about that https://github.com/openapi-ts/openapi-typescript/pull/1986 :)

packages/openapi-typescript-helpers/index.d.ts
drwpow: Oh that’s a great point, yeah. To be honest I haven’t come across many schemas in the wild actually using the `XX` error codes; this was just an attempt to support the spec in its entirety. Agree that if this causes typescript confusion, I’d err on refactoring this or just taking it out.

packages/openapi-typescript-helpers/index.d.ts
drwpow: Agreed. As stated in the other comment, I’d be OK with doing another pass on the `XX` codes, either removing them or improving inference so that they don’t overlap with specific status codes. But I don’t want to put that on this PR.

drwpow (approved):
Thanks for addressing the comments! This is a really big upgrade, thank you!

Unless I’m missing something, I believe this is a non-breaking change, and ready to release. But will revisit anything remaining in a breaking future change, if necessary (e.g. wildcards)

benevbright:
Hey, thanks. This PR would fix https://github.com/openapi-ts/openapi-typescript/issues/1723 right?

benevbright:
actually, https://github.com/openapi-ts/openapi-typescript/issues/1723 seems have been fixed by https://github.com/openapi-ts/openapi-typescript/pull/1937 already? @drwpow

afarmerdev:
👀

afarmerdev:
Thanks for the good work!
May I ask what the blocker is on this? In most cases, `openapi-react-query` loses much of its usefulness without this feature, since status codes carry important semantic meaning in REST APIs.

In its current form, we can't even distinguish whether an error is a 400 or a 500 when using `openapi-react-query`. For example, we use 402 for validation errors, and we'd really like to have type-safe access to responses based on status codes.

gzm0:
> May I ask what the blocker is on this?

IIUC mainly the fixme in the test as pointed out by @drwpow . Basically, our understanding is that in the current form, this isn't backwards compatible.

IMHO we could discuss a breaking release, but then I'd like to look at:
- removing the error `error` return value
- potentially replacing the openapi-fetch specific return object with a native fetch response with narrower types (would imply the call sites have to call `json()`).

edwardmp:
this is a much needed feature :)  @gzm0 have you decided whether a breaking release is the way to go here?

afarmerdev:
> this is a much needed feature :) @gzm0 have you decided whether a breaking release is the way to go here?

We ended up using our own [modified version of `openapi-react-query`](https://gist.github.com/afarmerdev/2d2e7d3956d358da4f5bc4e21fd70c8a#file-openapi-react-query-ts-L218) which enriches the `error` type with `responseStatus` and `responseBody`. Unfortunately, `responseBody` is a union of all error body schemas rather than a specific one, but I believe this can be improved with some TypeScript judo. 

@gzm0, if you're considering a breaking release, you might want to update the `error` schema as we did, instead of altering the idiomatic usage of react-query. Since this package handles HTTP calls, it makes perfect sense to include the status code and the corresponding error body object within the error type.

I am imaging a usage like this

```
  const { mutate, isError, error } = $api.useMutation(
    'post',
    '/v1/users',
    {}
  );

  if (error?.responseStatus === 400) {
    const { error_message } = error.responseBody;
    return <div>{error_message}</div>;
  }

  if (error?.responseStatus === 402) {
    // validation error
    const { validationErrors } = error.responseBody;
    return (
      <div>
        Please fill out the fields corrrectly{' '}
        {validationErrors.map((error) => {
          return (
            <div>
              {error.field}:{error.message}
            </div>
          );
        })}
      </div>
    );
  }

  if (isError) {
    //at this point it is either 500 or non-fetch related error
    error.cause && logger(error.cause); // log non-fetch errror
    return <div>Something went wrong</div>;
  }
```

And please don't go this path for discriminating fetch errors from non-fetch errors(which are very rare in our usage, perhaps should be handled centrally) I think ergonomics of testing error status right away without a step like this is better:
```
   if (error instanceof FetchError) { // please let's not do this
       if (error.responseStatus == 400)
          ...
    }
```

we can consider renaming `error.responseStatus` to `error.statusCode` and `error.responseBody` to `error.body`

gzm0:
@edwardmp @afarmerdev thanks for keeping the ball rolling on this one.

Overall, I think I'm in favor of a breaking change to get this moving forward. I've tried to write down my thoughts below. LMK what you think.

# Requirements

There are a couple of things I'd like us to look out for:

- Keep an easy way to distinguish (HTTP) success from error (i.e. an equivalent of [`Response.ok`](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok)).
- Make sure the breaking change is not silent on usage sites during typechecking: i.e. usage sites should fail to typecheck and not just do something different.
- Nice to have: work towards `data` and `error` not being different.
- Nice to have: even the path to return a fetch response directly.

# Potential solution

A way I think this could work ([TS playground with example](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgLIE8DyAjAVsgbwChlkAPALmRAFcBbbaIgXyKNElkRQFEooA9lAAiEMHGAAbQiVLJ0VWg2gBuFmzDoADigwAlCAGctAkIZQBeZEQA+MucgEBrKmCg0IAGlmlD4sDSGVABMAAyhnsgA9FGOTj7I2AIAJgpoWHjqdsQOzq7uXgl+cAFByGEAjJExyAhQECUQyQlJqVQYOLjezNbZCXnIbh7eDsWlIaEALNWxIAK1ppDgLSlpNCDJEDCgTd229nID8JLmI3JjgVSTwcFnpK1pfIIiYhKSe325LsjHp0X+l2Qkymd0Sqyo6022xATRYaiICFMfmQAAs4BtJBBMABpZBWAAU9WM7XQBmMSIgAEo8QA+A6IsxgQhxSIPZA9KxErRqUjAGDIfHOak5ORswwlYCGbZGdKdZB2SFbHbJHns5AQE4oEX3VbIcVgSXSwzIJ5CUTiKTy5CK6FNVWsVgIpFMtEYiDmt5NPECrkkskmMxU2n053M5yRC6GVm6jnILnw3n8-GRvEWKzXYLClapPUSqXAGWml4WyT29Wa5B8gUptNWMKTLMOMV5o3WjZKmEq2Q9DXmStJoUHHU5-WGgvGjp4Mu9rUJGoAcwgMKgwAQ6v4QlR6OSklA84AdN2WEA)) is as follows:

- Replace the `data` and `error` fields with a new `body` field, use it to return the parsed response body (do not change parsing behavior for now).
- Add an `ok` field next to the existing `status` field.

I think this approach would satisfy 3 of the 4 requirements above: it does not satisfy "even the path to return a fetch response directly": `body` is the raw response. But maybe that's OK and we can deal with this in a future breaking change[^1].


[^1]: I realize splitting into multiple breaking changes is not ideal from a users POV, but given the limited resources, I think it makes sense to focus on the status improvement first. Of course, I'm more than happy to discuss other opinions.

# Alternative

An alternative I have considered is to keep `data` and also use it to return errors. My main concern with that approach is that it might be difficult to debug on upgrade for users:

```ts
const { data, error } = await client.GET("/blogposts");

if (data) {
  // do stuff
} else {
  // handle error
}
```

If we re-use `data` both typechecking (and execution) will always take the "do stuff" branch. So whether or not typechecking fails will depend on what "do stuff" is.

PR #1975, opened 2024-10-28
Fix invalid data check in query/mutation fn

### LINKED ISSUE #1973: openapi-react-query with null values
**Description**

openapi-react-query crashes if endpoint returns nulls.

**Reproduction**

any endpoint which returns null with useQuery exposed by this library

**Expected result**

null is both valid json and react-query allows to return it form queryFn
https://github.com/openapi-ts/openapi-typescript/blob/781cf926288d6f735096437b529cc20d51142ccf/packages/openapi-react-query/src/index.ts#L144
This line causes the issue - `!data` should be replaced with `data!==undefined`

**Checklist**

HugeLetters:
## Changes

Closes #1973

## How to Review

-

## Checklist

- [x] Unit tests updated
- [x] ~~`docs/` updated (if necessary)~~
- [ ] ~~`pnpm run update:examples` run (only applicable for openapi-typescript)~~

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,7 +8,7 @@
   "[markdown]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
-  "[typescript]": {
+  "[typescript][typescriptreact]": {
     "editor.defaultFormatter": "biomejs.biome"
   }
 }
--- a/packages/openapi-react-query/src/index.ts
+++ b/packages/openapi-react-query/src/index.ts
@@ -115,7 +115,7 @@ export default function createClient<Paths extends {}, Media extends MediaType =
     const mth = method.toUpperCase() as Uppercase<typeof method>;
     const fn = client[mth] as ClientMethod<Paths, typeof method, Media>;
     const { data, error } = await fn(path, { signal, ...(init as any) }); // TODO: find a way to avoid as any
-    if (error || !data) {
+    if (error || data === undefined) {
       throw error;
     }
     return data;
@@ -141,7 +141,7 @@ export default function createClient<Paths extends {}, Media extends MediaType =
             const mth = method.toUpperCase() as Uppercase<typeof method>;
             const fn = client[mth] as ClientMethod<Paths, typeof method, Media>;
             const { data, error } = await fn(path, init as InitWithUnknowns<typeof init>);
-            if (error || !data) {
+            if (error || data === undefined) {
               throw error;
             }
             return data;
--- a/packages/openapi-react-query/src/index.ts
+++ b/packages/openapi-react-query/src/index.ts
@@ -115,9 +115,10 @@ export default function createClient<Paths extends {}, Media extends MediaType =
     const mth = method.toUpperCase() as Uppercase<typeof method>;
     const fn = client[mth] as ClientMethod<Paths, typeof method, Media>;
     const { data, error } = await fn(path, { signal, ...(init as any) }); // TODO: find a way to avoid as any
-    if (error || data === undefined) {
+    if (error) {
       throw error;
     }
+
     return data;
   };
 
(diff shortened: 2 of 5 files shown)


kerwanp (changes requested):
Thanks, could you please provide a test case as well as a patch commit?

HugeLetters:
After working on tests it got me thinking - should we even check if data is undefined?
Perhaps just an `if(error)` check would be sufficient?

Because the problem is
- if we throw when data is undefined, we just throw undefined as error(cause error is undefined too)
- react query already handles returning undefined as an error and will throw accordingly - perhaps it should just remain their responsibility

kerwanp:
I do not exactly remember why we checked for that data in the first place. But maybe yes we could just remove this check as we need to handle responses with no content anyway (see https://github.com/openapi-ts/openapi-typescript/issues/1973#issuecomment-2445815605).

.vscode/settings.json
```diff
@@ -8,7 +8,7 @@
   "[markdown]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
-  "[typescript]": {
```
HugeLetters, line 11: Since tests are written in .tsx files I had to add this for formatting to work

packages/openapi-react-query/src/index.ts
```diff
@@ -115,9 +115,10 @@ export default function createClient<Paths extends {}, Media extends MediaType =
     const mth = method.toUpperCase() as Uppercase<typeof method>;
     const fn = client[mth] as ClientMethod<Paths, typeof method, Media>;
     const { data, error } = await fn(path, { signal, ...(init as any) }); // TODO: find a way to avoid as any
-    if (error || !data) {
```
HugeLetters, line 118: I think it's best we don't check for data at all. 
If it's undefined - react-query will throw an error for us. Other falsy values are fine

packages/openapi-react-query/src/index.ts
```diff
@@ -141,9 +142,14 @@ export default function createClient<Paths extends {}, Media extends MediaType =
             const mth = method.toUpperCase() as Uppercase<typeof method>;
             const fn = client[mth] as ClientMethod<Paths, typeof method, Media>;
             const { data, error } = await fn(path, init as InitWithUnknowns<typeof init>);
-            if (error || !data) {
+            if (error) {
               throw error;
             }
+
+            if (data === undefined) {
```
HugeLetters, line 149: Since mutationFn type doesn't allow returning undefined I had to resort to this - this changes return type a little bit but should be fine since data should only be a valid JSON and undefined is not valid. 

React-query does allow returning undefined from mutationFn tho - that restriction is only for queries. 

But the problem here is at the type level - data is basically `T | undefined` but mutationFn basically has to return `T`. 
- If we don't do an undefined check - we're returning `undefined` which not part of `T`. 
- If do an undefined check - what if `T` was instantiated as `number | undefined`. Then we're excluding `undefined` from the return type even if it part of `T` type. 

We don't know at runtime if data is undefined because it `fn` errored or because that's what data is - but since it should be impossible to return undefined that's why I went with the 2nd option

.vscode/settings.json
```diff
@@ -8,7 +8,7 @@
...
+  "[typescript][typescriptreact]": {
```
HugeLetters, line 11: Since tests are written in .tsx files I had to add this for formatting to work

kerwanp (approved):
I'm ok with the implementation, thanks for the work. Before we merge can you just double check that we cant return undefined because of tanstack query and not because our types are wrong?

packages/openapi-react-query/src/index.ts
HugeLetters: `data` is `T | undefined` but the return type says we can only return `T` - so we somehow need to exclude undefined from `data` even though `T` itself can containe undefined. It can be tricky to do so - I've assumed before that T can only be valid JSON, thus can't in practice be undefined but it seems that's not true.

packages/openapi-react-query/src/index.ts
```diff
@@ -141,10 +142,11 @@ export default function createClient<Paths extends {}, Media extends MediaType =
             const mth = method.toUpperCase() as Uppercase<typeof method>;
             const fn = client[mth] as ClientMethod<Paths, typeof method, Media>;
             const { data, error } = await fn(path, init as InitWithUnknowns<typeof init>);
-            if (error || !data) {
+            if (error) {
               throw error;
             }
-            return data;
+
+            return data as Exclude<typeof data, undefined>;
```
HugeLetters, line 149: shouldn't matter since the type is dictated by return type of `createClientz anyway, its not inferred from here - just checked that what we return is valid

kerwanp (approved):
Looks good to me! Good series of tests as well, thanks for taking the time.

PR #1976, opened 2024-10-30
Fix middleware docs

jo-m:
## Changes

Fix throwing middleware example in docs

Fixes https://github.com/openapi-ts/openapi-typescript/issues/1954

## How to Review

Doc change only.

## Checklist

- [ ] Unit tests updated

### DIFF
--- a/.changeset/.changeset/silver-wasps-relate.md
+++ b/.changeset/.changeset/silver-wasps-relate.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Fix throwing middleware example in docs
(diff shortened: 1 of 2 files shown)


.changeset/.changeset/silver-wasps-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Fix throwing middleware example in docs
```
drwpow, line 5: Thanks for the changeset, but it’s not needed for docs updates! This bumps the version on npm which we don’t need to do here

jo-m: removed.

Issue #1977, opened 2024-10-30
`Unexpected end of JSON input` with empty response


SandroMaglione:
**Description**

When a response has an empty body, the client fails with a `JSON.parse` error:

```sh
SyntaxError: Unexpected end of JSON input
      at JSON.parse (<anonymous>)
      at parseJSONFromBytes (node:internal/deps/undici/undici:4747:19)
      at successSteps (node:internal/deps/undici/undici:4718:27)
      at specConsumeBody (node:internal/deps/undici/undici:4724:9)
      at HttpResponse.json (node:internal/deps/undici/undici:4614:18)
```

This error happens [here](https://github.com/openapi-ts/openapi-typescript/blob/781cf926288d6f735096437b529cc20d51142ccf/packages/openapi-fetch/src/index.js#L164). `response[parseAs]()` fails to convert the empty response to json and crashes, never reaching [the fallback](https://github.com/openapi-ts/openapi-typescript/blob/781cf926288d6f735096437b529cc20d51142ccf/packages/openapi-fetch/src/index.js#L168C5-L168C39).

**Expected result**

The client should be able to handle empty responses without crashing with `JSON.parse` errors.

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

ViktorPontinen:
We've had this issue as well. Generally you want your API to return 204 status if there is no body, which would make open-api skip the json parse and avoid crash.

In our case, we had PUT requests that returned empty body by design, and returned 201 status. For those requests you can add `parseAs: 'stream'` option and it should skip the json parse as well.

For cases where you sometimes get response, I have no idea how to resolve.

srbarba:
Opened a PR to solve this issue: https://github.com/openapi-ts/openapi-typescript/pull/2508

PR #1986, opened 2024-11-05
fix(#1977): refactor handling for empty responses

DjordyKoert:
## Changes

This PR includes a couple of different changes for `openapi-fetch`:

- A refactor (Breaking change?) to ensure the behaviour of `data` and `error` are the same to reduce complexity.
As an example, this means that whenever a user uses the `parseAs: 'stream'` both the `data` and `error` property will return a `ReadableStream` from `response.body`
- The logic for empty response bodies has been changed:
  - It no longer explicity checks for `204` status to determine if a response is empty
  - It now follows[ the spec](https://developer.mozilla.org/en-US/docs/Web/API/Response/body) by checking if `response.body` is null
    - Backwards compatibality with browsers: empty string (`''`) get converted to `undefined` after failing to parse it as `json` (fixes https://github.com/openapi-ts/openapi-typescript/issues/1977)

## How to Review

See changes & tests

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/great-zoos-live.md
+++ b/.changeset/great-zoos-live.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+fixed `parseAs` behaviour being different for error responses
--- a/.changeset/thirty-singers-join.md
+++ b/.changeset/thirty-singers-join.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+fixed empty responses body causing error `Unexpected end of JSON input`
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -106,7 +106,7 @@ export type FetchResponse<T extends Record<string | number, any>, Options, Media
     }
   | {
       data?: never;
-      error: ErrorResponse<ResponseObjectMap<T>, Media>;
+      error: ParseAsResponse<ErrorResponse<ResponseObjectMap<T>, Media>, Options>;
       response: Response;
     };
 
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -150,28 +150,33 @@ export default function createClient(clientOptions) {
       }
     }
 
+    const resultKey = response.ok ? "data" : "error";
+
     // handle empty content
     if (response.status === 204 || response.headers.get("Content-Length") === "0") {
-      return response.ok ? { data: undefined, response } : { error: undefined, response };
+      return { [resultKey]: undefined, response };
+    }
+
+    if (parseAs === "stream") {
+      return { [resultKey]: response.body, response };
     }
 
     // parse response (falling back to .text() when necessary)
-    if (response.ok) {
-      // if "stream", skip parsing entirely
-      if (parseAs === "stream") {
-        return { data: response.body, response };
-      }
-      return { data: await response[parseAs](), response };
-    }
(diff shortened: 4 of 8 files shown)


drwpow:
I think this is a breaking change, but one most seem to be in favor of. Love the direction of this so far! Please let us know if you need any assistance with passing tests, or have any questions.

gzm0:
Looking at #1970, I have been thinking about the "parse-as" problem a bit as well.

I'm wondering if it would make sense to make the parsing lazier. One issue I see with the current API is that the caller cannot look at the status code to decide on how to parse the response. I'm not sure this is very ergonomic.

For example, it is very possible that an API returns a binary stream on 200, but a JSON response on other error codes. In this case, being able to specify `parseAs` only individually might not be sufficient.

One way to work around this problem is to provide accessors:

```ts
const resp = client.GET("/thing");

if (resp.status == 200) {
  resp.stream()
} else {
  resp.json()
}
```

This is somewhat inspired by [requests](https://pypi.org/project/requests/).

I realize 
- this wouldbe a major change
- we need to be careful about the performance implications of the potentially additional wrapper we'd need

However, I fear if we go down the "pass parseAs parameter path", we'll get more and more complicated options (e.g. parseAs per response code, etc.).

drwpow:
@gzm0 I completely agree—`parseAs` is not an ideal solution; it’s just a bandaid to not introduce performance issues. Many people find it hard to use, and want it to be more automatic.

This is related to some longer conversations we maintainers have had where some part of openapi-fetch’s runtime is code-generated from peoples’ specs—that would remove the need for parseAs in exchange for bigger client weight and slightly-more memory. That’s of course a completely separate conversation, but not entirely unrelated. I do think that decision of “does openapi-fetch continue on its current path of requiring zero codegen beyond openapi-typescript” will likely affect the decisions on the `parseAs` API. So we should probably set that in stone once and for all to help guide this decision better.

gzm0:
> I do think that decision of “does openapi-fetch continue on its current path of requiring zero codegen beyond openapi-typescript” will likely affect the decisions on the parseAs API. So we should probably set that in stone once and for all to help guide this decision better.

Agreed.

FWIW: Have we considered simply returning the `fetch` `Response` object with a more narrowed type directly?

That would allow something like:

```ts
const response = await client.GET("/path");

if (!response.ok) { // type narrowing on both `ok` and `status` might be tricky but probably necessary for ergonomics.
  throw new Error("foo");
}

const data = await response.json();
```

While all the lower-level body retrieval methods of `Response` remain available.

drwpow:
> Have we considered simply returning the fetch Response object with a more narrowed type directly?

That was one of the earliest decisions I made to just save hundreds of `.json()`s sprinkling peoples’ code just for the few times when you needed something else (with the assumption OpenAPI is 99% JSON-centric). I’m open to an argument to have a lower-level API if we feel it’s a more cohesive design. I think this would be worth an RFC to get feedback from everyone if that’s a solution to these problems (which I think it would be)

gzm0:
> That was one of the earliest decisions I made to just save hundreds of .json()s sprinkling peoples’ code just for the few times when you needed something else (with the assumption OpenAPI is 99% JSON-centric).

Yeah, that's fair.

>  I’m open to an argument to have a lower-level API if we feel it’s a more cohesive design. I think this would be worth an RFC to get feedback from everyone if that’s a solution to these problems (which I think it would be)

Fair enough. So it really seems this warrants wider discussion.

PR #2027, opened 2024-12-02
Add Turborepo

### LINKED ISSUE #2016: (Chore) Add Turborepo
**Description**

As we’re running a TS monorepo, using [Turborepo](https://turbo.build/repo/docs) would help speed up local dev and simplify configuration

**Proposal**

We don’t have a ton of packages, but even understanding the build order from openapi-typescript → openapi-fetch → openapi-react-query can be an annoyance. pnpm does handle some of this, but a simple version. But it can’t handle, say, testing, where running `pnpm run test` knows that packages need to be built (or rebuilt). At best you end up with complex configuration; at worst, a small change breaks everything

Turborepo is an elegant, drop-in solution for what we have already. It doesn’t change how packages are built or tested. It only reduces misconfigurations and mistakes when working locally. Unlike some other build systems, Turborepo doesn’t tell you how your packages should or shouldn’t be built, and can be added without disrupting anything. It can also be easily removed if we don’t like it (no lock in). For all these reasons, I’ve really enjoyed projects that use it, and would like to add it here.

Turborepo also has configurable caching. That’s a secondary benefit to the real advantage—simpler cross-package dependency management (it “just works”). But in case the caching causes any undesirable results, it can be tuned down or opted out of, which is nice.

There are really no downsides with Turborepo; this just simplifies monorepo configuration.

**Alternatives**

Nx and Bazel are the two other popular build systems. Nx I’ve had bad experiences with (the plugins are invasive, and each one comes with its own set of debugging woes—nice when you get it running but the setup can be painful). Bazel is a nonstarter because it’s completely incompatible with JS. I’m not aware of any other alternatives.

**Checklist**

- [x] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

drwpow:
## Changes

Fixes #2016 by adding Turborepo. This is merely a helper to reduce complexity in the monorepo, e.g. remembering to run `pnpm run build` in the right packages before running `pnpm test`. I like adopting build tooling minimally in that it can be “opted out” from the people that don’t want to deal with it, and that’s why it’s only in the workspace root (for now).

Ideally no one notices it’s there, because it just automatically sets up and installs with `pnpm i` with no changes. It just reduces footguns especially as one optimizes it.

## How to Review

- CI should pass

## Checklist

- [x] Unit tests updated


.github/workflows/ci.yml
```diff
@@ -6,6 +6,9 @@ on:
       - main
   pull_request:
 
+env:
+  GITHUB_TOKEN: ${{ secrets.OPENAPI_TS_BOT_GITHUB_TOKEN }}
```
drwpow, line 10: Note: this was added because Turborepo adds a “shield” around env vars and doesn’t accidentally expose something sensitive, so you declare what to expose. But the flipside is it throws an error when this _isn’t_ set. I may be missing a configuration setting, but until then, this was simpler.

turbo.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "globalEnv": ["GITHUB_TOKEN"],
+  "tasks": {
+    "build": {
+      "dependsOn": ["^build"]
```
drwpow, line 6: Recommended to read the [Turborepo docs](https://turbo.build/repo/docs/reference/configuration), but the `^` character means “it has no dependencies.” Why they did this, and not an empty array, I do not know. But that’s the way it is.

turbo.json
```diff
@@ -0,0 +1,25 @@
...
+    },
+    "lint": {
+      "inputs": ["**/*.ts", "**/*.tsx", "**/*.js", "**/*.css", "**/biome.json"],
+      "dependsOn": ["build"]
+    },
+    "size-limit": {
+      "dependsOn": ["build"]
+    },
+    "test": {
+      "dependsOn": ["build"]
```
drwpow, line 19: This is something cool—here’s where we depend on another task, so it will get fancy here and **only run the minimal build commands needed.** So now, you don’t need to remember to run `pnpm run build && pnpm test`; you can just let Turborepo handle it for you.

PR #2043, opened 2024-12-05
feat: explicitly export root types

### LINKED ISSUE #2033: Export types/interfaces instead of inlining objects
**Description**

Right now, `openapi-typescript` returns inlined objects for `components`. For example, the `github` api has an operation for getting the readme:

```
'repos/get-readme': {
  responses: {
    /** @description Response */
      200: {
        headers: {
          [name: string]: unknown
        }
        content: {
          'application/json': components['schemas']['content-file']
        }
      }
    }
  }
}
```

(note: I’ve left out `parameters` for brevity)

Here, this `operation` references `components['schemas']['content-file']`, which is correct, but `'content-file'` is an _inline object_ on `components['schemas']`:

```
/**
 * Content File
 * @description Content File
 */
'content-file': {
    type: 'file'
    encoding: string
    size: number
    name: string
}
```

(note: I’ve only shown some properties for brevity)

** Why is this a problem **

We are working in a composite TypeScript monorepo with `declarationEmit`. That means each package in he monorepo will create a `.d.ts` file of its exported content for other packages to consume (instead of consuming the source code directly). This greatly speeds-up type performance in large monorepos.

The problem is that, because types are inlined here, whenever we use a function that returns something from our contracts (e.g. with `openapi-fetch`), the inferred type is also _inlined_.

This can create _huge_ declaration files, and we’ve already hit the limit, where TypeScript errors out with:

```
TS7056: The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.
```

Some of our endpoint responses are inlined multiple times in the emitted declaration, which is the main cause of reaching the maximum length.

**Potential solution**

TypeScript will correctly keep import references when packages have their own exported types or interfaces. I was hoping that the `--root-types` config option would achieve exactly that, but it does not. All it does is create an extra, internally unused type that is in itself, a reference:

```
export type SchemaContentFile = components['schemas']['content-file'];
```

**Proposal**

I think a good solution would be to change how `--root-types` works, in a way that it doesn’t just reference `components['schemas']`, but that it _becomes_ the definition for it. In the above example, if we have:

```
export type SchemaContentFile = {
    type: 'file'
    encoding: string
    size: number
    name: string
};
```

and then use this internally for `components['schemas']['content-file']`:

```
/**
 * Content File
 * @description Content File
 */
'content-file': SchemaContentFile
```

I think it would work the same as before for external consumers of `SchemaContentFile`, but it would allow `openapi-fetch` to infer the response of a get request towards `repos/get-readme` to be inferred as a reference to `SchemaContentFile` rather than an inline type.

Note that `operations` doesn’t need to change at all - it can still just refer to `components['schemas']['content-file']`, because that would then be a reference to `SchemaContentFile`.

Note: It doesn’t matter if those are exported as `interface` or `type` - TypeScript will keep importing the root type in both ways. 

**Checklist**

We are currently in the process of trying out to change `--root-types` in the proposed direction for us. If you agree on that improvement, we’ll gladly commit it upstream.

- [x] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

TkDodo:
## Changes

This PR implements the changes proposed in:
- #2033

Instead of inlining objects into `components` and exporting a type alias, we are now exporting the inlined object (technically still a type alias) and reference it from `components`.

This should be a backwards compatible change, but if you prefer, we could use a new cli arg too.

fixes #2033 

## How to Review

## Checklist

I don’t think a docs update is necessary as this PR only changes an implementation detail on how types are exported; it should be non-breaking for users.

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/src/transform/components-object.ts
+++ b/packages/openapi-typescript/src/transform/components-object.ts
@@ -71,30 +71,29 @@ export default function transformComponentsObject(componentsObject: ComponentsOb
             aliasName = aliasName.replace(componentKey, "");
           }
           const typeAlias = ts.factory.createTypeAliasDeclaration(
-              /* modifiers      */ tsModifiers({ export: true }),
-              /* name          */ aliasName,
-              /* typeParameters */ undefined,
-              subType,
+            /* modifiers      */ tsModifiers({ export: true }),
+            /* name          */ aliasName,
+            /* typeParameters */ undefined,
+            subType,
           );
 
           rootTypeAliases[aliasName] = typeAlias;
 
           const property = ts.factory.createPropertySignature(
-              /* modifiers     */ tsModifiers({ readonly: ctx.immutable }),
-              /* name          */ tsPropertyIndex(name),
-              /* typeParameters */ undefined,
-              /* type          */ ts.factory.createTypeReferenceNode(aliasName),
+            /* modifiers     */ tsModifiers({ readonly: ctx.immutable }),
+            /* name          */ tsPropertyIndex(name),
+            /* typeParameters */ undefined,
+            /* type          */ ts.factory.createTypeReferenceNode(aliasName),
           );
 
           addJSDocComment(item as unknown as any, property);
           items.push(property);
-
         } else {
           const property = ts.factory.createPropertySignature(
-              /* modifiers     */ tsModifiers({ readonly: ctx.immutable }),
-              /* name          */ tsPropertyIndex(name),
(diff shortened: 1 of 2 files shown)


drwpow:
Thanks for putting in the work for opening this PR! But this heads in a direction that I strongly believe will cause breakages from people, [partially explained in this comment](https://github.com/openapi-ts/openapi-typescript/issues/2033#issuecomment-2532132019).

PR #2049, opened 2024-12-13
Add schema to postTransform options

### LINKED ISSUE #2013: postTransform would be more useful with schema information
**Description**

The `postTransform` option is handed the ts types as they're generated from each OpenAPI schema member in `transformSchemaObject`.
However, although there is a 1:1 mapping between calls to the `postTransform`, and OpenAPI schema members, `postTransform` has no access to the OpenAPI schema object associated with the generated ts node.

I think `postTransform` would be more useful if it gave access to the schema node, and it should be added to the `options` argument, passed to `postTransform`, along with `path` and `ctx`.

**Checklist**

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

duncanbeevers:
Closes #2013

## Changes

* [x] Add `schema` to `postTransform` options
* [x] Add documentation about `postTransform` options
* [x] Add specs / example using the `schema` option to declare a string Set

## How to Review

* Make sure `TransformNodeOptions` is extension is acceptable.
* Decide whether `schema` should be populated for `transform` options; it's redundant since the first argument to transform is already the schema node, but it introduces consistency between the `transform` and `postTrannsform` APIs that I think might be desirable.

## Checklist

- [x] Unit tests updated
- [ ] `pnpm run update:examples` run (only applicable for openapi-typescript) (I tried this and got swamped with changes, please advise!)

### DIFF
--- a/.changeset/four-worms-leave.md
+++ b/.changeset/four-worms-leave.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Add schema to postTransform options
--- a/docs/6.x/node.md
+++ b/docs/6.x/node.md
@@ -122,6 +122,16 @@ Resultant diff with correctly-typed `file` property:
 +    file?: Blob;
 ```
 
-Any [Schema Object](https://spec.openapis.org/oas/latest.html#schema-object) present in your schema will be run through this formatter (even remote ones!). Also be sure to check the `metadata` parameter for additional context that may be helpful.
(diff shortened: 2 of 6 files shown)


drwpow (approved):
Great code, great docs, and great tests! I don’t have any feedback

PR #2051, opened 2024-12-13
Fix colliding enumValues exports

duncanbeevers:
## 🗣️ Discussion

This work stemmed from an error encountered when using `enumValues: true`, and escalated from there.

### 🐞 The bugs

* Types of exported literals are non-exhaustive.
* Composing OpenAPI enums generates colliding variable names.

#### 💨 Non-exhaustive types

The existing values exported when using `enumValues` express their type relative to the union defined in the TS schema.

```ts
// Before
export interface components {
  schemas: {
    /** @enum {string} */
    Status: "active" | "inactive";
  };
}
export const statusValues: ReadonlyArray<components["schemas"]["Status"]> = ["active", "inactive"];
```

Instead, if we start with an exhaustive concrete value, the union type can be derived.

```ts
// After
export const StatusValues = ["active", "inactive"] as const;
export type Status = (typeof StatusValues)[number];
export interface components {
  schemas: {
    Status: Status;
  }
}
```

#### ➕ Composing OpenAPI enums generates conflicting variable names

Given the following schema composing two enum schemas, the following bad code is generated.

```yaml
components:
  schemas:
    Status:
      anyOf:
        - type: string
          enum:
            - active
            - inactive
        - type: string
          enum:
            - appealed
```

```ts
export const statusValues: components["schemas"]["Status"] = ["active", "inactive"];
export const statusValues: components["schemas"]["Status"] = ["appealed"];
```

By unifying the enum member and literal string traversal + generation paths, I was able to generate non-conflicting export names.

## Changes

- [x] Preserve capitalization of named, exported enum values.
- [x] Type exported const values as const, instead of their location within the operations or components schemas.
- [x] Derive and export types for enum values from concrete values in const arrays.
- [x] Use derived enum value types in operations and components schemas.
- [x] Use non-conflicting variable names for composed OpenAPI enums (anyOf: [enum1, enum2])
- [x] Export a [type-predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) function for each enum constant.

## How to Review

- [ ] There's an unhealthy behavior around composition + `enum: true`. Since enums aren't composable, we still generate nonsense code like `ModeratedStatus: components["schemas"]["Status"] | ModeratedStatusAnyOf1;` (see the node-utils `options > enum` case for this example. This isn't resolved by this PR, and may be entirely unresolvable.
- [ ] Helpers and consts and such are now prepended before the primary root exports; this is necessary since some schema are now derived from these const values.
- [ ] Exported const values are no longer lower-cased. This eliminated some special-case code in a couple of spots, and allowed for exporting `sanitizeMemberName` and using it in `transformers/schema-object`. That alone makes this a breaking change.
- [ ] The `as const` style literals may not be for everyone, though I've found them to be succinct, reliable, and versatile for composition (literal and type-level) and iteration.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/four-worms-leave.md
+++ b/.changeset/four-worms-leave.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Add schema to postTransform options
--- a/.changeset/four-worms-leave.md
+++ b/.changeset/four-worms-leave.md
@@ -1,5 +1,0 @@
----
-"openapi-typescript": patch
----
-
-Add schema to postTransform options
--- a/.changeset/shaggy-impalas-develop.md
+++ b/.changeset/shaggy-impalas-develop.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": major
+---
+
+Export enumValues as const arrays. Derive schema types from literal values.
--- a/.github/ISSUE_TEMPLATE/01_openapi-typescript-bug.yml
+++ b/.github/ISSUE_TEMPLATE/01_openapi-typescript-bug.yml
@@ -6,18 +6,21 @@ labels:
   - triage
 body:
   - type: input
+    id: version
     attributes:
       label: openapi-typescript version
       placeholder: x.x.x
     validations:
       required: true
   - type: input
+    id: node
     attributes:
       label: Node.js version
       placeholder: 20.x.x
     validations:
       required: true
   - type: input
+    id: os
     attributes:
       label: OS + version
       placeholder: macOS 15.1.1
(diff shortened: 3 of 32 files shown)


drwpow (approved):
Same as the other PR—would love to use this as the base for 8.x. Thank you for adding such a big improvement to the library!

This can merge into 8.x with the minor conflict resolved (apologies if I merged them out-of-order)

duncanbeevers:
I rebased this off of `main`, then off of `8.x`, so it might be a little squirrely, but I think I got all the right commits in there.

drwpow:
> I rebased this off of `main`, then off of `8.x`, so it might be a little squirrely, but I think I got all the right commits in there.

Ah yeah no worries. 8.x may diverge from `main` for a little bit and that’s OK. But also being a major release it’s expected. Thanks again!

PR #2058, opened 2024-12-19
Docs: fix build

drwpow:
## Changes

Adds a `pnpm run -w build` to build all packages for docs.

## How to Review

- CI should pass

## Checklist

N/A

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -4,13 +4,14 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "build-docs": "pnpm run update-contributors && vitepress build && cp _redirects .vitepress/dist",
+    "build-docs": "pnpm run update-contributors && pnpm run -w build && vitepress build && cp _redirects .vitepress/dist",
     "dev": "vitepress dev",
     "update-contributors": "node scripts/update-contributors.js"
   },
   "devDependencies": {
-    "vitepress": "1.5.0",
-    "@shikijs/vitepress-twoslash": "^1.22.2",
-    "openapi-metadata": "workspace:*"
+    "@shikijs/vitepress-twoslash": "^1.24.3",
+    "openapi-metadata": "workspace:*",
+    "vite": "^6.0.4",
+    "vitepress": "1.5.0"
   }
 }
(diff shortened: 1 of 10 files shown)


docs/package.json
```diff
@@ -4,13 +4,14 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "build-docs": "pnpm run update-contributors && vitepress build && cp _redirects .vitepress/dist",
+    "build-docs": "pnpm run update-contributors && pnpm run -w build && vitepress build && cp _redirects .vitepress/dist",
```
drwpow, line 7: @kerwanp Probably worth explaining this weird little quirk of docs:

- Turborepo will just run the `build` script in every package. We recommend contributors do this when setting up the project locally
- However, because we have our contributors list pulling from GitHub, we need an API token to pull this data
  - To fix this, either we require everyone to have a GitHub API token (annoying), or we just make the docs build a different command (less annoying)
- Because the docs command is intentionally different, this means it runs at different times than the build
- The docs before didn’t need a build of all packages, so we just skipped it. But since that’s now a requirement, we’ll just build as part of the docs.

This may add a tiny bit of delay to the docs, but the monorepo builds really quickly, so it’s better to just build everything than try and surgically determine the minimal build. It’s always a balance of “what’s efficient compute-wise” and “what’s easiest to work with people-wise” and it’s always good to try and satisfy both reasonably.

packages/openapi-react-query/vitest.config.ts
```diff
@@ -1,8 +1,8 @@
 import react from "@vitejs/plugin-react";
-import { defineConfig } from "vitest/config";
+import { defineConfig, type Plugin } from "vitest/config";
 
 export default defineConfig({
-  plugins: [react()],
+  plugins: [react() as unknown as Plugin],
```
drwpow, line 5: Ugh this weird type issue is back :/ I feel like this comes and goes a lot

drwpow:
If it’s alright, I’m just going to force this through without approval because it doesn’t change anything; just unblocks the docs deploy

PR #2059, opened 2024-12-19
fix(openapi-typescript): handle nullable schemas

gduliscouet-ubitransport:
## Changes

Allow all types except UNKNOWN to be nullable.
Fixes https://github.com/openapi-ts/openapi-typescript/issues/1821 and https://github.com/openapi-ts/openapi-typescript/issues/1940 .

## How to Review

I re-applied https://github.com/openapi-ts/openapi-typescript/pull/1959 (it didn't move for a while)

I added tests for the new syntax for OAS 3.1 (not the one with `oneOf`, but with a `type` array, as [documented here](https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0#:~:text=Swap%20nullable%20for%20type%20arrays)).

## Checklist

- [x] Unit tests updated
- [x] [add a minor changeset patch](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)

### DIFF
--- a/packages/openapi-typescript/test/index.test.ts
+++ b/packages/openapi-typescript/test/index.test.ts
@@ -759,6 +759,119 @@ export enum ApiPaths {
         },
       },
     ],
+    [
+      "nullable > 3.0 syntax",
+      {
+        given: {
+          openapi: "3.0.3",
+          info: {
+            title: "Test",
+            version: "0",
+          },
+          paths: {},
+          components: {
+            schemas: {
+              obj1: {
+                title: "Nullable object",
+                type: "object",
+                properties: {
+                  id: {
+                    type: "string",
+                  },
+                },
+                nullable: true,
+              },
+              obj2: {
+                title: "Nullable empty object",
+                type: "object",
+                nullable: true,
+              },
+              str: {
+                title: "Nullable string",
+                type: "string",
+                nullable: true,
+              },
+            },
+          },
+        },
+        want: `export type paths = Record<string, never>;
+export type webhooks = Record<string, never>;
+export interface components {
+    schemas: {
+        /** Nullable object */
+        obj1: {
+            id?: string;
+        } | null;
+        /** Nullable empty object */
+        obj2: Record<string, never> | null;
+        /** Nullable string */
+        str: string | null;
+    };
+    responses: never;
+    parameters: never;
+    requestBodies: never;
+    headers: never;
+    pathItems: never;
+}
+export type $defs = Record<string, never>;
+export type operations = Record<string, never>;`,
+      },
+    ],
+    [
+      "nullable > 3.1 syntax",
+      {
+        given: {
+          openapi: "3.1.0",
+          info: {
+            title: "Test",
+            version: "0",
+          },
+          paths: {},
+          components: {
+            schemas: {
+              obj1: {
+                title: "Nullable object",
+                type: ["object", "null"],
+                properties: {
+                  id: {
+                    type: "string",
+                  },
+                },
+              },
+              obj2: {
+                title: "Nullable empty object",
+                type: ["object", "null"],
+              },
+              str: {
+                title: "Nullable string",
+                type: ["string", "null"],
+              },
+            },
+          },
+        },
+        want: `export type paths = Record<string, never>;
+export type webhooks = Record<string, never>;
+export interface components {
+    schemas: {
+        /** Nullable object */
+        obj1: {
+            id?: string;
+        } | null;
+        /** Nullable empty object */
+        obj2: Record<string, never> | null;
+        /** Nullable string */
+        str: string | null;
+    };
+    responses: never;
+    parameters: never;
+    requestBodies: never;
+    headers: never;
+    pathItems: never;
+}
+export type $defs = Record<string, never>;
+export type operations = Record<string, never>;`,
+      },
+    ],
   ];
 
   for (const [testName, { given, want, options, ci }] of tests) {
--- a/.changeset/strange-cherries-collect.md
+++ b/.changeset/strange-cherries-collect.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": minor
+---
+
+Allow all types except UNKNOWN to be nullable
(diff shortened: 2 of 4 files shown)


gduliscouet-ubitransport:
Hello @drwpow ,
Could I get your input for [this schema](https://github.com/openapi-ts/openapi-typescript/issues/1821#issuecomment-2399221160) ?
I [added it to the tests](https://github.com/openapi-ts/openapi-typescript/pull/2059/files#diff-825cc17817d6be362cee3768df588ac6fa8ee329b4b8990dcc85767d8ad3cf3bR806-R820), but it doesn't pass yet.
From what I see [in the code](https://github.com/openapi-ts/openapi-typescript/blob/3a725fd937e9fc8b9b788c77f728c5a00dac1218/packages/openapi-typescript/src/transform/schema-object.ts#L76), it seems to me that if we have a `$ref` in an object schema, all other keys are ignored. Is it me that misunderstood how $ref work ? Or is it a bug we should fix ?

Once we resolve that I'll finalize the PR

drwpow:
> it seems to me that if we have a `$ref` in an object schema, all other keys are ignored. Is it me that misunderstood how `$ref` work ? Or is it a bug we should fix ?

That’s my understanding. If you look at the [OpenAPI 3.1 specification](https://spec.openapis.org/oas/v3.1.0.html#components-object), you’ll often see typings like this: `SchemaObject | ReferenceObject`. The fact that a reference object IS NOT a SchemaObject, or any other type for that matter, means they aren’t to be mixed-and-matched. If you want to combine multiple objects, you MUST use `allOf`/`oneOf`/`anyOf` to compose an object (but within those arrays, you are allowed to have some object be ReferenceObject and some SchemaObjects, but there’s no such thing as combining the two together.

packages/openapi-typescript/test/index.test.ts
```diff
@@ -694,6 +694,184 @@ export type operations = Record<string, never>;`,
...
+            schemas: {
+              obj1Ref: {
+                properties: {
+                  id: { type: "string" },
+                },
+              },
+              obj1: {
+                type: "object",
+                nullable: true,
+                $ref: "#/components/schemas/obj1Ref",
```
drwpow, line 808: As discussed, this is simply an invalid type. We shouldn’t be testing this. To combine `obj1Ref` and these properties, it must be within `allOf` or `anyOf`. We can also simply remove this test if it’s the only one failing (haven’t looked a the failure logs yet)

gduliscouet-ubitransport: Thank you for the quick reply. 

Well you are right with the fact that in OAS 3.1 spec there are both the SchemaObject (a thing of JSON Schema spec) and the ReferenceObject (a thing of OAS spec). And that the ReferenceObject "cannot be extended with additional properties".

But, and this is what confused me initially, **having a `$ref` in an object doesn't make it a ReferenceObject**. I think [this](https://github.com/OAI/OpenAPI-Specification/issues/1514#issuecomment-1105514112) and [this](https://github.com/OAI/OpenAPI-Specification/issues/1514#issuecomment-719993986) comments resume it best.

So in the end, here:
- obj1 is a SchemaObject
- additionnal properties are allowed
- the merging of properties should be done following JSON schema's spec

What I propose is to keep this PR only for the nullable subject. And I'll open another issue for the ref handling subject. What do you think ?

drwpow: I do see those comments, but that does seem to conflict with what I’m reading in the specs themselves.

OpenAPI 3.1 has [language like the following](https://spec.openapis.org/oas/v3.1.0.html#fixed-fields-6):

> In case a Path Item Object field appears both in the defined object and the referenced object, the behavior is undefined.

Also JSON Schema has [language like so](https://json-schema.org/understanding-json-schema/structuring#dollarref):

> When an object contains a $ref property, the object is considered a reference, not a schema. Therefore, any other properties you put in that object will not be treated as JSON Schema keywords and will be ignored by the validator.

The fact that I can’t find this is explicitly allowed and has defined behavior means I’d rather not support it, mainly because going beyond the spec tends to come back to cause other unexpected behavior in the weirdest ways that are hard to backpedal from.

> What I propose is to keep this PR only for the nullable subject. And I'll open another issue for the ref handling subject. What do you think ?

That sounds great to me! I think an issue explicitly discussing this behavior would be great. Like with anything I could be wrong. I would just like additional input and arguments that make a good case for what the behavior should be, and point to sources (and more importantly, how it differs from anyOf/allOf composition!)

gduliscouet-ubitransport: Ok I'll take care of opening the new issue and finishing this PR when I have a computer at hand.

For your 1st quote, on the Path Item Object: it is not a Schema Object. For some reason they introduced a 3rd use of the $ref here, instead of proposing to use a Reference Object. In places like the response body schema, the only definition of the $ref we should consider is the one of the Schema Object, and it is defined in JSON Schema's spec.

For your second quote: it is taken from an insert about previous draft versions (Draft 4 to Draft 7). Openapi 3.1 is relying on Draft 2020-12, which introduced the new way to handle $ref.
However openapi versions 3.0.x are relying on Draft 7, so the change came with the 3.1

gduliscouet-ubitransport: So as discussed I removed anything related to $ref handling from this PR @drwpow. Now it is only for the `nullable` handling fix.

drwpow: Ah thank you for taking the time to explain it clearly. I had not been distinguishing between Draft 2020-12 and other drafts. To be honest I had been mostly looking at JSON Schema through the lens of what appears in the OpenAPI 3.0 and 3.1 specs, rather than diving deep into those specs themselves. Funny enough, I don’t think anyone had raised this behavior before as something they’d been relying on, which is why it was new to me.

Thanks for making the change here. I’m open to introducing this behavior as supported. If we can support that without changing existing behavior we could do that in a followup PR. If it’s more of a breaking change we could schedule it for the next major release.

drwpow (approved):
Great tests, and great fix! Replied to the comment about `$ref` handling, but let’s merge & release the rest of it as-is

packages/openapi-typescript/test/index.test.ts
gduliscouet-ubitransport: Thank you !
Well, after seeing all the confusion on ref resolution between 3.0, 3.1 (and 3.2 will probably change this again 😅), we prefered not to relie on additionnal properties for now

PR #2061, opened 2024-12-20
Drop init argument when not needed in query key

kevmo314:
## Changes

For urls that do not include an init param, generate a query key of length 2. This allows the value to be passed directly to `invalidateQueries()`.

Without this change, the package generates `["get", "/foo", undefined]` which does not correctly match the corresponding get query.

Related to https://github.com/openapi-ts/openapi-typescript/issues/1806

## How to Review

This doesn't quite solve the problem when there are init params but at least solves for the case where there are no params...

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/sour-steaks-double.md
+++ b/.changeset/sour-steaks-double.md
@@ -1,0 +1,5 @@
+---
+"openapi-react-query": patch
+---
+
+Drop init argument when not needed in query key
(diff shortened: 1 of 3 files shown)


kerwanp:
Looks good to me ! Could you add a minor changeset?

drwpow (approved):
Great test, thank you! If you could add a `patch` **changeset** (see comment) I’d be happy to approve & release

PR #2063, opened 2024-12-21
feat(openapi-fetch): add support for arbitrary method

### LINKED ISSUE #1808: [openapi-fetch] Provide `client.request()` to support dynamic method.
**Description**

Currently, `openapi-fetch` uses the method as the function name (ex: `client.GET()`) and there is no way to use the method as an argument. When building around `openapi-fetch` it requires a bit of gymnastic and manual typing to have a dynamic method.

```ts
          const mth = method.toUpperCase() as keyof typeof client;
          const fn = client[mth] as ClientMethod<Paths, typeof method, Media>;
```

We can also find this pattern in libraries like axios, ky, etc.

**Proposal**

Internally, `openapi-fetch` already use `coreFetch` method to pass the method as an argument.

We could have a `request` (or `REQUEST` to respect casing) method on the client. (or `FETCH`).

```ts
  return {
    /** Call an endpoint */
    async REQUEST(method, url, init) {
          return coreFetch(url, { ...init, method });
    }
    /** Call a GET endpoint */
    async GET(url, init) {
      return coreFetch(url, { ...init, method: "GET" });
    },
    /** Call a PUT endpoint */
    async PUT(url, init) {
      return coreFetch(url, { ...init, method: "PUT" });
    },
    /** Call a POST endpoint */
    async POST(url, init) {
      return coreFetch(url, { ...init, method: "POST" });
    },
    [...]
  }
```

**Checklist**

- [x] I’m willing to open a PR for this (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

kerwanp:
## Changes

This Pull request closes #1808 by providing a `request` method that accept a method as its first arguments allowing to pass any `HttpMethod`. Making library authors' life easier.

This does not introduce any breaking change.

## How to Review

A new `request` should be available on the client that **must behave exactly** like the other `GET`, `OPTIONS`, etc methods but with the ability to pass the HttpMethod as first parameter.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/hot-pants-sit.md
+++ b/.changeset/hot-pants-sit.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+add support for arbitrary method
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -199,6 +199,16 @@ export type ClientMethod<
   ...init: InitParam<Init>
 ) => Promise<FetchResponse<Paths[Path][Method], Init, Media>>;
 
+export type ClientRequestMethod<Paths extends Record<string, Record<HttpMethod, {}>>, Media extends MediaType> = <
+  Method extends HttpMethod,
+  Path extends PathsWithMethod<Paths, Method>,
+  Init extends MaybeOptionalInit<Paths[Path], Method>,
+>(
+  method: Method,
+  url: Path,
+  ...init: InitParam<Init>
+) => Promise<FetchResponse<Paths[Path][Method], Init, Media>>;
+
 export type ClientForPath<PathInfo extends Record<string | number, any>, Media extends MediaType> = {
   [Method in keyof PathInfo as Uppercase<string & Method>]: <Init extends MaybeOptionalInit<PathInfo, Method>>(
     ...init: InitParam<Init>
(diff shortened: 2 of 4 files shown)


.changeset/hot-pants-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+"openapi-fetch": minor
```
drwpow, line 2: This is a patch because it doesn’t introduce a breaking change (that I’m aware of)

```suggestion
"openapi-fetch": patch
```

kerwanp: According to semver:
```
MAJOR version when you make incompatible API changes
MINOR version when you add functionality in a backward compatible manner
PATCH version when you make backward compatible bug fixes
```

drwpow: That only applies for > 1.0! For 0.x, it’s more standard to smush bugfixes/features into patches, and use minor for breaking changes. It helps people plan more accordingly. Semver dictates that 0.x releases are allowed to make breaking changes at any time, and even though this method isn’t “official,” it’s a very common pattern, and one that openapi-fetch has followed since the beginning.

But you are correct—that becomes the pattern as soon as 1.0 is released and thereafter

kerwanp: I was not aware of this! Thanks for letting me know, I'll update that

drwpow (approved):

gzm0:
FWIW: I've briefly checked whether it is worth backporting this to the path-based client and I believe for now it isn't: The API is lower-level than what the path-based client offers. So IMO it is unlikely someone would want the combination of path-based but arbitrary method.

PR #2065, opened 2024-12-21
Add Netlify badge

drwpow:
## Changes

Adds Netlify Badge because we’re moving hosts for the documentation. And as part of their OSS program they generously donate free hosting in exchange for a badge. Though the docs site has been on Cloudflare until now, Netlify is much friendlier for OSS collaboration and anyone being able to deploy docs previews.

_Note: you’ll still see Cloudflare on PRs temporarily, while we give DNS a generous amount of time to switch over. Wouldn’t want to accidentally have the docs site be disabled for some folks during the transition!_

CleanShot 2024-12-21 at 15 08 03@2x

## How to Review

- Any feedback on size/placement?

## Checklist

N/A

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -4,7 +4,7 @@ import zh from "./zh";
 import ja from "./ja";
 import shared from "./shared";
 import { transformerTwoslash } from "@shikijs/vitepress-twoslash";
-import { ModuleKind, ModuleResolutionKind } from "typescript";
+import { ModuleResolutionKind } from "typescript";
 
 // https://vitepress.dev/reference/site-config
 export default defineConfig({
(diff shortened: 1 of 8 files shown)


kerwanp:
Looks good to me! Is there any other available logo format? I'm really not fan of this box.
And the "Built with" is really confusing

kerwanp (approved):

drwpow:
> Looks good to me! Is there any other available logo format? I'm really not fan of this box. 

Yeah… they rebranded to a very “tall” logo that doesn’t fit inside a box easily. They have some color options but this is basically the only “badge” version they provide. I kinda agree—maybe I’ll look into another logo that still meets their requirements. They also allow a link without a logo; I can try that as well.

drwpow:
Simplified it; went with a less-noticeable link that still meets their terms of use requirements

PR #2066, opened 2024-12-21
Auto-assign issues

drwpow:
## Changes

GitHub, by default, doesn’t allow auto issue assignment. Try out a GitHub action to automatically round-robin select one maintainer to triage the issue, and follow up.

## How to Review

- Any additional settings desirable?
- Any other GitHub Actions we should evaluate (I’ll admit I just picked the first one I found)?

## Checklist

N/A

### DIFF
--- a/.github/workflows/issue.yml
+++ b/.github/workflows/issue.yml
@@ -1,0 +1,21 @@
+name: Issue creation
+
+on:
+  issues:
+    types:
+      - opened
+
+jobs:
+  auto-assign:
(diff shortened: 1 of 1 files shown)


gzm0 (commented):
> Any additional settings desirable?

Settings LGTM.

> Any other GitHub Actions we should evaluate (I’ll admit I just picked the first one I found)?

IMO just picking something is fine here. It's not like we're building on this. So if it stops working for us, I expect it to be easy to change.

PR #2091, opened 2025-01-04
feat(openapi-fetch): Allow returning Response from onRequest callback

paveldubovitsky:
## Changes

Added ability to return `Response` from `onRequest` middleware hook. This allows middleware to short-circuit the request chain by returning a response directly, which is useful for cases such as deduplicating or caching responses to avoid unnecessary network requests.

## How to Review

Please check that:
- Early `Response` from `onRequest` properly skips remaining middleware chain
- Test coverage is sufficient
- Documentation clearly explains new functionality

Key implementation details:
- No changes to existing middleware behavior
- When `Response` is returned:
  - The request is not sent to the server
  - Subsequent `onRequest` handlers are skipped
  - `onResponse` handlers are skipped

## Checklist

- [x] Unit tests updated


packages/openapi-fetch/test/middleware/middleware.test.ts
```diff
@@ -443,3 +443,64 @@ test("type error occurs only when neither onRequest nor onResponse is specified"
   assertType<Middleware>({ onResponse });
   assertType<Middleware>({ onRequest, onResponse });
 });
+
+test("can return response directly from onRequest", async () => {
+  const customResponse = Response.json({});
+  const client = createObservedClient<paths>();
```
gzm0, line 449:
```suggestion
  const client = createObservedClient<paths>({}, () => throw new Error("unexpected call to fetch"));
```

This way, we make sure we really do not call fetch.

packages/openapi-fetch/test/middleware/middleware.test.ts
```diff
@@ -443,3 +443,64 @@ test("type error occurs only when neither onRequest nor onResponse is specified"
...
+  const { response } = await client.GET("/posts/{id}", {
+    params: { path: { id: 123 } },
+  });
+
+  expect(response).toBe(customResponse);
+});
+
+test("skips subsequent onRequest handlers when response is returned", async () => {
+  let onRequestCalled = false;
+  const customResponse = Response.json({});
```
gzm0, line 466: Nit: Inline below? You don't actually need the reference for this test.

packages/openapi-fetch/test/middleware/middleware.test.ts
```diff
@@ -443,3 +443,64 @@ test("type error occurs only when neither onRequest nor onResponse is specified"
...
+  );
+
+  await client.GET("/posts/{id}", { params: { path: { id: 123 } } });
+
+  expect(onRequestCalled).toBe(false);
+});
+
+test("skips onResponse handlers when response is returned from onRequest", async () => {
+  let onResponseCalled = false;
+  const customResponse = Response.json({});
```
gzm0, line 490: Same: Inline below?

packages/openapi-fetch/test/middleware/middleware.test.ts
gzm0, line 1: Nice tests!

gzm0 (changes requested):
Nice! (and sorry for the somewhat tardy review).

Mostly nits on tests, and one doubt in the example that we probably should address.

packages/openapi-fetch/src/index.js
```diff
@@ -124,79 +125,84 @@ export default function createClient(clientOptions) {
...
           if (result) {
-            if (!(result instanceof CustomRequest)) {
-              throw new Error("onRequest: must return new Request() when modifying the request");
+            if (result instanceof CustomRequest) {
+              request = result;
+            } else if (result instanceof Response) {
+              response = result;
+              break;
+            } else {
+              throw new Error("onRequest: must return new Request() or Response() when modifying the request");
```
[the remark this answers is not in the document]
drwpow: Yeah I’m keen to just remove this; this is not true in some cases. It was meant to provide a friendlier error for folks, but I actually think just letting the platform surface the error is more helpful, rather than us getting in the way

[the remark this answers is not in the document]
p-dubovitsky: I'm not sure I understood you correctly. Do you want to throw an error without a message or something else?

packages/openapi-fetch/test/middleware/middleware.test.ts
p-dubovitsky: Thanks!

gzm0 (approved):

StrongerMyself:
Please update the openapi-fetch package with this functionality. The latest available version in npm is 0.13.4, does not have ‘Early Response’

docs/openapi-fetch/middleware-auth.md
```diff
@@ -64,6 +64,38 @@ onRequest({ schemaPath }) {
...
+    if (cached) {
+      // Return cached response, skipping actual request and remaining middleware chain
+      return cached.clone();
+    }
+  },
+  onResponse({ request, response }) {
+    if (response.ok) {
+      const key = getCacheKey(request);
+      cache.set(key, response.clone());
+    }
```
StrongerMyself, line 88: Here you must return `return response.clone()`, otherwise the request body will be read on line 222 in index.js and repeated requests will be rejected with an error

gzm0: Are you sure? The response in the cache is cloned.

I say this only from looking at the code, if you have tried this and it fails, well, then it fails  :)

PR to fix the example in this case is very much appreciated.

StrongerMyself: I am using nextjs 15 (server-side requests). At some point `response` becomes bodyUsage=true, maybe it is related to the environment - `nodesj`. Most likely your example works for client-side code, but not for server-side code. I'm not sure how best to flesh out the documentation. In my case, I save data(as text), headers and statuses in Map. 

```ts
const cache = new Map<string, string>();
const getCacheKey = (request: Request) => `${request.method}:${request.url}`;

const cacheMiddleware: Middleware = {
  onRequest({ request }) {
    const key = getCacheKey(request);
    const cached = cache.get(key);
    if (cached) {
      // Return cached response, skipping actual request and remaining middleware chain
      const response = new Response(cached);
      return response;
    }
  },
  async onResponse({ request, response }) {
    if (response.ok) {
      const key = getCacheKey(request);
      const responseBody = await response.clone().text();
      cache.set(key, responseBody);
    }
  }
};
```

gzm0: TY for the example. Unfortunately, without more understanding of what is going on here, from the maintainers side, I don't think we have a strong enough signal to take this further.

If you would like, please file a bug with this, and investigation can proceed there. However, please understand that you will be expected to produce a reproducible example that relies on a (standard) fetch implementation only (of course that needs to run in browsers and Node.js, but no SSR magic, just free-standing code).

PR #2096, opened 2025-01-11
Encode the request body if `Content-Type` header is `x-www-form-urlencoded`

### LINKED ISSUE #2069: Automatically encode the body when `"Accept": "application/x-www-form-encoded"` header is passed
## Problem
For the endpoints requiring `x-www-form-encoded` body parameters, the request body is serialized as a string, and then sent as an empty object string `{}`. The solution is to provide a custom body parser that converts the plain object of the body into URL-encoded string. This is not documented, and is difficult to debug.

## Possible solutions
1. The simplest solution would be to document the need to pass a custom `bodySerializer`.

2. A better solution would be to automatically encode the body if the header is passed.
The default `defaultBodySerializer` could accept the `fetchOptions.headers` parameter, and return `new URLSearchParams(body).toString` if the `"Accept": "x-www-form-encoded"` header is passed by the user.

https://github.com/openapi-ts/openapi-typescript/blob/d4689b10af870b6978e5c6107a55fff044936b39/packages/openapi-fetch/src/index.js#L564-L573

3. Even better would be to automatically infer when the endpoint requires encoded body from the OpenAPI types, and automatically encode it, even without the need to set the header.

## Additional context
I ran into this problem when working on [`@openverse/api-client`](https://github.com/WordPress/openverse/tree/main/packages/js/api-client). Openverse API uses [Django OAuth Toolkit](https://django-oauth-toolkit.readthedocs.io/en/latest/) for authentication, and the token requests require the body to be url-encoded. The API requests from the `api-client` were failing due to `invalid_grant_type`. Turned out that the request was sending an empty body `{}`.

obulat:
## Changes

Fixes #2069 
This PR updates the `defaultBodySerializer` to correctly serialize the body if `Content-Type` header is set to `x-www-form-urlencoded`. This is useful for OAuth2 token requests.

## How to Review

I'm not sure about the headers: what type are they, and are they normalized?
I added handling for both a plain JS object (`headers['Accept-Content']`) and a [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers/get) object (`headers.get['Accept-Content']`), as well as both capitalized and lower-case versions.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -72,7 +72,7 @@ export default function createClient(clientOptions) {
             });
     }
 
-    const serializedBody = body === undefined ? undefined : bodySerializer(body);
+    const serializedBody = body === undefined ? undefined : bodySerializer(body, headers);
 
     const defaultHeaders =
       // with no body, we should not to set Content-Type
@@ -568,9 +568,18 @@ export function defaultPathSerializer(pathname, pathParams) {
  * Serialize body object to string
  * @type {import("./index.js").defaultBodySerializer}
  */
-export function defaultBodySerializer(body) {
+export function defaultBodySerializer(body, headers) {
   if (body instanceof FormData) {
     return body;
+  }
+  if (headers) {
+    const contentType =
+      headers.get instanceof Function
+        ? (headers.get("Content-Type") ?? headers.get("content-type"))
+        : (headers["Content-Type"] ?? headers["content-type"]);
+    if (contentType === "application/x-www-form-urlencoded") {
+      return new URLSearchParams(body).toString();
+    }
   }
   return JSON.stringify(body);
 }
--- a/docs/openapi-fetch/api.md
+++ b/docs/openapi-fetch/api.md
@@ -192,6 +192,22 @@ or when instantiating the client.
 
 :::
 
+### URL-encoded body
+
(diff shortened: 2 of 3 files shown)


packages/openapi-fetch/src/index.js
```diff
@@ -72,7 +72,7 @@ export default function createClient(clientOptions) {
             });
     }
 
-    const serializedBody = body === undefined ? undefined : bodySerializer(body);
+    const serializedBody = body === undefined ? undefined : bodySerializer(body, headers);
```
drwpow, line 75: So these are the headers unique to this request (probably could be named better). Below on line `92`, we merge all the headers together. We should probably move that up, adn pass that in here to `bodySerializer`. That way it takes into account headers that users have set in their `createClient()` call, and it handles the normalization for you so it’s always a `Header` type (and you aren’t dealing with what the user passed in)

drwpow:
This approach looks good to me, and good test! But before we merge, we’ll want to:

1. Use the merged/normalized headers like the comment mentioned
2. Add a `patch` **changeset** (see comment)

drwpow:
@obulat Still willing to accept this PR with the changes mentioned!

drwpow (approved):
This PR is close enough! I’ll make the changes post-merge to make sure we’re not grabbing incorrect/stale headers, and add a changeset retroactively.

obulat:
@drwpow, sorry I couldn't finish up the work here, and thank you so much for merging!

drwpow:
Oh no worries at all; the maintainers and I all had individual things come up that made reviews earlier in the year slower than we’d like too. Thanks again for the work! 🙏

Issue #1949, opened 2024-10-16
openapi-react-query select option does not infer types correctly


lukasbash:
**Description**

The `select` property of the query options does not infer the type correctly.

**Reproduction**

Simply return a different type from the select call.
Example:
```ts
const {data:firstUser} = QAPI.useQuery(
    "get",
    "/api/users", // endpoint usually returns an array of users, e.g. User[]
    {},
    {
      select: (data) => data[0], // want to return something different here, i.e. User
    },
  );
```

In this case you get an error from the compiler stating `Type User is not assignable to type User[]` ...

**Expected result**

Type inference just works.

**Checklist**

- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-fetch/CONTRIBUTING.md))

kerwanp:
Closing as duplicate of #1845

PR #2105, opened 2025-01-20
fix(openapi-react-query): correctly infer select return value

### LINKED ISSUE #1845: Wrong type in openapi-react-query when using select to transform data
*Mind: This is a bug in openapi-react-query. There is no template for opening Bugs for the openapi-react-query package yet.*

**Description**

The `select` transformer in the `queryOptions` of the useQuery method expects a value of type `data` as return value. The value should be transformable to an arbitrary value (and be inherited correctly). Example:

Cool:
```ts
$api.useQuery('get', '/pets/person', {},
  {
    select: (data) => data,
  }
);
```
Not cool:
```ts
$api.useQuery('get', '/pets/person', {},
  {
    select: (data) => data.age, // or any other value
  }
);
```

**Reproduction**

See the implementation at the bottom of `fetchClient.ts` in this  [Stackblitz](https://stackblitz.com/edit/vitejs-vite-gwavds?file=src%2FfetchClient.ts).

**Expected result**

Arbitrary return values in the `select` method do not throw type errors. Furthermore, `data` should infer the type of the returned value of the `select` method (if set):

```ts
const { data } = $api.useQuery('get', '/pets/person', {},
  {
    select: (data) => data.age, // this shouldn't throw any type errors
  }
);

const str: number | undefined = data; // this shouldn't throw any type errors either as date.age is of type number and optional.
```

HagenMorano:
## Changes

Resolves #1845 #1949

## How to Review

See added test. When adding the select option, `useQuery` and `useSuspenseQuery` now infers the correct type when using the `select` transformer.

Example: 

```ts
client.useQuery(
  "get",
  "/string-array",
  {},
  {
    select: () => 1, // resolves to type number
  },
),
```

```ts
client.useQuery(
  "get",
  "/string-array",
  {},
  {
    select: (data) => [data], // resolves to type `data[]`
  },
),
```

```ts
client.useQuery("get","/string-array") // still infers correct data (no change)
```

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/chilly-meals-act.md
+++ b/.changeset/chilly-meals-act.md
@@ -1,0 +1,5 @@
+---
+"openapi-react-query": patch
+---
+
+[#1845](https://github.com/openapi-ts/openapi-typescript/pull/2105): The return value of the `select` property is now considered when inferring the `data` type.
--- a/packages/openapi-react-query/src/index.ts
+++ b/packages/openapi-react-query/src/index.ts
@@ -15,6 +15,9 @@ import {
 import type { ClientMethod, FetchResponse, MaybeOptionalInit, Client as FetchClient } from "openapi-fetch";
 import type { HttpMethod, MediaType, PathsWithMethod, RequiredKeysOf } from "openapi-typescript-helpers";
 
+// Helper type to dynamically infer the type from the `select` property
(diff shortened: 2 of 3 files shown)


kerwanp:
Looks really good to me, thanks for the contrib!

Could you add a patch changeset so I can merge this?

Issue #1752, opened 2024-07-08
Generates invalid types when API paths are nested


mat-certn:
**Description**

We have multiple paths in our schema that are "nested" below each other, like so:

```
/some/path/{id}
/some/path/{id}/summary
```

openapi-typescript using the argument --path-params-as-types generates the following:

```
  [path: `/some/path/${string}`]: SchemaOne;
  [path: `/some/path/${string}/summary`]: SchemaTwo;
```

which conflicts with the error

```
'x' index type 'SchemaOne' is not assignable to 'x' index type 'SchemaTwo'.
```

this is because technically `/some/path/${string}/summary` is generated as incorrectly assignable to `/some/path/${string}` - instead of ${string} it should be some sort of type that cannot include "/"

| Name                 | Version                      |
| :------------------- | :--------------------------- |
| `openapi-typescript` | `7.0.2`                      |
| TypeScript              | `5.3.3` or `5.5.3` (have tried a few versions)                     |
| Node.js              | `18.18.2`                      |
| OS + version         | `macOS 14.5` |

**Reproduction**

Generate types using the following schema: 

```
openapi: 3.0.3
servers:
  - url: http://some-server.com
    description: Local development server
info:
  title: Example API with conflicting paths
  version: 0.0.1
  license:
    name: MIT
    url: https://opensource.org/licenses/MIT
paths:
  /api/internal/cases/{short_id}:
    get:
      summary: "Get case details"
      operationId: cases_retrieve
      parameters:
        - in: path
          name: short_id
          schema:
            type: string
          required: true
      tags:
        - cases
      security:
        - AuthZeroUserAuth: []
      responses:
        "200":
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/CaseDetail"
          description: ""
        "404":
          description: No response body
  /api/internal/cases/{short_id}/summary:
    get:
      summary: "Get case summary"
      operationId: cases_summary_retrieve
      parameters:
        - in: path
          name: short_id
          schema:
            type: string
          required: true
      tags:
        - cases
      security:
        - AuthZeroUserAuth: []
      responses:
        "200":
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/CaseSummary"
          description: ""
        "404":
          description: No response body
components:
  schemas:
    CaseDetail:
      type: object
      properties:
        case_name:
          type: string
          format: uuid
          readOnly: true
      required:
        - case_name
    CaseSummary:
      type: object
      properties:
[...]

PR #2106, opened 2025-01-20
feat: fixed typescript error dynamic paths

walid-mos:
## Changes

Openapi-typescript creates an faulty file when using --path-params-as-types to generate types when nested path parameters exists.

This is particularly annoying when used with a Restful CRUD like : 
- /users
- /users/{userId}
... 

[#1752](https://github.com/openapi-ts/openapi-typescript/issues/1752)

## How to Review

## Checklist

- [ x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/src/index.ts
+++ b/packages/openapi-typescript/src/index.ts
@@ -18,7 +18,7 @@ export { default as transformOperationObject } from "./transform/operation-objec
 export { default as transformParameterObject } from "./transform/parameter-object.js";
 export * from "./transform/path-item-object.js";
 export { default as transformPathItemObject } from "./transform/path-item-object.js";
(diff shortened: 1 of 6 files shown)


walid-mos:
Hello, I drafted the PR because I thought there was a bug in my solution.

Finally I came across this observation, when you use a path that has a static part after the dynamic part, you have to add a / to make the typing work.

I don't know if that's likely to cause a problem, in my case it's fine, but I don't want to speak for other people.

Here's some example of what i am talking about : 
Capture d’écran 2025-01-20 à 23 07 06
Capture d’écran 2025-01-20 à 23 10 20

without trailing slash: 
Capture d’écran 2025-01-20 à 23 10 43


any other path works fine

PR #2114, opened 2025-01-23
fix: use index.d.ts as types for cjs

### LINKED ISSUE #2098: Build output is missing the index.d.cts file causing typescript to fail
### openapi-react-query version

0.2.9

### Description

The build output is missing the .d.cts file for Require.types settings in package.json.

### Reproduction

Trying to 

```
import { createClient } from 'openapi-react-query'
// Could not find a declaration file for module 'openapi-react-query'. '.../node_modules/openapi-react-query/dist/index.cjs' implicitly has an 'any' type.
```

From package.json I see:

```
  "exports": {
    ".": {
      "import": {
        "types": "./dist/index.d.ts",
        "default": "./dist/index.js"
      },
      "require": {
        "types": "./dist/index.d.cts", // THIS SETTING
        "default": "./dist/index.cjs"
      }
    },
    "./*": "./*"
  },
```

But it does not exist in node modules (openapi-fetch for comparison):

image

### Expected result

I am unsure how exactly to fix this, but clearly for me the file is missing from the build output. Can you add it?

### Extra

- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-react-query/CONTRIBUTING.md))

kerwanp:
## Changes

This pull request closes #2098. `exports.require.types` was pointing to a file that does not exist. It now points to the same `exports.imports.types` file (`index.d.ts`) fixing type issues when using this library with `require`. 

## How to Review

Types should properly resolved when using a `require` import.

### DIFF
--- a/.changeset/witty-bottles-carry.md
+++ b/.changeset/witty-bottles-carry.md
@@ -1,0 +1,5 @@
+---
+"openapi-react-query": patch
+---
+
+[#2098](https://github.com/openapi-ts/openapi-typescript/pull/2098): Fix CJS type issues by pointing to proper d.ts file
--- a/packages/openapi-react-query/package.json
+++ b/packages/openapi-react-query/package.json
@@ -18,7 +18,7 @@
         "default": "./dist/index.js"
       },
       "require": {
-        "types": "./dist/index.d.cts",
+        "types": "./dist/index.d.ts",
         "default": "./dist/index.cjs"
       }
     },


drwpow (approved):
I think this will fix it as this only builds one file. But when packages have multiple files and they import from one another, you have to generate a separate set of types generally (or manually redeclare types for every file. Unbuild does a really good job with this though, and agree we should just switch all packages to using that.

PR #2126, opened 2025-01-28
Better typesafety for `useInfiniteQuery`

patrickariel:
## Changes

Makes `pageParamName` typesafe and required. This PR also removes the fallbacks for `pageParamName` and `pageParam`, because they shouldn't have default values in the first place (and they're also no longer needed).

The `should use custom cursor params` test is also removed, because I believe that it shouldn't be allowed now that the param is typesafe. Users can always cast the type if they think the schema isn't right.

## How to Review

Most of the changes here are type-level. I'm still not very familiar with the type utilities, so if there are any mistakes let me know.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-react-query/src/index.ts
+++ b/packages/openapi-react-query/src/index.ts
@@ -116,9 +116,9 @@ export type UseInfiniteQueryMethod<Paths extends Record<string, Record<HttpMetho
       InfiniteData<Response["data"]>,
       Response["data"],
       QueryKey<Paths, Method, Path>,
-      Query[PageParamName]
+      NonNullable<Query[PageParamName]>
     >,
-    "queryKey" | "queryFn" | "initialPageParam"
+    "queryKey" | "queryFn"
   >,
 >(
   method: Method,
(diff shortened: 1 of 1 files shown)


minchodang:
Yes, I agree with this PR. 
Initially, when I looked at the design of useInfiniteQuery, the pageParamName contains cursor by default. 
So, when I tested it, queryParams automatically contains cursor on default request.

packages/openapi-react-query/test/index.test.tsx
```diff
@@ -947,6 +948,7 @@ describe("client", () => {
             {
               getNextPageParam: (lastPage) => lastPage.nextPage,
               initialPageParam: 0,
+              pageParamName: "cursor",
```
lukasedw, line 951: Maybe here we could use a different param name other than the default one?

patrickariel: Hmm, not sure what you mean by "the default one". The schema only provides `cursor` and `limit` as parameters, so any other param name here will trigger a type error (due to the changes in this PR). Or are you suggesting a separate schema for the second test?

patrickariel:
Anything I can do here to move this forward?

PR #2152, opened 2025-02-13
fix(openapi-typescript): --make-paths-enum renames the paths URL

insertmike:
## Changes

Fixes: https://github.com/openapi-ts/openapi-typescript/issues/2101, when the --make-paths-enum was reintroduced back it was reintroduced with previously fixed bug: https://github.com/openapi-ts/openapi-typescript/issues/950

## How to Review

It's straight-forward change :) 

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/little-meals-hide.md
+++ b/.changeset/little-meals-hide.md
@@ -2,4 +2,4 @@
 "openapi-typescript": patch
 ---
 
-fix: --make-paths-enum renames the paths URL (:id instead of {id})
+- Fixed --make-paths-enum option transforming the paths URL (`:id` instead of `{id}`)
--- a/.changeset/little-meals-hide.md
+++ b/.changeset/little-meals-hide.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+fix: --make-paths-enum renames the paths URL (:id instead of {id})
--- a/packages/openapi-typescript/src/transform/paths-enum.ts
+++ b/packages/openapi-typescript/src/transform/paths-enum.ts
@@ -30,11 +30,7 @@ export default function makeApiPathsEnum(pathsObject: PathsObject): ts.EnumDecla
           })
           .join("");
       }
-
-      // Replace {parameters} with :parameters
-      const adaptedUrl = url.replace(/{(\w+)}/g, "{$1}");
-
-      enumKeys.push(adaptedUrl);
+      enumKeys.push(url);
       enumMetaData.push({
         name: pathName,
       });
--- a/packages/openapi-typescript/src/transform/paths-enum.ts
+++ b/packages/openapi-typescript/src/transform/paths-enum.ts
@@ -32,7 +32,7 @@ export default function makeApiPathsEnum(pathsObject: PathsObject): ts.EnumDecla
       }
 
       // Replace {parameters} with :parameters
-      const adaptedUrl = url.replace(/{(\w+)}/g, ":$1");
+      const adaptedUrl = url.replace(/{(\w+)}/g, "{$1}");
 
       enumKeys.push(adaptedUrl);
       enumMetaData.push({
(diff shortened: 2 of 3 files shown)


insertmike:
Can we make a patch release based on this one @htunnicliff ?

semanser:
@insertmike I also opened a PR #2153  a few minutes ago! Thanks for the reminder :) As for the fix, it's up to the maintainers but the `adaptedUrl` line can be completely removed.

insertmike:
@Semigradsky You are right, I did not look into the code too much, then the maintainers can feel free to close my PR and merge yours #2153 but please take some action, it's straight-forward change and the community needs it @htunnicliff 🙏

theo-staizen:
@semanser @insertmike great minds think alike 😅

htunnicliff:
Hi all. We are happy to take a look at this, but please refrain from directly mentioning maintainers unless it's absolutely critical. Thank you!

insertmike:
Noted - won't happen again. Thanks for your time :)

htunnicliff:
@drwpow it looks like this does fix a reintroduced bug. However, applying it will affect the behavior for anyone who has started using the flag. I know you mentioned the following when this was fixed, so I want to double-check before we change its behavior:

> since it’s an opt-in feature flag I support adding it! (I’ll only try to ensure that default behavior doesn’t change).


Do you think a patch release is appropriate?

insertmike:
@htunnicliff Looking back in time, when this feature was first introduced, it was implemented incorrectly ([issue](https://github.com/openapi-ts/openapi-typescript/issues/950)) and was patched with this [Pull Request](https://github.com/openapi-ts/openapi-typescript/pull/951). I suspect that when reintroducing the feature, the author of the pull request simply copied the call from the original implementation without realizing that it was initially incorrect and patched in time.

We could add it as a flag, but to me, it just seems fundamentally wrong—ApiPaths is not useful if it doesn’t match the original paths of the OpenAPI schema. WDYT?

insertmike:
@theo-staizen @semanser 

## Temporary Workaround for the --make-paths-enum Bug

Assuming you saved the snippet as postprocess.js:

``` (assuming is named postprocess.js)
/**
 * Workaround for --make-paths-enum bug:
 * https://github.com/openapi-ts/openapi-typescript/pull/2152
 *
 * Generated ApiPaths enum values use ":id" instead of "{id}".
 * This script converts ":param" to "{param}".
 *
 * Usage: node posprocessOpenApiGeneration.js <generated-file-path>
 */
import fs from "fs";

// Get file path from command-line arguments
const filePath = process.argv[2];

if (!filePath) {
  console.error("Usage: node postprocess.js <file-path>");
  process.exit(1);
}

let content = fs.readFileSync(filePath, "utf8");

// Regex to capture the ApiPaths enum block
// Group 1: "export enum ApiPaths {"
// Group 2: the enum body (non-greedy)
// Group 3: the closing brace
const enumRegex = /(export\s+enum\s+ApiPaths\s*{)([\s\S]*?)(\n?\s*})/;

// Replace only inside the ApiPaths enum block
content = content.replace(enumRegex, (match, start, enumBody, end) => {
  // Replace occurrences of ":param" with "{param}" in the enum body
  const modifiedBody = enumBody.replace(/:([a-zA-Z0-9_]+)/g, "{$1}");
  return `${start}${modifiedBody}${end}`;
});

fs.writeFileSync(filePath, content, "utf8");
console.log(`File "${filePath}" updated successfully!`);

```

Add a post-generation script to your package.json so that the patch is applied automatically after generating your API client:

```
{
  "scripts": {
    "script-generating-openapi": "your-openapi-generation-command OUTPUT_PATH",
    "postscript-generating-openapi": "node postprocess.js OUTPUT_PATH"
  }
}
```

When you run the script-generating-openapi command, the post script will automatically be invoked, updating your ApiPaths enum to use the correct placeholder syntax.

gzm0:
@insertmike coming here from #2164. I realize that we should release this soon, but I do not want to overrule ongoing discussions between maintainers. (IMO this clearly is a bugfix, but that's only one opinion).

igieon:
@htunnicliff  can you aprove and release this really bother me.

htunnicliff (approved):

insertmike:
I hope that we soon come to get on the master train :)

PR #2185, opened 2025-02-28
Fix invalid $ref index into path

duncanbeevers:
## ⚠️ The Error

Refs into the parameters array are turned into invalid indexes into the `paths` structure.

## 🗣️ Discussion

The problematic behavior is triggered using a $ref that points into the `parameters` array.
The problematic generated type then tries to index into `["parameters"]["0"]` as though it were an array, when in fact it should be indexing into `["parameters"]["query"]`.

This particular weird ref was generated when I bundled a big schema with [@apidevtools/json-schema-ref-parser](https://github.com/APIDevTools/json-schema-ref-parser/blob/main/docs/ref-parser.md#the-refparser-class).

## :wrench: Fix

* While parsing `oapiRefs` from parameters arrays, pass the resolved ParameterObject and use its `in` method to index into `parameters` type.

This probably doesn't work for more complex $refs, but I think it does slightly improve the semantic awareness of `oapiRef`.

### DIFF
--- a/packages/openapi-typescript/test/transform/webhooks-object.test.ts
+++ b/packages/openapi-typescript/test/transform/webhooks-object.test.ts
@@ -105,10 +105,10 @@ describe("transformWebhooksObject", () => {
                 schema: { type: "string" },
                 required: true,
               },
-              { $ref: "#/components/parameters/query/utm_source" },
-              { $ref: "#/components/parameters/query/utm_email" },
-              { $ref: "#/components/parameters/query/utm_campaign" },
-              { $ref: "#/components/parameters/path/version" },
+              { $ref: "#/components/parameters/utm_source" },
+              { $ref: "#/components/parameters/utm_email" },
+              { $ref: "#/components/parameters/utm_campaign" },
+              { $ref: "#/components/parameters/version" },
             ],
           },
         },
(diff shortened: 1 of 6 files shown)


gzm0:
> The problematic generated type then tries to index into ["parameters"]["0"] as though it were an array, when in fact it should be indexing into ["parameters"]["query"].

` ["parameters"]["path"][0]` in the example? (otherwise I need to read up on how the ref `"#/paths/~1endpoint/get/parameters/0"` is supposed to work :laughing: )

duncanbeevers:
I was also not really familiar with that `$ref` syntax.

It took a bit of digging, but this portion of the $ref is defined by the [JSON Pointer](https://www.rfc-editor.org/rfc/rfc6901) standard.

> Evaluation of each reference token begins by decoding any escaped
   character sequence.  This is performed by first transforming any
   occurrence of the sequence '~1' to '/', and then transforming any
   occurrence of the sequence '~0' to '~'.

This ref is created during bundling, allowing the same schema to be used throughout the document _without_ first extracting it to `components/schemas`. Instead, the first instance is inlined, and subsequent uses reference the initial inlined version via a JSON pointer.

However, this JSON Pointer is indexing into the OpenAPI structure, *not* the openapi-ts `paths` structure. It seems like there's some oversight in how that JSON Pointer is being transformed between OpenAPI land and openapi-ts land.

packages/openapi-typescript/test/transform/webhooks-object.test.ts
```diff
@@ -105,10 +105,10 @@ describe("transformWebhooksObject", () => {
                 schema: { type: "string" },
                 required: true,
               },
-              { $ref: "#/components/parameters/query/utm_source" },
-              { $ref: "#/components/parameters/query/utm_email" },
-              { $ref: "#/components/parameters/query/utm_campaign" },
-              { $ref: "#/components/parameters/path/version" },
```
duncanbeevers, line 111: I feel like these refs were artificially-constructed to match the expected ts output, rather than extracted from a real OpenAPI ref.

Therefore, now that the parameter `in` is respected, I made them more realistic, and adjusted the custom resolver and expected test output.

drwpow: Yup that seems right. I‘ve operated off the assumption that parameters in, say, `query` vs `path` are allowed to share names. But here in `#/components/parameters` you’re right there’s no such thing as “query” or “path”

drwpow (approved):
This was such a  tricky bug, great job! Needs a changeset though to bump the release version (patch?)

drwpow:
I’ll merge and add a manual changeset (for this and a few other PRs)

PR #2205, opened 2025-03-17
fix(openapi-fetch): prevent body parsing for HEAD requests regardless of Content-Length

### LINKED ISSUE #2195: HEAD method with non-zero Content-Length triggers JSON parsing error
### openapi-fetch version

0.13.0

### Description

HEAD method request returns 200 status code but triggers React Query's onError handler with the following error:
> onError called SyntaxError: Failed to execute 'json' on 'Response': Unexpected end of JSON input 

I'm not sure if this is the root cause, but looking at the code:
1. `head.test.ts` only tests with `Content-Length: "0"` case
2. In `src/index.js`, line 205-208:
```javascript
// handle empty content
if (response.status === 204 || response.headers.get("Content-Length") === "0") {
  return response.ok ? { data: undefined, response } : { error: undefined, response };
}
```
According to [RFC 9110 Section 9.3.2](https://httpwg.org/specs/rfc9110.html#HEAD):

> The HEAD method is identical to GET except that the server MUST NOT send content in the response. The metadata contained in the HTTP headers in response to a HEAD request SHOULD be identical to the information sent in response to a GET request.

Currently, when Content-Length is non-zero, the code attempts to parse the response body causing an error. We need to modify this to properly handle HEAD requests regardless of the Content-Length value.

### Reproduction

1. Modify head.test.ts:
   - Change Content-Length from "0" to any non-zero value (e.g., "1234")
   - Test will fail with: `SyntaxError: Failed to execute 'json' on 'Response': Unexpected end of JSON input`

Image

2. Using React Query:
"@tanstack/react-query": "^5.50.1",
```typescript
const client = createClient({
  baseUrl: 'your-api-url'
});

const query = useQuery({
  queryKey: ['test'],
  queryFn: () => client.HEAD('/some-endpoint')
});

// Even with 200 response:
// - query.isError: true
// - onError is triggered with: SyntaxError: Failed to execute 'json' on 'Response': Unexpected end of JSON input
```

### Expected result

- HEAD method responses should not attempt body parsing
- 200 responses should be treated as successful
- React Query onSuccess handler should be called

Current behavior:
- SyntaxError due to body parsing attempt
- onError handler triggered despite successful response

Additional references:
[MDN Web Docs - HEAD method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD)
> a HEAD request can read the [Content-Length](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Length) header to check the file size before downloading the file with a GET.

### Extra

live-small:
## Changes

- This PR fixes how HEAD method requests are handled in openapi-fetch by preventing body parsing regardless of the Content-Length header value.

- [According to RFC 9110,](https://httpwg.org/specs/rfc9110.html#HEAD) HEAD responses should have the same metadata as GET but must not include a body. 
Previously, the library only handled this correctly when Content-Length was "0", causing JSON parsing errors in other cases.

- Fixes #2195.

## How to Review

- Check that the condition now properly skips body parsing for HEAD requests
- Verify the new test case ensures HEAD requests with non-zero Content-Length are handled correctly

## Checklist

- [x] Unit tests updated
- [ ] `docs/` updated (if necessary) - N/A
- [ ] `pnpm run update:examples` run (only applicable for openapi-typescript) - N/A

### DIFF
--- a/.changeset/breezy-bottles-mix.md
+++ b/.changeset/breezy-bottles-mix.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Fix HEAD method requests to prevent body parsing regardless of Content-Length header value
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -209,7 +209,7 @@ export default function createClient(clientOptions) {
     }
 
     // handle empty content
-    if (response.status === 204 || response.headers.get("Content-Length") === "0") {
+    if (response.status === 204 || request.method === "HEAD" || response.headers.get("Content-Length") === "0") {
       return response.ok ? { data: undefined, response } : { error: undefined, response };
     }
 
--- a/packages/openapi-fetch/test/http-methods/head.test.ts
+++ b/packages/openapi-fetch/test/http-methods/head.test.ts
@@ -13,4 +13,16 @@ describe("HEAD", () => {
     await client.HEAD("/resources/{id}", { params: { path: { id: 123 } } });
     expect(method).toBe("HEAD");
   });
+
+  test("handles HEAD requests with non-zero Content-Length without parsing the body", async () => {
+    const client = createObservedClient<paths>({}, async () => {
+      return new Response(null, {
+        headers: { "Content-Length": "42", "Content-Type": "application/json" },
+        status: 200,
+      });
+    });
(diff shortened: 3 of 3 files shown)


drwpow (approved):
Great, thank you!

PR #2235, opened 2025-03-28
React query handle 204 or zero content length

wheelebin:
## Changes

I created an issue about how 204 status codes are handled and an error that's thrown by react-query due to data being returned as undefined (Something that react-query does not support for it's queryFn's).

Issue: https://github.com/openapi-ts/openapi-typescript/issues/2234

This PR handles 204 or zero Content-Length header by returning data as null in the queryFn.

## How to Review

Use the useQuery hook towards an API endpoint that returns a 204 status or zero value in the Content-Length header, the data should be set to null and no error should be thrown.

## Checklist

- [x] Unit tests updated


drwpow (approved):
Thank you! 204 responses are still tricky for folks, and agree this is better behavior.

Sorry for the delay in reviewing. But if you add a `patch` changeset (see GitHub comment), I‘d love to merge & release.

PR #2241, opened 2025-04-02
Replacing `any` with `unknown` in the `ReadonlyArray` type declaration.

stanleyk:
## Changes

The current declaration of the enhanced `ReadonlyArray` type uses the [undesired](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any) `any` type instead of the more appropriate `unknown` type.

- [x] Unit tests updated

### DIFF
--- a/.changeset/wet-cougars-behave.md
+++ b/.changeset/wet-cougars-behave.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Replacing `any` with `unknown` in the `ReadonlyArray` type declaration.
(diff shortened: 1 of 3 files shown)


gzm0 (changes requested):
Heya! Thanks a lot for the contribution. Definitely happy to see all these any's go :)

- Just to double check my understanding: This is not a breaking change, correct?
- Please add a changeset so we get a package version bump and an entry in the changelog. ([Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).)

stanleyk:
Thanks, @gzm0 , I have added the changeset. You are right, this should not break anything.

gzm0 (approved):
LGTM! TY!

PR #2246, opened 2025-04-08
feat: Attach fetch response to SWR data/error

Tommos0:
## Changes
It is useful to have the original fetch response available in the hook result. For instance to distinguish the HTTP status code in case of an `error`. Since SWR only returns `error` / `data` properties, this provides an escape hatch to be able to access to original `fetch` response, as returned by the `openapi-fetch` client.

https://github.com/openapi-ts/openapi-typescript/issues/2238

## Checklist

I was unable to run the test suite, making it hard to add / update unit tests for me.

### DIFF
--- a/packages/swr-openapi/src/query-base.ts
+++ b/packages/swr-openapi/src/query-base.ts
@@ -4,6 +4,8 @@ import type { Fetcher, SWRHook } from "swr";
 import type { TypesForGetRequest } from "./types.js";
 import { useCallback, useDebugValue, useMemo } from "react";
 
+export const RESPONSE = Symbol.for('response');
+
 /**
  * @private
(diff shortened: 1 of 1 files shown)


drwpow (changes requested):
Thanks! Not opposed to this fix itself, but we will need a test to make sure this doesn’t regress in the future.

Could you describe the error you were seeing when running the test suite? Are you on the current LTS Node version (22) and did you follow the instructions in `CONTRIBUTING.md`?

drwpow:
Just a heads up: the repo has since been moved here, so this PR will need to be moved to changeset-release/main. Please feel free to repoen there! (But the test requirements will be the same)

PR #2248, opened 2025-04-09
feat(openapi-typescript): add jsdoc comments to response object

backflip:
## Changes

This PR adds JSDoc comments like `@example` to response objects. We are missing them in our generated types. They are already present on request objects and in many other places.

## How to Review

See changes in `examples/digital-ocean-api.ts` below.

## Checklist

- [ ] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/examples/digital-ocean-api.ts
+++ b/packages/openapi-typescript/examples/digital-ocean-api.ts
@@ -17215,6 +17215,10 @@ export interface components {
                 [name: string]: unknown;
             };
             content: {
+                /** @example {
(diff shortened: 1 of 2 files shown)


gzm0 (approved):
TY very much for the contribution.

Review note: I have verified that the added examples appear verbatim in the original OpenAPI (and are not generated). As such, I am not worried about code bloat or duplicated examples. It is up to the OpenAPI author to decide at which level to add examples (object or object field, etc.).

gzm0:
Dang... We should have included a changeset here :(

PR #2250, opened 2025-04-10
fix: get post types with "post" option

etc-tiago:
## Changes

Change the type of return

## How to Review

Generate a post open-api file and filter by path.

## Checklist

- [x] Unit tests updated


drwpow:
Thanks for opening a PR! But as there’s no issue attached, no reasoning, and the tests are failing, I’m going to have to reject this change.

If you could open an issue first (or refer to a currently-existing issue) so we are clear on why this change needs to happen, we’d be open to a new PR! But we’ll also need to make sure the test suite passes.

PR #2253, opened 2025-04-13
feat(openapi-fetch): add transform options for response data handling

Newbie012:
## Changes

This PR adds the option to transform the response data. This is useful when you want to deserialize the returned output. This is not achievable via middleware.

## How to Review

I would personally start from reading the test. The PR is quite minimal.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/docs/openapi-fetch/api.md
+++ b/docs/openapi-fetch/api.md
@@ -19,6 +19,7 @@ createClient<paths>(options);
 | `fetch`           | `fetch`         | Fetch instance used for requests (default: `globalThis.fetch`)                                                                          |
 | `querySerializer` | QuerySerializer | (optional) Provide a [querySerializer](#queryserializer)                                                                                |
 | `bodySerializer`  | BodySerializer  | (optional) Provide a [bodySerializer](#bodyserializer)                                                                                  |
+| `transform`       | TransformOptions| (optional) Provide [transform functions](#transform) for response data                                                          |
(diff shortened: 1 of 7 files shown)


drwpow (changes requested):
Thanks for submitting! The “this is not achievable in middleware” callout is a great observation, and would also like to fix. Thank you for the proposal!

But since this is a bigger change, I’d like to pause on this and back up to a proposal first. Please **Open an issue** that has a more in-depth proposal for how you’d like to fix this flaw in middleware, and discuss some tradeoffs.

Adding another API that is 90% similar to what the middleware is doing now, but with some minor differences, isn’t a cohesive API design because it gives users multiple ways to accomplish the same goals.

Instead, I’d love to see a proposal that goes more in depth on:

- How is the current middleware API sufficient?
- What serialization/deserialization is needed?
- What are some alternatives—how might the middleware API be modified to accomplish these goals?

Thinking through this more holistically will benefit more users, and end up as a stronger design. To that end, would love to align more on the plan so we fix the real problems you’re pointing out, but in a way that improves everyones’ experience.

Newbie012:
Sorry for the late response, I initially ended up doing:

```ts
async onResponse({ response }) {
    const cloned = response.clone();
    const json = await cloned.json();

    // ... deserialize via superjson ...

    return Object.assign(cloned, {
      json: () => Promise.resolve(json),
    });
  },
  ```
  
  and then ended up passing everything through a wrapper.
  
  Closing this PR as I don't need this feature anymore. Thanks for looking into this.

PR #2256, opened 2025-04-15
fix(openapi-fetch): dynamic access to methods using wrapAsPathBasedClient

KajSzy:
## Changes

When using `wrapAsPathBasedClient` to access methods of client dynamically it was throwing an error as `this` context was not assigned to `PathCallForwarder` class.
Rewriting `PathCallForwarder` to use arrow function fixes the issue.

## How to Review

I've added a test case which has example how `wrapAsPathBasedClient` can be used in generic function to retrieve methods, which also verifies the fix.

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/heavy-onions-bake.md
+++ b/.changeset/heavy-onions-bake.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+fix dynamic access to methods using wrapAsPathBasedClient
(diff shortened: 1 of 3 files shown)


drwpow (approved):
Good fix, and great test, thank you!

drwpow:
Not sure what happened with a couple cancelled tests, but all the tests that ran passed. The skipped ones shouldn’t matter.

Issue #2266, opened 2025-04-29
TypeError: Cannot read properties of null (reading 'useContext')


jean-smaug:
### openapi-react-query version

0.3.1

### Description

I'm building a React packages in order to share clients initialization across multiple apps
Inside those apps I try to make unit tests with Jest and I encountered the error in title `TypeError: Cannot read properties of null (reading 'useContext')`

Technical context : 
- Monorepo managed with PNPM `10.8.1`
- Node `v22.14.0`
- jest `29.7.0`
- Transpiler for unit tests `@swc/jest` in `0.2.37`

![Image](https://github.com/user-attachments/assets/04492e2f-a551-4e2f-b737-ceb2f9c71aa2)

![Image](https://github.com/user-attachments/assets/88a999b6-e3a3-496d-b926-f70c039f5495)

There are some issues, on invalid calls of hooks too 

![Image](https://github.com/user-attachments/assets/7e4c08a2-a56c-4164-be22-b3839a08a881)

### Reproduction

import createFetchClient from "openapi-fetch";
import createClient from "openapi-react-query";
import { QueryClient, QueryClientProvider } from "@tanstack/react-query";

import { render } from "@testing-library/react";
import type MetadataService from "../../types/MetadataService";

const metadataClient = createClient(createFetchClient<MetadataService.paths>());

const TechnicalProof = () => {
  const { data: countries } = metadataClient.useQuery("get", "/v1/countries");

  console.log("🚀 ~ TechnicalProof ~ countries:", countries);

  return <div>Hello</div>;
};

describe("Example", () => {
  it("should render a component that use SDKs", () => {
    render(
      <QueryClientProvider client={new QueryClient()}>
        <TechnicalProof />
      </QueryClientProvider>,
    );
  });
});

### Expected result

Understand this issue

### Extra

- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/openapi-ts/openapi-typescript/blob/main/packages/openapi-react-query/CONTRIBUTING.md))

mynnx:
Yeah, we're running into this too.  Currently troubleshooting.

caribouflex:
I’m running into a similar issue here. 

At first, I thought it might be because the monorepo had different versions of Tanstack/react-query, but I double-checked and everything looks good. 

Using Tanstack’s useQuery directly works fine, but when I try using the client from openapi-react-query, I get the same error as mention above. 

[...]

PR #2292, opened 2025-05-05
feat(openapi-react-query): Add query client to create client

arnaud9145:
## Changes

This PR is similar to https://github.com/openapi-ts/openapi-typescript/pull/1814, but instead of adding the query client each time we call useQuery/useMutation/useSuspenseQuery/useInfiniteQuery, we pass the queryClient to the createClient once and for all. It is still possible to pass an other queryClient to those methods to override the default one if needed.

## How to Review

Create a client by passing a queryClient, check that it is this client that is used instead of the default one.

## Checklist

- [X] Unit tests updated
- [X] `docs/` updated (if necessary)


eyestorefivemstore:
Interesting solution to this problem! I've been working on similar challenges recently.

eyestoreoye:
Interesting solution to this problem! I've been working on similar challenges recently.

jean-smaug (approved):
👏 
Think it should fix #2266

PR #2308, opened 2025-05-10
Improve header handling

drwpow:
## Changes

Followup from #2096, also passes in mergedHeaders to the bodySerializer so we’re not dropping user-set headers.

## How to Review

- Tests should pass

## Checklist

No tests updated; this should be an invisible change

### DIFF
--- a/.changeset/bumpy-parrots-lie.md
+++ b/.changeset/bumpy-parrots-lie.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Improve header handling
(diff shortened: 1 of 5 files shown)


toofishes:
As a heads up, `index.d.ts` was not updated for this, even though the body serializer function type signature has changed.

gzm0:
@toofishes thanks a lot for the heads-up! I assume this causes some form of bug. If you're able to confirm that, a bug report would be very much appreciated!

gzm0:
Actually, are you sure this PR is the culprit? It seems like indeed, `BodySerializer` is missing the `header` argument:

https://github.com/openapi-ts/openapi-typescript/blob/9c71799bcd1ca8d9bc5f3c7fc5ee973c80134b38/packages/openapi-fetch/src/index.d.ts#L65

However, this PR doesn't seem to change that; it just changes which headers are passed to `bodySerializer`. Am I missing something?

PR #2310, opened 2025-05-10
Build packages with unbuild to improve CJS support

drwpow:
## Changes

[Unbuild](https://github.com/unjs/unbuild) is a fantastic packaging system that is the best available in my opinion. Build all our packages with it to reduce maintenance and overhead, and improve support. It’s still ESM-first, so there should be no changes to runtime, client weight, or bundlesize downstream, and only improves stability and support.

This also aligns openapi-typescript, openapi-fetch, openapi-react-query, and swr-openapi to all be built the same way.

If this causes any issues, please file an issue! Anything that crops up is fixable with the Unbuild system because it’s based on Rollup.

## How to Review

- Tests cover existing usecases (we often test built code)!

## Checklist

- [x] Unit tests updated


packages/openapi-fetch/package.json
```diff
@@ -8,19 +8,12 @@
...
-  "types": "./dist/index.d.ts",
+  "main": "./dist/index.mjs",
   "exports": {
     ".": {
-      "import": {
-        "types": "./dist/index.d.ts",
-        "default": "./dist/index.js"
-      },
-      "require": {
-        "types": "./dist/cjs/index.d.cts",
```
drwpow, line 21: With separate `.d.cts` and `.d.mts` declarations generated, we don’t need to declare `types` separately. This was a little bit of a hack from previous builds that stuck around. With unbuild’s better generation in general, TypeScript can automatically locate types.

packages/openapi-fetch/package.json
```diff
@@ -8,19 +8,12 @@
   },
   "license": "MIT",
   "type": "module",
-  "main": "./dist/index.js",
-  "module": "./dist/index.js",
```
drwpow, line 12: `"module"` has never been standardized, but was just a common community convention—common enough it started to sneak into some bundle setups. All Node LTS versions and all bundlers I’m aware of now respect `"exports"`, which means this is no longer needed.

`"main"` is still standard. And technically it’s superceded by `exports`, but we’ll just keep it around for safety. But `"module"` is safe to remove at this point.

packages/openapi-fetch/package.json
```diff
@@ -8,19 +8,12 @@
...
-  "types": "./dist/index.d.ts",
```
drwpow, line 13: Same for `"types"`—at this point TypeScript can now just locate the types due to unbuild’s careful handling

packages/openapi-typescript/package.json
```diff
@@ -11,12 +11,14 @@
...
+  "main": "./dist/index.mjs",
   "exports": {
     ".": {
-      "import": "./dist/index.js",
-      "require": "./dist/index.cjs"
+      "import":  "./dist/index.mjs",
+      "require": "./dist/index.cjs",
+      "default": "./dist/index.mjs"
     },
+    "./*.js": "./*.mjs",
```
drwpow, line 21: Unbuild intentionally ships `.mjs` so there’s no ambiguity about the `.js` files being ESM, and there’s no way to opt out of this behavior. While in our case we’ve been shipping ESM for a while, I still agree in general—`.js` files is a bit of a clusterfuck trying to sniff out `package.json`s to know how to parse it.

This is encroaching on a breaking change for this library, but I’m just considering it a minor version because:

- Deep imports have never had a strict guarantee they’ll be preserved perfectly
- This extension rewriting should preserve `.js` imports as-authored
- We’re still building files 1:1 to their original versions, just with `.mjs` extensions, so even if a setup does break it’s a mild annoyance not a show-stopper

packages/openapi-fetch/package.json
```diff
@@ -47,19 +40,18 @@
...
+    "lint:ts-no-strict": "tsc --noEmit -p test/no-strict-null-checks/tsconfig.json",
     "generate-types": "openapi-typescript -c test/redocly.yaml",
+    "prepack": "pnpm run build",
     "pretest": "pnpm run generate-types",
-    "test": "pnpm run \"/^test:/\"",
+    "test": "pnpm run test:js && pnpm run test:exports",
     "test:js": "vitest run",
-    "test:ts": "tsc --noEmit",
-    "test:ts-no-strict": "tsc --noEmit -p test/no-strict-null-checks/tsconfig.json",
+    "test:exports": "pnpm run build && attw --pack .",
```
drwpow, line 54: I really like the “are the types wrong” type tests, so added to all packages. We were also able to remove the `cjs-resolves-to-esm` rule 😎

packages/openapi-react-query/.npmignore
```diff
@@ -1,4 +1,5 @@
 .turbo
+*.config.*
```
drwpow, line 2: Ignore `build.config.ts` configs for unbuild in package publishing

packages/openapi-metadata/build.config.ts
```diff
@@ -0,0 +1,20 @@
...
+    "./src/ui/index.ts",
+  ],
+  declaration: "compatible",
+  clean: true,
+  sourcemap: true,
+  rollup: {
+    // Ship CommonJS-compatible bundle
+    emitCJS: true,
+    // Don’t bundle .js files together to more closely match old exports (can remove in next major)
+    output: { preserveModules: true },
```
drwpow, line 18: openapi-metadata was already using unbuild! But adding this config just shores it up more closely with the other packages without changing output too much.

drwpow:
Merging this not to bulldoze anyone; this is just a very unopinionated PR that is just package maintenance. The change does require a version-bump just to guard against accidental regressions, but anything that crops up is fixable/changeable.

Overall this should be a net win for everybody, improving CJS support and build quality.

PR #2332, opened 2025-05-25
openapi-fetch: `Request` parameter in client methods was ignored

sgrishchenko:
## Changes

Small bag is fixed. Before these changes `Request` parameter in client methods was ignored. 

Also, it turned out that `Request` parameter was not added in `RequestOptions`, so I added the corresponding declaration. 

Also, it turned out that `Request` parameter was not added in API docs, so I tried to add it there. 

## How to Review

I provided a unit test to demonstrate the problem, you can run it with and without my fix and check the results.

## Checklist

- [x] Unit tests updated


drwpow:
Sorry for the delay in reviewing. We’ve had some browser-specific bugs here when trying to work with Requests. Normally we’d ask to add tests to the browser tests as well, because believe-it-or-not, browsers do have some longstanding bugs in the Request API, and they differ from Node.js (see test suite for more details).

However, in the time we’ve taken to review (which is 100% on us), we’re also [putting openapi-fetch into maintenance mode](https://github.com/openapi-ts/openapi-typescript/discussions/2559) to focus on the core libraries. So unfortunately I’ll have to close this PR, albeit tiny, as not acceptable in the last few releases before putting into maintenance mode. But we do appreciate the contribution, and hope you’ll contribute to the core library further!

PR #2351, opened 2025-06-08
fix(transform): avoid duplicate enum declaration when using root types with no prefix

Newbie012:
## Changes

given the following settings:

```
enum: true,
rootTypes: true,
rootTypesNoSchemaPrefix: true
```

the generated output will declare an enum and a type with the same name. This PR will avoid redeclaring a type if its node is an enum.

## How to Review

The PR is quite minimal - code adjustment + a test

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/src/transform/components-object.ts
+++ b/packages/openapi-typescript/src/transform/components-object.ts
@@ -80,6 +80,9 @@ export default function transformComponentsObject(componentsObject: ComponentsOb
           }
           const ref = ts.factory.createTypeReferenceNode(`components['${key}']['${name}']`);
           if (ctx.rootTypesNoSchemaPrefix && key === "schemas") {
+            if (ctx.enum && item.enum) {
+              continue;
+            }
             aliasName = aliasName.replace(componentKey, "");
           }
           const typeAlias = ts.factory.createTypeAliasDeclaration(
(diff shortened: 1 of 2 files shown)


gzm0:
@duncanbeevers can I hand this over to you? I'm not feeling sufficiently comfortable with the internals to review this.

duncanbeevers:
@gzm0 Yes I can take this. I think it's a duplicate of a couple other issues, so I'll see about taking some time to gather the various bugs + fixes together to pick the best solution.

iw-tundo:
Hi, I'm also encountering this issue. Do you know if you have a rough estimate of when this will be merged and released?
@duncanbeevers

drwpow:
@Newbie012 could you add a `patch` changeset with a note on the fix? If you add one I’d be happy to merge this as-is. Thanks!

fluctus:
@Newbie012 @gzm0 @duncanbeevers @sultaniman This is a duplicate of [this PR](https://github.com/openapi-ts/openapi-typescript/pull/2206) which was not yet merged; you can read the discussion there but the general idea I had is that we should avoid generating root types for enums when these flags are used together, instead of generating an exported type with Schema prefix just for enums, as it would be confusing. WDYT?

Newbie012:
self closing this PR due to inactivity 🙂

PR #2375, opened 2025-07-14
Enum with Root Types Filtering Recursive Reference

ParkerVR:
## Changes

Using the --enum and --root-types in combination causes recursive reference to enum types used in schemas, breaking the file.
A few issues related:

https://github.com/openapi-ts/openapi-typescript/issues/2099
https://github.com/openapi-ts/openapi-typescript/issues/2341 

## How to Review

A specific combination of flags causes recursive imports, marked as patch since it seems incredibly unlikely anyone could successfully use this output in a production system. Had a custom in my project and was trying to use the flags instead, but the combined behavior broke me. https://github.com/Equal-IQ/api-model/blob/e99585b6244ac814ea1044f62333474a5857c3b8/Containers/generate.js 

## Checklist

- [X] Unit tests updated (CJS Test is broken before and after my change, unrelated)
- [] `docs/` updated (if necessary) don't think any change is required, as this is a reasonable expected behavior
- [] `pnpm run update:examples` run (only applicable for openapi-typescript) no changes required to these, since the combination is not used for them


ParkerVR:
I added some snapshot tests here since this change is a bit particular to CLI Flag Combination, though I guess it could be inlined instead like some of the other partial api spec examples. I'll leave it to maintainers to decide on that

drwpow:
> I added some snapshot tests here since this change is a bit particular to CLI Flag Combination, though I guess it could be inlined instead like some of the other partial api spec examples. I'll leave it to maintainers to decide on that

All tests are great! My take is this is a maintainer concern shuffling tests around. Them just existing in a PR is all that matters, thank you!

PR #2393, opened 2025-08-08
Enum generation descriptions - allow for nulls/empty strings

simon-curtis:
## Changes

Expands the condition for skipping enum descriptions to also ignore null and empty strings.

## Details
Since NSwag.AspNetCore v14.5, it seems that a descriptions array is always generated for enums, with null values for any undefined entries. This appears to be intended to maintain positional alignment with enum values, and there’s no built-in way to override it due to being a private field. This change ensures such null or empty entries are not appended.

I can't see anything in the spec that would suggest that null is an invalid value for a description but at the moment you get a null reference exception:

```
file:///D:/git/cma/main/Technical/SourceCode/TestHarness%202.0/TestHarness.UI/node_modules/openapi-typescript/dist/lib/ts.mjs:271
    " ".concat(metadata.description.trim()),
                                    ^

TypeError: Cannot read properties of null (reading 'trim')
    at tsEnumMember (file:///D:/git/cma/main/Technical/SourceCode/TestHarness%202.0/TestHarness.UI/node_modules/openapi-typescript/dist/lib/ts.mjs:271:37)
```

## How to Review

Give it an openapi spec like this and check that the process runs to completion with no errors and that no descriptions are generated.

```json
{
  "schemas": {
    "ProjectType": {
      "type": "integer",
      "description": "",
      "x-enumNames": [
        "Normal",
        "Snapshot"
      ],
      "x-enum-descriptions": [
        null,
        null,
      ],
      "enum": [
        0,
        1
      ]
    }
  }
}
```

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/fluffy-pumas-run.md
+++ b/.changeset/fluffy-pumas-run.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Now checking for null & empty string when generating enum description


drwpow:
This fix looks good! Could you add a `patch` changeset? We could merge & release with that!

PR #2407, opened 2025-08-12
Fix Request parameter being ignored by client methods

jthacker:
## Changes

This fixes a bug in openapi-fetch where the client methods would ignore the `Request` parameter being passed in. Previously passing in the `Request` parameter works when constructing the client but not if passed in via `GET`, `POST`, etc. on a constructed client.

Related to but not blocking https://github.com/openapi-ts/openapi-typescript/issues/2224

## How to Review

No impact to the API, adds tests for checking the functionality in both cases.

## Checklist

- [X] Unit tests updated
- [ ] N/A ~`docs/` updated (if necessary)~
- [ ] N/A ~`pnpm run update:examples` run (only applicable for openapi-typescript)~

### DIFF
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -109,7 +109,7 @@ export default function createClient(clientOptions) {
 
     let id;
     let options;
-    let request = new CustomRequest(
+    let request = new Request(
       createFinalURL(schemaPath, { baseUrl: finalBaseUrl, params, querySerializer }),
       requestInit,
     );
@@ -143,7 +143,7 @@ export default function createClient(clientOptions) {
             id,
           });
           if (result) {
-            if (result instanceof CustomRequest) {
+            if (result instanceof Request) {
               request = result;
             } else if (result instanceof Response) {
               response = result;
--- a/.changeset/empty-rooms-attack.md
+++ b/.changeset/empty-rooms-attack.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Fix Request parameter being ignored by client methods


gzm0:
Note: It seems this has been broken since the introduction of the feature: https://github.com/openapi-ts/openapi-typescript/pull/1907

packages/openapi-fetch/test/common/create-client.test.ts
```diff
@@ -57,6 +57,47 @@ describe("createClient options", () => {
     expect(actualURL.href).toBe("https://api.foo.bar/v2/resources");
   });
 
+  test("Request option is used", async () => {
```
gzm0, line 60: There is already a test for the client `test/common/request.test.ts`. Could you move (and probably align) the test for the case this PR is fixing there, please?

packages/openapi-fetch/src/index.js
```diff
@@ -109,7 +109,7 @@ export default function createClient(clientOptions) {
 
     let id;
     let options;
-    let request = new CustomRequest(
+    let request = new Request(
```
gzm0, line 112: `CustomRequest` is also used in the middleware application lower down to perform an instanceof check. I think we should adjust that check as well for consistency.

The middleware specification in this case is arguably a bit broken, but I'd rather have that failing, than allowing for the inconsistent check. In the sense: At the moment, an onRequest middleware is required to return the same type of request than we send. If we do not adjust the check in this PR, it is required to return the type of the request configured on the client, even if a different request was provided. IMO that's very odd.

gzm0 (changes requested):
Good catch and thanks for the fix. I've left some comments.

jthacker:
@gzm0 Thanks for the feedback. Changes pushed. Now testing for all methods and removed the duplicate test for setting `Request` via the client constructor.

packages/openapi-fetch/test/common/request.test.ts
```diff
@@ -316,6 +316,30 @@ describe("request", () => {
...
+        expect(input).instanceOf(SpecialRequestImplementation);
+        return Promise.resolve(Response.json({ hello: "world" }));
+      };
+
+      const client = createClient<paths>({
+        baseUrl: "https://fakeurl.example",
+        fetch: customFetch,
+      });
+
+      await (client as any)[method]("/resources", { Request: SpecialRequestImplementation });
```
gzm0, line 339: Uh, oh. With the  `as any`, we don't actually test if the typings allow passing the `{ Request: SpecialRequestImplementation }` (because with `any` anything will be allowed).

Why is it necessary?

If it is difficult to get the typing to work with `test.each`, it feels like it might also just test a single method.

gzm0 (changes requested):
Thanks for the iteration! Looks great. Just one comment about the typing in the tests.

packages/openapi-fetch/test/common/request.test.ts
jthacker: The reason I added the type assertion was due to an error about the signatures not being compatible.
```
This expression is not callable.
  Each member of the union type 'ClientMethod<paths, "get", `${string}/${string}`> | ClientMethod<paths, "put", `${string}/${string}`> | ClientMethod<paths, "post", `${string}/${string}`> | ClientMethod<...> | ClientMethod<...> | ClientMethod<...> | ClientMethod<...>' has signatures, but none of those signatures are compatible with each other.ts(2349)
```

It actually does seem to work, I get 7 failures without the fix implemented, here's a sample
```
 FAIL  test/common/request.test.ts > request > custom Request > uses custom Request class - PATCH
AssertionError: expected Request{} to be an instance of SpecialRequestImplementation
 ❯ customFetch test/common/request.test.ts:330:23
    328|       const customFetch = async (input: Request) => {
    329|         // make sure that the request is actually an instance of the custom request we provided
    330|         expect(input).instanceOf(SpecialRequestImplementation);
       |                       ^
    331|         return Promise.resolve(Response.json({ hello: "world" }));
    332|       };
 ❯ coreFetch src/index.js:162:26
 ❯ Object.PATCH src/index.js:283:14
 ❯ test/common/request.test.ts:339:35
```

I'm not an expert in TypeScript though so happy to adjust as desired. Would you prefer the more explicit non type asserted single method version? I.e. `client.GET(...)`? Based on the implementation this should have the same result as the current testing approach.

gzm0 (approved):

PR #2419, opened 2025-08-17
feat(openapi-metadata): Update Scalar UI integration and add Scalar UI configuration options

Ackuq:
## Changes

This updates the Scalar UI integration to follow the default implementation from Scalar: https://guides.scalar.com/scalar/scalar-api-references/integrations/htmljs.

This also adds an optional configuration option when generating the UI, to e.g. customize theme or proxy.

## How to Review

Simplest way to try this out is to locally link with a bare-bones service, e.g. an Express service and try out the options.

```js
import express from "express";
import { generateScalarUI } from "openapi-metadata/ui";

const app = express();

app.get("/scalar", async (_req, res) => {
  const ui = generateScalarUI(
    "https://petstore3.swagger.io/api/v3/openapi.json",
    {
      theme: "kepler",
      hideModels: true,
      hideDarkModeToggle: true,
    }
  );
  res.send(ui);
});

app.listen("3000", () => {
  console.log(`Server is running on port 3000`);
});
```

## Checklist

- [ ] Unit tests updated

### DIFF
--- a/packages/openapi-metadata/src/ui/scalar.ts
+++ b/packages/openapi-metadata/src/ui/scalar.ts
@@ -14,32 +14,32 @@ export type GenerateScalarUIOptions = {
   searchHotKey?: string;
   servers?: any[];
   showSidebar?: boolean;
-  theme?: 
-    | 'alternate'
-    | 'default'
-    | 'moon'
-    | 'purple'
-    | 'solarized'
-    | 'bluePlanet'
-    | 'deepSpace'
-    | 'saturn'
-    | 'kepler'
-    | 'elysiajs'
-    | 'fastify'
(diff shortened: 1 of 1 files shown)


drwpow:
Thanks for contributing, but unfortunately [we won’t be releasing future versions of openapi-metadata as discussed on our roadmap](https://github.com/openapi-ts/openapi-typescript/discussions/2559). The TL;DR is we’re spread too thin to give this library the time and attention it needs. So with no releases planned, I’ll have to close this PR.

That said, this is MIT code! You can fork the existing copy, release your own version, etc. But unfortunately this maintainer team for now has to focus on upholding one good library instead of too many half-baked ones.

PR #2429, opened 2025-09-01
docs: fix typo in example

evermake:
This PR fixes a typo in example of using `application/x-www-form-urlencoded` content type.


gzm0 (approved):
TY

PR #2508, opened 2025-10-19
fix(openapi-fetch): use text() when no content-length is provided to avoid errors parsing empty response (200 with no content)

srbarba:
## Changes

Fixes an issue where responses with HTTP 200 status and no `Content-Length` header would fail when parsing JSON, even when the response body is empty.

**Problem**: When a server returns a successful response (200 OK) without a `Content-Length` header and an empty body, calling `response.json()` directly throws an error because there's no content to parse. This is a valid HTTP scenario but was causing runtime errors.

**Solution**: Added a check for missing `Content-Length` headers when `parseAs: "json"`. The fix:
1. First reads the response as text using `response.text()`
2. Then conditionally parses it as JSON only if content exists
3. Returns `undefined` if the response is empty

## How to Review

1. **Review the logic**: Check the added condition in `index.js` around line 245 that handles the case when `parseAs === "json" && !contentLength`
2. **Run the test**: The new test `"json with empty response"` in `response.test.ts` verifies the fix works correctly
3. **Verify no regressions**: All existing tests should pass, confirming backward compatibility
4. **Consider edge cases**: The fix only applies when:
   - Response is successful (2** except 204)
   - `parseAs` is set to `"json"` (default)
   - `Content-Length` header is absent (not just `"0"`)

Key review points:
- Empty responses now return `{ data: undefined }` instead of throwing an error
- Responses with `Content-Length: "<any>"` are still handled by the existing logic
- The fix doesn't affect error responses or other `parseAs` options

## Checklist

- [x] Unit tests updated
- [ ] `docs/` updated (if necessary) - No docs changes needed, internal bug fix
- [ ] `pnpm run update:examples` run (only applicable for openapi-typescript) - N/A, this is openapi-fetch

### DIFF
--- a/.changeset/wet-waves-turn.md
+++ b/.changeset/wet-waves-turn.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+Use text() when no content-length is provided to avoid errors parsing empty bodies (200 with no content)
--- a/packages/openapi-fetch/src/index.js
+++ b/packages/openapi-fetch/src/index.js
@@ -230,18 +230,29 @@ export default function createClient(clientOptions) {
       }
     }
 
+    const contentLength = response.headers.get("Content-Length");
     // handle empty content
-    if (response.status === 204 || request.method === "HEAD" || response.headers.get("Content-Length") === "0") {
+    if (response.status === 204 || request.method === "HEAD" || contentLength === "0") {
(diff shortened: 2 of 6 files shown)


srbarba:
Related issue: https://github.com/openapi-ts/openapi-typescript/issues/1977

PR #2526, opened 2025-11-20
fix(helpers): handle any type in RequiredKeysOf

jtomaszewski:
## Problem

When using `createClient<any>()` with placeholder or fallback OpenAPI types, TypeScript incorrectly requires the second parameter in API calls:

```typescript
// With placeholder types: export type paths = any;
const client = createClient<any>({ baseUrl: '...' });

// Error: Expected 2 arguments, but got 1
client.GET('/users/me');
```

This is problematic in scenarios where:
- Docker builds need to work without the actual OpenAPI spec file
- Using placeholder/fallback types during development
- Gradual migration to OpenAPI-based type systems

## Root Cause

The `RequiredKeysOf` type helper doesn't properly handle the `any` type. When `RequiredKeysOfHelper<T>` evaluates to `any`, the current implementation doesn't recognize it as "no required keys", causing TypeScript to treat parameters as required.

## Solution

Add an additional conditional check in `RequiredKeysOf`:

```typescript
RequiredKeysOfHelper<T> extends any ? never
```

This ensures that when the type parameter is `any`, `RequiredKeysOf` correctly returns `never`, indicating no required keys. This makes the second parameter optional in API calls.

## Impact

- ✅ Fixes type errors when using `createClient<any>()`
- ✅ Enables Docker builds with placeholder OpenAPI types
- ✅ Maintains backward compatibility for all existing use cases
- ✅ No changes to runtime behavior (types only)

Relates to #1778

### DIFF
--- a/packages/openapi-fetch/test/types.test.ts
+++ b/packages/openapi-fetch/test/types.test.ts
@@ -1,4 +1,10 @@
-import type { ErrorResponse, GetResponseContent, OkStatus, SuccessResponse } from "openapi-typescript-helpers";
+import type {
+  ErrorResponse,
+  GetResponseContent,
+  OkStatus,
+  RequiredKeysOf,
+  SuccessResponse,
+} from "openapi-typescript-helpers";
 import { assertType, describe, test } from "vitest";
 
 describe("types", () => {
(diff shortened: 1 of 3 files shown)


jtomaszewski:
FYI I've added some tests and fixed all CI checks.

PR #2527, opened 2025-11-21
feat(openapi-fetch): enable middleware request param module augmentation

Katli95:
## Changes

Decided to open PR in lieu of issue.

I just changed the `.d.ts` file manually and extracted the `MiddlewareRequestParams` into their own interface. It opens up the possibility for using TS module augmentation to add parameters to the `params` property and use those in the middleware. 

My use-case is changing authentication logic based on a variable that isn't passed through to the API being invoked.

## How to Review

Look at the file, I guess? 🤗

Also, I would be open to going deeper in a follow-up PR to look into how the client facing types could mirror the custom parameter added, if this is indeed an avenue you're open to exploring.

## Checklist
(skipped all since it's just a small TS update and I didn't pull to local)
- [ ] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -131,18 +131,20 @@ export type MergedOptions<T = unknown> = {
   fetch: typeof globalThis.fetch;
 };
 
+export interface MiddlewareRequestParams {
+  query?: Record<string, unknown>;
+  header?: Record<string, unknown>;
+  path?: Record<string, unknown>;
+  cookie?: Record<string, unknown>;
+}
+
 export interface MiddlewareCallbackParams {
   /** Current Request object */
   request: Request;
   /** The original OpenAPI schema path (including curly braces) */
   readonly schemaPath: string;
   /** OpenAPI parameters as provided from openapi-fetch */
-  readonly params: {
-    query?: Record<string, unknown>;
-    header?: Record<string, unknown>;
-    path?: Record<string, unknown>;
-    cookie?: Record<string, unknown>;
-  };
+  readonly params: MiddlewareRequestParams;
   /** Unique ID for this request */
   readonly id: string;
   /** createClient options (read-only) */


drwpow (approved):
This makes sense! Little PRs like this that don’t fundamentally change things, just provide little cleanups like this are always fine. Thank you!

sk22:
> It opens up the possibility for using TS module augmentation to add parameters to the `params` property and use those in the middleware.

Wait, I can't seem to fully get this working...

```ts
declare module 'openapi-fetch' {
	interface MiddlewareRequestParams {
		timeout?: number;
	}
}
```

I can write a timeout middleware like this:

```ts
// Timeout middleware
fetchClient.use({
	async onRequest({ request, params: { timeout } }) {
		const signal =
			timeout === Infinity ?
				request.signal
			:	AbortSignal.any([
					request.signal,
					AbortSignal.timeout(typeof timeout === 'number' && timeout >= 0 ? timeout : REQUEST_TIMEOUT),
				]);
		return new Request(request, { signal });
	},
});
```

But I can't really use my custom parameter because TypeScript will yell at me if I try to do this:

```ts
client['/api/takes-a-while'].PUT({ params: { timeout: 120_000 } })
```

> Object literal may only specify known properties, and timeout does not exist in type
> `{ query?: undefined; header?: undefined; path?: undefined; cookie?: undefined; }`. (ts 2353)

Did you somehow get around this without telling TypeScript to ignore this line (or was this just not your use case)?

Katli95:
> Did you somehow get around this without telling TypeScript to ignore this line (or was this just not your use case)?

Nooo, I was working on it in November and I had to circumvent it 😅 I never found a way to require the param on the calling side.

I used the `client.GET(url, {timeout: 20})` syntax. There timeout was allowed but not required, should be simple enough but I don't have the time at the moment.

sk22:
> I used the `client.GET(url, {timeout: 20})` syntax.

i was about to comment "unfortunately, custom options aren't passed along to middleware" - but it seems like they do, but apparently they're passed directly into the request :') that would work, thanks for your reply!

image

PR #2528, opened 2025-11-23
Add option to handle parameters with default as non-optional

luqasn:
## Changes

This PR introduces a new option that makes it possible to request that non-required parameters still be generated as non-optional (no question token) if they have a default value set in their schema.

This allows to generate types for certain server implementation use where the default value is injected automatically for the handling function to consume (I am working on one atm) and the signature on the handler should indicate that the value can not be undefined.

This makes it so that the default value can be derived and set by the framework from the openapi spec and does not need to be repeated by the person implementing the operation.

This an example of an implementation that leverages the new setting: https://github.com/henhal/openapi-ts-backend/pull/12/commits/ee87617ef148214fb8ad5c019e588f5e62923553

Fixes https://github.com/openapi-ts/openapi-typescript/issues/2319

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-typescript/src/transform/parameters-array.ts
+++ b/packages/openapi-typescript/src/transform/parameters-array.ts
@@ -1,7 +1,7 @@
 import ts from "typescript";
 import { addJSDocComment, NEVER, oapiRef, QUESTION_TOKEN, tsModifiers, tsPropertyIndex } from "../lib/ts.js";
 import { createRef } from "../lib/utils.js";
-import type { ParameterObject, ReferenceObject, TransformNodeOptions } from "../types.js";
+import type { ParameterObject, ReferenceObject, SchemaObject, TransformNodeOptions } from "../types.js";
 import transformParameterObject from "./parameter-object.js";
 
 // Regex to match path parameters in URL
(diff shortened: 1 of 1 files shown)


luqasn:
Test failure seems unrelated, docly have refactored their tests and moved a bunch of files: https://github.com/Redocly/redocly-cli/commit/cf26d6038e7c75af3335164f4797133393ac84d3#diff-819418d7530fd6a1220968de72601d679939eb5a01c3b9cc9e388050bd58ff77

sacha-c:
> Test failure seems unrelated, docly have refactored their tests and moved a bunch of files: [Redocly/redocly-cli@cf26d60#diff-819418d7530fd6a1220968de72601d679939eb5a01c3b9cc9e388050bd58ff77](https://github.com/Redocly/redocly-cli/commit/cf26d6038e7c75af3335164f4797133393ac84d3#diff-819418d7530fd6a1220968de72601d679939eb5a01c3b9cc9e388050bd58ff77)

Indeed, I've address this in a separate PR a few weeks back which is awaiting review: https://github.com/openapi-ts/openapi-typescript/pull/2522

luqasn:
Thanks @sacha-c, I rebased on top of your merged commit and removed my fixes.

@duncanbeevers could you have a look? Or is there someone else I should ask for a review?

luqasn:
@drwpow could you have a look?

luqasn:
What can I do to help getting this reviewed/merged? @duncanbeevers

luqasn:
Anyone, please? I'll be glad to rebase if I get any indication that this will be looked at @duncanbeevers @drwpow

I appreciate that a big project like this is a lot of work, please let me know if there is any way how I can make the review easier for you.

duncanbeevers:
I think this cuts against the grain of the library. In fact, I think this functionality may have once existed in the library, and was subsequently removed.

Rather than tweaking this within the openapi-typescript library, I suggest you look into [OpenAPI Overlays](https://learn.openapis.org/overlay/) which allow you to make modifications to your schema, prior to generating types.

luqasn:
Thanks for the review!

That sounds like a reasonable alternative, I will try going that route.

PR #2536, opened 2025-11-28
fix(openapi-typescript): enumValues generates valid TypeScript for oneOf/anyOf unions

darkbasic:
## Changes

When using `enumValues: true` with OpenAPI schemas containing `oneOf`/`anyOf` unions, the generated TypeScript type paths would fail because properties might only exist on some union variants.

Before (invalid TypeScript):

```ts
export const resourceItemsNestedCodeValues: ReadonlyArray<
  FlattenedDeepRequired<components>["schemas"]["Resource"]["items"]["nested"]["code"]
> = ["a", "b"];
// Error: Property 'nested' does not exist on type '{ type: "simple"; } | { type: "complex"; nested: { code: "a" | "b"; }; }'
```

After (valid TypeScript):

```ts
export const resourceItemsOneOf1NestedCodeValues: ReadonlyArray<
  Extract<Extract<FlattenedDeepRequired<components>["schemas"]["Resource"]["items"], {
    nested: unknown;
  }>["nested"], {
    code: unknown;
  }>["code"]
> = ["a", "b"];
```

The fix wraps property accesses after `oneOf`/`anyOf` indices with `Extract<>` to narrow the union type before accessing variant-specific properties.

This PR has been tested in a production environment with a custom build of openapi-typescript.

## How to Review

I've added three tests:
- Basic union variant narrowing
- Same property name with different inner schemas
- Deeply nested unions (multiple Extract levels)

## Checklist

- [x] Unit tests updated
- [] `docs/` updated (if necessary)
- [] `pnpm run update:examples` run (only applicable for openapi-typescript)

### DIFF
--- a/.changeset/bright-mice-greet.md
+++ b/.changeset/bright-mice-greet.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Fix `enumValues: true` output for schemas that use `oneOf`/`anyOf` unions by narrowing union types with `Extract<>` before accessing variant-specific properties. This prevents invalid type paths when an accessed property only exists on some union members.


darkbasic:
Note: two tests will fail due to network errors (probably the original gist has been deleted). Nothing I can do on that front (they were already failing on main).

sacha-c:
> Note: two tests will fail due to network errors (probably the original gist has been deleted). Nothing I can do on that front (they were already failing on main).

Indeed, I've address this in a separate PR a few weeks back which is awaiting review: https://github.com/openapi-ts/openapi-typescript/pull/2522

darkbasic:
@sacha-c thanks, I've rebased so that tests can pass in CI now

drwpow:
Thank you! Tests look good and this seems like an improvement. Could you please add a changeset (see comment) so we can version & release? Thank you!

darkbasic:
@drwpow done

PR #2549, opened 2025-12-16
feat(openapi-typescript): add readOnly/writeOnly support via markers

### LINKED ISSUE #604: Support "readOnly" and "writeOnly" OpenAPI properties

abumalick:
## Summary

- Add opt-in `--read-write-markers` flag that generates `$Read<T>` and `$Write<T>` marker types for readOnly/writeOnly OpenAPI properties
- Wrap readOnly properties with `$Read<T>`, writeOnly with `$Write<T>`
- Generate inline helper types (`Readable`, `Writable`) in output
- Update openapi-fetch to use `Writable<T>` for request bodies and `Readable<T>` for responses

This enables proper type-checking:
- readOnly properties are excluded from request bodies
- writeOnly properties are excluded from response types

Closes #604

## Test plan

- [x] Added TDD tests in `packages/openapi-fetch/test/read-write-visibility/`
- [x] Tests verify compile-time type errors for invalid usage
- [x] Tests verify runtime behavior
- [x] All existing tests pass

### DIFF
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -99,11 +99,12 @@ export type ParamsOption<T> = T extends {
   : DefaultParamsOption;
 
 // Writable<T> strips $Read markers (readOnly properties excluded from request body)
-export type RequestBodyOption<T> = Writable<OperationRequestBodyContent<T>> extends never
-  ? { body?: never }
-  : IsOperationRequestBodyOptional<T> extends true
-    ? { body?: Writable<OperationRequestBodyContent<T>> }
-    : { body: Writable<OperationRequestBodyContent<T>> };
+export type RequestBodyOption<T> =
+  Writable<OperationRequestBodyContent<T>> extends never
+    ? { body?: never }
+    : IsOperationRequestBodyOptional<T> extends true
+      ? { body?: Writable<OperationRequestBodyContent<T>> }
+      : { body: Writable<OperationRequestBodyContent<T>> };
 
 export type FetchOptions<T> = RequestOptions<T> & Omit<RequestInit, "body" | "headers">;
 
--- a/packages/openapi-typescript-helpers/src/index.ts
+++ b/packages/openapi-typescript-helpers/src/index.ts
@@ -216,15 +216,16 @@ export type $Write<T> = { readonly $write: T };
  * - $Write<T> → never (excluded from response)
  * - object → recursively resolve
  */
-export type Readable<T> = T extends $Write<any>
-  ? never
-  : T extends $Read<infer U>
-    ? Readable<U>
-    : T extends (infer E)[]
-      ? Readable<E>[]
-      : T extends object
-        ? { [K in keyof T as NonNullable<T[K]> extends $Write<any> ? never : K]: Readable<T[K]> }
-        : T;
+export type Readable<T> =
+  T extends $Write<any>
+    ? never
+    : T extends $Read<infer U>
+      ? Readable<U>
+      : T extends (infer E)[]
+        ? Readable<E>[]
+        : T extends object
+          ? { [K in keyof T as NonNullable<T[K]> extends $Write<any> ? never : K]: Readable<T[K]> }
+          : T;
 
 /**
  * Resolve type for writing (requests): strips $Read properties, unwraps $Write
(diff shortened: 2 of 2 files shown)


abumalick:
this one is a new implementation of #1863 @drwpow FYI

abumalick:
mmh, sorry it is not fully ready yet, after more testing I get some bugs

abumalick:
Here it is ready incha Allah.
I tested it by migrating my enterprise's app to use this patched version. It is working well and it is really good to have this painful problem fixed.
I could remove all `@ts-expect-error`s on network requests payloads.

Please tell me if you need any changes or if you have any concern @htunnicliff @drwpow

abumalick:
@htunnicliff @drwpow any feedback please?

abumalick:
**Ping for feedback** @drwpow @kerwanp @gzm0 @htunnicliff

  This implements the `readOnly`/`writeOnly` markers discussed in #1863 following @drwpow's suggested approach.

  **Production validation:** Running in 2 company projects with `openapi-fetch` + `openapi-react-query`. Eliminated nearly all `@ts-expect-error` annotations for request/response type mismatches.

  **Key decisions needing review:**
  - Opt-in via `--read-write-markers` flag
  - `$Read<T>`/`$Write<T>` marker approach
  - Helper types in `openapi-typescript-helpers`

  Happy to adjust based on feedback. Is this direction acceptable or should I take a different approach?

abumalick:
@duncanbeevers or any other maintainer. Can I get a review on this?

drwpow (approved):
Wow, thanks for all the thought and effort put into this PR! This has been a long-requested feature, and this adds in the ability in what looks to be a far more backwards-compatible way than some alternate solutions. I was worried this would be a breaking change, and so this (and the draft PR) fell behind in getting it merged in. But your approach looks sound, and looks like it wouldn’t disrupt users.

For `openapi-fetch` we’ve been reserving minor releases for breaking changes, and even though this is is backwards-compatible, I think a minor change makes sense here given it requires a more recent version of `openapi-typescript`. Folks will be happy to get this merged and released, thank you! 🙂

drwpow:
It looks like the PR has some minor conflicts, if these get resolved I’d be happy to get this merged & released

abumalick:
Thanks @drwpow.

I really love this library and was really bothered by the typing issues with our APIs.
Since I started using the changes from this PR (as a patch) on our projects, we were able to confidently remove all TypeScript silencing and casting related to API types, and it greatly improved our confidence in the backend types. I am really happy with the result.

I was worried that you would not accept this change because of the complexity. I believe it makes a lot of sense to put this behind a flag so as not to bother users who don't need it or don't want to spend time refactoring their code for these changes.

Thank you for taking the time to work on this wonderful library.

PR #2556, opened 2025-12-23
fix(openapi-fetch): omit serializing for already serialized data

justAnArthur:
## Changes

New changes makes omitting serializing for already serialized data, such as string, primary fixing issue with sending `string` as `body`

## How to Review

Take a look at the issue: https://github.com/openapi-ts/openapi-typescript/issues/2555 and on updated code and test. 

## Checklist

- [x] Unit tests updated
-  ~~`docs/` updated (if necessary)~~
- ~~`pnpm run update:examples` run (only applicable for openapi-typescript)~~

### DIFF
--- a/packages/openapi-fetch/test/common/request.test.ts
+++ b/packages/openapi-fetch/test/common/request.test.ts
@@ -213,30 +213,32 @@ describe("request", () => {
     });
 
     test.each(BODY_ACCEPTING_METHODS)("`''` body (with body serializer) - %s", async (method) => {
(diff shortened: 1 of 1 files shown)


drwpow:
Apologies, I am onboard with this change but the tests were actually failing. I had thought the CI failures were unrelated, caused by something else in CI was failing. But you actually introduced breaking behavior in your change.

Please feel free to take another shot at it, off of latest `main`. The tests should pass this time 🙂

PR #2572, opened 2026-01-28
Do not treat Content-Length=0 as empty when Transfer-Encoding is chunked

luis-guideti:
Some HTTP servers incorrectly send both:

Content-Length: 0
Transfer-Encoding: chunked

Per RFC 7230, Transfer-Encoding: chunked takes precedence and the body may still contain data. Our previous logic treated any Content-Length=0 response as empty, causing valid responses from misconfigured servers to be incorrectly ignored.

This change updates the empty-body detection logic to only treat Content-Length=0 as empty when the response is not chunked.

## Changes

## How to Review

## Checklist

- [x] Unit tests updated


luis-guideti:
This also improves compatibility with CapacitorJS — the @capacitor/http plugin can send both Content-Length: 0 and Transfer-Encoding: chunked, and with this change those responses work as expected.

drwpow:
This looks great, thank you! Can you add a changeset (see comment)?

PR #2575, opened 2026-02-03
feat(openapi-fetch): only parse as JSON if content-type indicates JSON

### LINKED ISSUE #2574: Fails to parse responses when `Content-Length` header is stripped by proxy
### openapi-fetch version

0.13.8

### Description

When a response with an empty body passes through Cloudflare (or other proxies), the client attempts to parse the empty response as JSON and fails. This happens because the current logic incorrectly treats the absence of a `Content-Length` header as indicating content is present.

### Reproduction

```javascript
// Mock server that returns empty body without Content-Length header
import express from 'express';

const app = express();

app.post('/test', (req, res) => {
  res.removeHeader('Content-Length');
  res.status(200).end();
});

app.listen(3000);
```

Or simulate with fetch mock:
```javascript
global.fetch = async () => new Response(null, {
  status: 200,
  headers: {} // No Content-Length header
});

// Now use openapi-fetch - it will fail trying to parse empty body as JSON
```

### Expected result

The client should gracefully handle empty responses regardless of whether the `Content-Length` header is present or absent. When a response has no body, `response.data` should be `undefined` or an appropriate empty value without attempting to parse it as JSON.

### Extra

brunolca:
Fixes #2574 

## Changes

- For success responses with parseAs: "json" (the default), check if Content-Type includes application/json before calling response.json(). Falls back to response.text() otherwise.                                                                                                                 
- For error responses, only attempt JSON.parse() if the Content-Type includes application/json.
  
## How to Review

As I understand this it is a breaking change. You can test by follwing the reproduction steps here #2574

## Checklist

- [x] Unit tests updated

### DIFF
--- a/packages/openapi-fetch/test/common/request.test.ts
+++ b/packages/openapi-fetch/test/common/request.test.ts
@@ -340,8 +340,9 @@ describe("request", () => {
     function createCustomFetch(data: any) {
       const response = {
         clone: () => ({ ...response }),
-        headers: new Headers(),
+        headers: new Headers({ "content-type": "application/json" }),
         json: async () => data,
+        text: async () => JSON.stringify(data),
         status: 200,
         ok: true,
       } as Response;
@@ -353,7 +354,7 @@ describe("request", () => {
 
     const customFetch = createCustomFetch({});
     const client = createClient<paths>({ baseUrl: "https://fakeurl.example", fetch: customFetch });
-    client.GET("/resources", {
+    await client.GET("/resources", {
       customProperty: "value",
     });
   });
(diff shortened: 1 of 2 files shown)


drwpow:
Thanks for raising! We’ll need a few things for this PR:

1. A changeset
2. All CI passing (including lint)

There’s also a conflict with another PR that merged before yours, which should be fixable. Once all are in place, happy to review! 🙂

brunolca:
The conflict was due to #2605 resolving my issue. I am closing the PR. 

Thanks for the work!

PR #2605, opened 2026-02-10
fix(openapi-fetch): handle empty error responses when Content-Length header is stripped

### LINKED ISSUE #2574: Fails to parse responses when `Content-Length` header is stripped by proxy

kaigritun:
## Summary

Fixes #2574

When a proxy (like Cloudflare) strips the `Content-Length` header from an error response with no body, the client would return an empty string as the error value. Since empty string is falsy, this causes issues when checking `if (error)` - the code incorrectly treats it as "no error".

## Changes

This fix applies the same safe handling to error responses that already exists for success responses. When `Content-Length` is absent and the body is empty:
- Use `response.text()` first to get the raw content
- If the raw content is empty, return `undefined` for the error value
- This is consistent with the existing handling for status 204 and `Content-Length: 0`

## Testing

Added two new test cases in `delete.test.ts`:
1. Error response (500) with empty body and no Content-Length header
2. Success response (200) with empty body and no Content-Length header (verifies existing behavior still works)

All 442 tests pass including the new ones.

## Reproduction

```javascript
// Simulate proxy stripping Content-Length header
global.fetch = async () => new Response(null, {
  status: 500,
  headers: {} // No Content-Length header
});

// Before fix: { error: "", response } - empty string is falsy
// After fix: { error: undefined, response } - consistent with 204 handling
```

### DIFF
--- a/.changeset/fix-empty-error-response.md
+++ b/.changeset/fix-empty-error-response.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
+fix(openapi-fetch): handle empty error responses when Content-Length header is stripped


packages/openapi-fetch/test/http-methods/delete.test.ts
```diff
@@ -46,4 +46,50 @@ describe("DELETE", () => {
...
+
+    // assert error is undefined for empty body (consistent with 204 and Content-Length: 0 handling)
+    expect(error).toBeUndefined();
+
+    // assert response status is preserved
+    expect(response.status).toBe(500);
+    expect(response.ok).toBe(false);
+  });
+
+  test("handles success response with empty body when Content-Length header is stripped by proxy", async () => {
```
drwpow: Great tests! 🎉

MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

drwpow-figma:
> I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

@MichaelDeBoey thanks for the heads up! We’re aware of the rise of agents making PRs. We’ll continue monitoring submissions just like any other. We don’t weigh repeat submissions any different, since it is a known tactic to land 3-4 innocuous PRs before introducing a backdoor etc.

As of now we don’t have any policies against agents submitting PRs but we’ll monitor submissions and react accordingly.

Also for you or anyone else reading we really appreciate flagging PRs like this. Please continue to call our attention to these matters!

duncanbeevers:
I suppose this was inevitable. These kinds of agents using "human-like" avatars is unnecessarily misleading.

@kaigritun Please update your avatar and profile to better-indicate your provenance before opening further PRs against this repo.

PR #2609, opened 2026-02-10
config(renovate): automerge minor & patch dev dependencies

gzm0:
## Changes

Do not create PRs for minor & patch dev dependency updates if CI passes.


drwpow:
Works for me! I think we talked about it but I never looked into the change. I’m fine with auto-bumping deps

PR #2637, opened 2026-02-20
use typeof for cross-realm function detection

erietz:
Replace `instanceof Function` with `typeof === 'function'` in `defaultBodySerializer` to fix form-urlencoded body serialization in environments with separate JavaScript realms (e.g., Jest with experimental VM modules).

The instanceof check fails across realms because the Function constructor differs between realms, causing headers.get to not be recognized as a function even though it is one.

## Changes

Related issue here https://github.com/openapi-ts/openapi-typescript/pull/1762

## How to Review

I've created a minimal example in this repo https://github.com/erietz/openapi-fetch-bug

## Checklist

- [x] Unit tests updated

### DIFF
--- a/.changeset/khaki-clowns-fly.md
+++ b/.changeset/khaki-clowns-fly.md
@@ -1,0 +1,5 @@
+---
+"openapi-fetch": patch
+---
+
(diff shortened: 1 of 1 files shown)


drwpow:
Thank you! I’m fine with this change, I don’t think there was an intentional reason for `instanceof`.

If you could submit a changeset (see comment), and fix the lint errors, would be happy to approve & merge 🙂

erietz:
Good morning @drwpow, I've got the lint errors fixed up and a changeset created.

PR #2650, opened 2026-02-27
fix: make Readable/Writable marker resolution opt-in, fix readonly arrays

### LINKED ISSUE #2615: `Readable` in new openapi-typescript-helpers is broken
### openapi-fetch version

0.17.0

### Description

it is broken using both the latest openapi-typescript package and using the previous version without re-generating the types.

I always use `--immutable=true` and almost all JsonSchemas in my openapi have:

```yaml
      additionalProperties: false
      readOnly: true
```

In the latest 0.17.0 version, Arrays in the response bodies of openapi-fetch are iterable objects that remove methods for reading arrays, like `.map` or `.toString`.

```
 '{ readonly [x: number]: string; readonly length: number; toString: {}; toLocaleString: {}; concat: {}; join: {}; slice: {}; indexOf: {}; lastIndexOf: {}; every: {}; some: {}; forEach: {}; ... 20 more ...; readonly [Symbol.unscopables]: { ...; }; }' 
```

I can workaround this and get the old behavior before version 0.17.0, by putting this iterable object into `Array.from`, then typing it as readonly:
```typescript
const openapiFetchArray = response.data.myArray; /* `Readable` object from openapi-typescript-helpers */
const desiredArray: readonly Item[] = Array.from(openapiFetchArray);
````

### Reproduction

this is reproduced on type-checking a project using openapi-fetch after updating to 0.17.0

### Expected result

I think this should be opt-in. I do not have `--read-write-markers` anywhere, or made any other changes to the code.

### Extra

### LINKED ISSUE #2620: `Readable` prevents type narrowing
### openapi-fetch version

0.17.0

### Description

After updating to `openapi-fetch@0.17.0` and `openapi-typescript@7.13.0` I get a type error when accessing a deeply nested property in combination with type narrowing. The original API model is rather complex and in parts driven by code generation requirements from the Java-based backend. However, it is a valid OpenAPI spec and everything worked fine until the update. I think it's the easiest if you take a look at the reproduction linked below.

I could narrow the issue down to the introduction of the `Readable` type. When I replace the `Readable` type in `openapi-typescript-helpers` with `type Readable<T> = T` everything works fine.

### Reproduction

The original OpenAPI spec is from a complex insurance application. I tried to boil it down to the absolute minimum. Without the context it may seem like a rather contrived example, but you have to trust me that it's a valid use case 😉

https://github.com/luchsamapparat/openapi-typescript-readable-bug

### Expected result

It should be possible to access `tinyFooProp` when narrowing the model down via `['@type'] === 'Foo'`.

### Extra

### PR DESCRIPTION
**Pull request by worjs**
Fixes #2615
Fixes #2620

## Motivation

PR #2549 (`openapi-fetch@0.17.0`) introduced `$Read<T>` / `$Write<T>` markers and applied `Readable<T>` unconditionally to all `FetchResponse` data. This caused two regressions for users who don't use `--read-write-markers`:

1. **#2615** — `Readable<readonly string[]>` falls through to the `T extends object` branch because `readonly T[]` doesn't extend `T[]`. The array gets destructured into `{ readonly [x: number]: string; length: number; ... }`.

2. **#2620** — The mapped type inside `Readable<T>` creates new structural types that break TypeScript's discriminated union narrowing, even when no `$Read`/`$Write` markers are present.

As @drwpow noted in #2615: "Given that this was less backwards-compatible than anticipated, we'll add a flag to make this change opt-in."

## Approach

Two-part fix:

**1. Fix `Readable<T>` / `Writable<T>` readonly array handling**

Added a mutable array branch before the readonly fallback. Order matters because `readonly T[]` doesn't match `(infer E)[]` but `T[]` *does* match `readonly (infer E)[]` — so mutable must come first:

```typescript
: T extends (infer E)[]            // mutable arrays
  ? Readable<E>[]
  : T extends readonly (infer E)[] // readonly arrays (fallback)
    ? readonly Readable<E>[]
```

**2. Make marker resolution opt-in via `Markers` type parameter**

Instead of always wrapping responses in `Readable<T>`, added `MaybeReadable<T, Markers>` / `MaybeWritable<T, Markers>` that act as identity when `Markers = false` (default):

```typescript
type MaybeReadable<T, Markers extends boolean> = Markers extends true ? Readable<T> : T;
```

Threaded `Markers extends boolean = false` through the entire openapi-fetch type chain (`FetchResponse`, `RequestBodyOption`, `ClientMethod`, `Client`, `createClient`, etc). Default is `false` so existing code is unaffected.

```typescript
// Default: no marker resolution (backwards compatible)
const client = createClient<paths>();

// Opt-in for --read-write-markers users
const client = createClient<paths, MediaType, true>();
```

## Breaking Change

Users of `--read-write-markers` need to pass `true` as the third type parameter. The `MediaType` positional argument is required because TypeScript doesn't support skipping positional type parameters. This is documented in the changeset.

## How to Review

- **Core type fix**: `openapi-typescript-helpers/src/index.ts` — `Readable`, `Writable`, `MaybeReadable`, `MaybeWritable` (~40 lines)
- **Type threading**: `openapi-fetch/src/index.d.ts` — mechanical `Markers` parameter addition through the chain
- **Inline helpers**: `openapi-typescript/src/transform/index.ts` — same readonly fix in the codegen string
- **Tests**: `openapi-fetch/test/read-write-visibility/` — regression tests for both issues + `MethodResponse` with markers

## Tests

- `openapi-fetch`: 460 passed, 0 type errors
- `openapi-react-query`: 37 passed (no changes needed, `Markers` defaults to `false`)
- `openapi-typescript`: 283 passed

## Checklist

- [x] Regression test for #2615 (readonly array preservation)
- [x] Regression test for #2620 (discriminated union narrowing)
- [x] `MethodResponse` type inference with `Markers`
- [x] `Markers=true` + `$Read` markers + discriminated union narrowing verified
- [x] `openapi-react-query` compatible without changes
- [x] Changeset with migration guide

### DIFF
--- a/.changeset/fix-readable-markers.md
+++ b/.changeset/fix-readable-markers.md
@@ -1,0 +1,25 @@
+---
+"openapi-typescript-helpers": patch
+"openapi-fetch": minor
+"openapi-typescript": patch
+---
+
+Fix `Readable` and `Writable` types to correctly handle readonly arrays, and make marker resolution opt-in via `Markers` type parameter (default: `false`). This fixes issues where `Readable<readonly string[]>` would destructure the array into an object type (#2615), and where applying `Readable`/`Writable` unconditionally broke discriminated union narrowing (#2620).
+
+### Breaking change for `--read-write-markers` users
+
+`Readable`/`Writable` are no longer applied unconditionally to all responses and request bodies. If you use `--read-write-markers` with `openapi-fetch`, you now need to pass `true` as the third type parameter to opt in to marker resolution:
+
+```ts
+// Before (0.17.x): markers resolved automatically
+const client = createClient<paths>();
+
+// After: opt in explicitly
+import type { MediaType } from "openapi-typescript-helpers";
+//                          ↓ MediaType is required as a positional placeholder
+const client = createClient<paths, MediaType, true>();
+```
+
+The second type parameter (`MediaType`) is the media type filter — it defaults to `MediaType` which accepts all content types. You must specify it explicitly to reach the third `Markers` parameter, since TypeScript does not support skipping positional type arguments.
+
+Without `Markers = true`, `$Read`/`$Write` markers will be treated as raw wrapper types and not resolved. This is the intended default for users who do not use `--read-write-markers`.
--- a/packages/openapi-fetch/src/index.d.ts
+++ b/packages/openapi-fetch/src/index.d.ts
@@ -3,14 +3,14 @@ import type {
   FilterKeys,
   HttpMethod,
   IsOperationRequestBodyOptional,
+  MaybeReadable,
+  MaybeWritable,
   MediaType,
   OperationRequestBodyContent,
   PathsWithMethod,
-  Readable,
   RequiredKeysOf,
   ResponseObjectMap,
   SuccessResponse,
-  Writable,
 } from "openapi-typescript-helpers";
 
 /** Options for each client instance */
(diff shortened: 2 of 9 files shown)


drwpow:
Thanks so much for submitting this! I just merged #2646, which allows that one manually-generated test fixture to be regenerated, and will fail without that flag.

If you would be willing, I‘d love to rebase this PR on top of `main`, and maybe we can add another fixture that’s the same `read-write-visibility` fixture, but without the `--read-write-markers` flag, so we have 1 test with it off and 1 with it on. As long as we have both fixtures added to `packages/openapi-fetch/test/redocly.yaml` and regenerating each time, we can have assurance that changes in `openapi-typescript` doesn’t affect `openapi-fetch` downstream related to this.

Thanks!

worjs:
Thanks for the review @drwpow! Rebased on main and added a flag-off fixture. Can you take another look?

worjs:
Rebased on latest main, all tests passing. Mind taking another look @drwpow?

PR #2774, opened 2026-04-15
feat: add TypeScript 6 support

benjamineckstein:
## Summary

Add full TypeScript 6 support — not just the peer dep range, but also a fix for a TS6 behavioral change that breaks `Readable<T>` and `Writable<T>`.

### Changes

**Peer dep & devDependency (commit 1):**
- `packages/openapi-typescript/package.json`: widen peer dep `^5.x` → `^5.x || ^6.x`
- `packages/openapi-typescript-helpers/package.json`: upgrade devDependency `5.9.3` → `6.0.2`

**Fix Readable/Writable types (commit 2):**
In TypeScript 6, `Date extends object` evaluates to `true` (changed from TS5). This caused `Readable<T>` and `Writable<T>` to structurally expand built-in objects like `Date` into `{ toString: {}; toDateString: {}; ... }` instead of preserving the `Date` type.

Added a guard clause for `Date | RegExp | ((...args: never[]) => unknown)` before the `T extends object` branch in both type utilities, preserving built-in objects. This fix is backward compatible with TS5.

**Changeset:** patch release for both `openapi-typescript` and `openapi-typescript-helpers`

### Validation
- ✅ `pnpm run build` — all 4 packages build successfully
- ✅ `pnpm test` — all 7 test suites pass
- ✅ `pnpm run lint` — clean

### Context
- Closes #2723
- Supersedes #2711 (Renovate PR, missing changeset and the Readable/Writable fix)
- Fix for `Readable<T>`/`Writable<T>` reported by @yamafaktory in #2723

### DIFF
--- a/packages/openapi-typescript-helpers/src/index.ts
+++ b/packages/openapi-typescript-helpers/src/index.ts
@@ -223,9 +223,11 @@ export type Readable<T> =
       ? Readable<U>
       : T extends (infer E)[]
         ? Readable<E>[]
-        : T extends object
-          ? { [K in keyof T as NonNullable<T[K]> extends $Write<any> ? never : K]: Readable<T[K]> }
-          : T;
+        : T extends Date | RegExp | ((...args: never[]) => unknown)
+          ? T
+          : T extends object
+            ? { [K in keyof T as NonNullable<T[K]> extends $Write<any> ? never : K]: Readable<T[K]> }
+            : T;
 
 /**
  * Resolve type for writing (requests): strips $Read properties, unwraps $Write
@@ -240,8 +242,10 @@ export type Writable<T> =
       ? Writable<U>
       : T extends (infer E)[]
         ? Writable<E>[]
-        : T extends object
-          ? { [K in keyof T as NonNullable<T[K]> extends $Read<any> ? never : K]: Writable<T[K]> } & {
-              [K in keyof T as NonNullable<T[K]> extends $Read<any> ? K : never]?: never;
-            }
-          : T;
+        : T extends Date | RegExp | ((...args: never[]) => unknown)
+          ? T
+          : T extends object
+            ? { [K in keyof T as NonNullable<T[K]> extends $Read<any> ? never : K]: Writable<T[K]> } & {
+                [K in keyof T as NonNullable<T[K]> extends $Read<any> ? K : never]?: never;
+              }
+            : T;
--- a/.changeset/widen-typescript-peer-dep.md
+++ b/.changeset/widen-typescript-peer-dep.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+feat(openapi-typescript): add TypeScript 6 support
--- a/.changeset/widen-typescript-peer-dep.md
+++ b/.changeset/widen-typescript-peer-dep.md
@@ -1,5 +1,9 @@
 ---
 "openapi-typescript": patch
+"openapi-typescript-helpers": patch
 ---
 
-feat(openapi-typescript): add TypeScript 6 support
+feat: add TypeScript 6 support
+
+- Widen peer dep to `^5.x || ^6.x`
+- Fix `Readable<T>` and `Writable<T>` to preserve built-in objects (Date, RegExp, functions) that now match `extends object` in TS6
--- a/packages/openapi-fetch/test/helpers.ts
+++ b/packages/openapi-fetch/test/helpers.ts
@@ -24,18 +24,18 @@ export function createObservedClient<T extends {}, M extends MediaType = MediaTy
 }
 
 /**
- * Convert a Headers object to a plain object for easier comparison
+ * Convert a Headers object to a plain object for easier comparison.
+ * Uses Headers.forEach rather than Headers.entries because forEach is in base
+ * lib.dom with an identical signature in both TS 5.x and 6.x, while entries
+ * requires DOM.Iterable in TS 5.x.
  */
 export function headersToObj(headers: Headers | Record<string, string>): Record<string, string> {
-  const iter =
-    headers instanceof Headers
-      ? headers
-          // @ts-expect-error FIXME: this is a missing "lib" in tsconfig.json but dunno what
-          .entries()
-      : Object.entries(headers);
-  const result: Record<string, string> = {};
-  for (const [k, v] of iter) {
-    result[k] = v;
+  if (!(headers instanceof Headers)) {
+    return { ...headers };
   }
+  const result: Record<string, string> = {};
+  headers.forEach((value, key) => {
+    result[key] = value;
+  });
   return result;
 }
--- a/packages/openapi-typescript-helpers/package.json
+++ b/packages/openapi-typescript-helpers/package.json
@@ -34,6 +34,6 @@
     "lint:ts": "tsc --noEmit"
   },
   "devDependencies": {
-    "typescript": "5.9.3"
+    "typescript": "6.0.2"
   }
 }
--- a/packages/openapi-typescript/package.json
+++ b/packages/openapi-typescript/package.json
@@ -59,7 +59,7 @@
     "version": "pnpm run build"
   },
   "peerDependencies": {
-    "typescript": "^5.x"
+    "typescript": "^5.x || ^6.x"
   },
   "dependencies": {
     "@redocly/openapi-core": "^1.34.6",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -67,7 +67,7 @@ importers:
         version: 7.3.1(@types/node@25.6.0)(jiti@2.6.1)(yaml@2.8.2)
       vitepress:
         specifier: 1.6.4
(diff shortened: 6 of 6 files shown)


packages/openapi-typescript-helpers/src/index.ts
```diff
@@ -223,9 +223,11 @@ export type Readable<T> =
       ? Readable<U>
       : T extends (infer E)[]
         ? Readable<E>[]
-        : T extends object
-          ? { [K in keyof T as NonNullable<T[K]> extends $Write<any> ? never : K]: Readable<T[K]> }
-          : T;
+        : T extends Date | RegExp | ((...args: never[]) => unknown)
```
benjamineckstein, line 226: In TypeScript 6, `Date extends object` now evaluates to `true` (it was `false` in TS5). Without this guard, `Readable<T>` would structurally expand `Date` fields into `{ toString: {}; toDateString: {}; ... }` instead of preserving the `Date` type — making any schema containing Date fields incompatible with itself when accessed through `FetchResponse`.

This guard preserves built-in objects (Date, RegExp, functions) by matching them before the generic `T extends object` branch. It's backward compatible with TS5 since these types were never matched by `T extends object` there.

benjamineckstein:
What is missing to merge this PR? ... now it contains merge conflicts ...

RaffaeleCanale:
Perhaps @drwpow isn't available at the moment, might be worth a try to request a review from another maintainer? Maintainers are listed here: https://github.com/openapi-ts/openapi-typescript/discussions/2559

ingalls:
@htunnicliff, @gr2m, or @duncanbeevers Would anyone be able to review & merge this?

ingalls:
I've sent @duncanbeevers an email, I'm happy to sponsor the time/work of a maintainer to review and merge this for $250 USD payable via Github Sponsors

maxwellsmart84:
would really like to see this get merged as we have had a Dependabot PR open for 2 months from this.

benjamineckstein:
I have solved the merge conflicts ... ready to be merged again ... you are welcome!

duncanbeevers:
@benjamineckstein I've looked over this PR, and though its quite small by itself, a major TypeScript version change can be a big move.

I've forked your PR and contributed what I think are important pieces to nail down before landing this.
https://github.com/duncanbeevers/openapi-typescript/tree/feat/typescript-6-support

Importantly, I think the changes to `Readable` and `Writable` can be simplified.

The changes include the introduction of a small test suite validating the `Readable` and `Writable` behaviors, and testing the types across multiple TypeScript versions.

Please take a look at these and consider applying some, or all of the changes to this PR.

marcomuser:
> @benjamineckstein I've looked over this PR, and though its quite small by itself, a major TypeScript version change can be a big move.
> 
> I've forked your PR and contributed what I think are important pieces to nail down before landing this. https://github.com/duncanbeevers/openapi-typescript/tree/feat/typescript-6-support
> 
> Importantly, I think the changes to `Readable` and `Writable` can be simplified.
> 
> The changes include the introduction of a small test suite validating the `Readable` and `Writable` behaviors, and testing the types across multiple TypeScript versions.
> 
> Please take a look at these and consider applying some, or all of the changes to this PR.

Why did you pin openapi-fetch to v5.9?

sosoba:
related to https://github.com/openapi-ts/openapi-typescript/issues/2550

duncanbeevers:
> Why did you pin openapi-fetch to v5.9?

Since I don't directly manage the openapi-fetch releases, and because there were a couple of changes that affected that repo differently than the rest of the packages, I thought to leave it alone. However, I recognize that this puts downstream consumers in a bad position, and I'll go ahead and include `openapi-fetch` in an across-the-board upgrade to TS 6 compatibility.

LMBernardo:
Thanks for this! I am currently blocked from upgrading to TS6 on a personal project due to this package, so definitely want this update.

One small thought on `headersToObj` - we could use `Headers.forEach` to avoid the `as unknown as Iterable<…>` double-cast:

```ts
export function headersToObj(headers: Headers | Record<string, string>): Record<string, string> {
  const result: Record<string, string> = {};
  if (headers instanceof Headers) {
    headers.forEach((value, key) => { result[key] = value; });
  } else {
    for (const [key, value] of Object.entries(headers)) result[key] = value;
  }
  return result;
}
```

A few reasons why it might be nicer:

- **No `as unknown` assertion** — `value`/`key` are checked against the real `Headers.forEach` signature, whereas a double-cast compiles even if the iterable shape is wrong.
- **Independent of `DOM.Iterable`** — `forEach` is in the base `lib.dom.d.ts` with an identical signature in both TS 5.x and 6.x ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Headers/forEach)), so it avoids the lib gap this is working around (`entries()` only moved into base `lib.dom` in TS 6).
- Minor: skips the intermediate `Array.from(...)` allocation.

Verified it type-checks clean under both `typescript@5.9.3` and `6.0.3`.

Either way, the work here is much appreciated!

benjamineckstein:
Thanks @LMBernardo, great suggestion! Applied in 6a805ba5293b8f9ce390a3546d82ad9ec5996743.

I went with your `Headers.forEach` approach and simplified the plain-object branch one step further: since the input is already a `Record<string, string>` in that branch, a spread is enough.

```ts
export function headersToObj(headers: Headers | Record<string, string>): Record<string, string> {
  if (!(headers instanceof Headers)) {
    return { ...headers };
  }
  const result: Record<string, string> = {};
  headers.forEach((value, key) => {
    result[key] = value;
  });
  return result;
}
```

This also drops the old `@ts-expect-error FIXME` workaround on `.entries()`, which was exactly the lib gap you described (`entries()` only landed in base `lib.dom` with TS 6). Verified clean under both `typescript@5.9.3` and `6.0.2`, and all 451 openapi-fetch tests pass.

Side note: since this PR has been open for a while, I started a small OpenAPI generator of my own that emits Zod schemas natively and targets TS 6+ out of the box: https://github.com/codewithagents/openapi-zod-ts. That said, I will happily keep this PR up to date, native TS 6 support here would help a lot of users.

shish:
FWIW typescript 7 is out now :eyes:

https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/

My understanding is that ts6 was designed to be a bridge, where ts5/ts6 share code, and ts6/ts7 share an API, so I hope this PR will work with 7 just by widening the peer dep / dev dependency? :heart:

sosoba:
> 
> My understanding is that ts6 was designed to be a bridge, where ts5/ts6 share code, and ts6/ts7 share an API 

TS 7 does not offer API yet. CLI only. Only TS7.1 will provide the new API.

shish:
> TS 7 does not offer API yet. CLI only. Only TS7.1 will provide the new API.

That's for the internal compiler API - I'm referring to the external interface (ie what code the CLI can parse and what output it generates - my understanding is that there should be zero behaviour changes between 6 and 7, so nothing like the "Date wasn't an object in the past, and now it is" issue that was discovered in the 5-to-6 change)

FWIW I have tested by taking this PR and adding

* `packages/openapi-typescript/package.json: widen peer dep ^5.x || ^6.x -> ^5.x || ^6.x || ^7.x`

on top, and it appears to work for my (admittedly very simple) use-case :eyes:

saidyeter:
looking forward to see this mr merged

darkbasic:
I did some further improvements: https://github.com/darkbasic/openapi-typescript/commit/219be04e522ab44c0fcd7b8b48cfbd3be237e8f1

- **Workspace-wide TypeScript 6:** use the shared TypeScript 6.0.3 catalog instead of updating only the helpers package to 6.0.2. Add narrowly scoped peer exceptions for the existing declaration-build tools.
- **Consistent generated helpers:** apply the fixes to both the helpers package and the types emitted by `--read-write-markers`; the original PR changed only the package implementation.
- **Safer built-in handling:** preserve callable methods without bypassing visibility markers on extra `Date`/`RegExp` properties. Document that callable arguments, results, and attached properties remain opaque. Correct the release note: the mapped-type bug also existed under TS5, not because `Date extends object` changed in TS6.
- **Readonly collections and generics:** resolve markers in array methods and iterators while preserving readonly indices, length, required numeric properties, and extra data. Cover constrained generics and recursive arrays/tuples without tuple-reconstruction or depth-limit machinery. Retain existing mutable-tuple behavior; excluded readonly elements now follow the mutable-array `never` rule.
- **Generated-name collisions:** inline the generated readonly-data mapping so it cannot conflict with a schema root type named `ReadonlyArrayData`. Keep the shared private alias in the package and an `any` fast path for compiler efficiency.
- **Regression coverage and packaging:** add helper assertions and actual generated-output compilation tests, including negative checks that reject accidental widening to `any`. Test both immutable settings and exclude test sources from published packages.
- **Compiler configuration and CI:** remove deprecated options, fix the nonstrict test's `rootDir`, test TS5/TS6 on Node 22/24, check framework examples, avoid duplicate lint/type-test execution, and pin browser testing to Ubuntu 24.04 / Node 24 with timeouts.

I'm still unsure if it's worth opening a PR until a maintainer shows up, for the moment I'm maintaining my own fork.

PR #2777, opened 2026-04-17
feat(openapi-react-query): add queryKeyFn option to createClient

codr:
## Problem

When an app uses two `openapi-fetch` clients targeting different servers (e.g. an app API at `:8000` and a PDM API at `:8100`), any endpoint path shared between them produces identical TanStack Query keys `[method, path, init]`. TanStack Query fires only one request and both hooks receive the same cached data — even though they point at different servers.

## Fix

Add an optional `queryKeyFn` to `createClient`. When provided, it replaces the default `[method, path, init]` key, giving callers a way to namespace keys per client instance:

```typescript
const pdmClient = createClient(fetchClient, {
  queryKeyFn: (method, path, init) => ["pdm", method, path, init],
});
```

This is purely additive — default behaviour is unchanged.

### Implementation note

The existing shared `queryFn` destructured `[method, path, init]` from the TanStack-provided `queryKey` context. Allowing custom key shapes would have silently broken that destructuring. This PR refactors `queryFn` into a per-call closure that captures `method`, `path`, and `init` from the outer `queryOptions` call instead. The same fix is applied to the `useInfiniteQuery` inline queryFn.

## Changes

- `src/index.ts` — exports `QueryKeyFn` and `CreateClientOptions` types; threads `queryKeyFn` through `createClient`; inlines the queryFn as a closure
- `test/index.test.tsx` — new `describe("queryKeyFn")` block: default key shape, custom key shape, and no-dedup-across-clients
- `README.md` — new Options section documenting `queryKeyFn` with two-client example
- `.changeset/query-key-fn.md` — minor bump changeset

## Breaking changes

None — purely additive; passing no `clientOptions` gives identical behaviour to before.

### DIFF
--- a/.changeset/query-key-fn.md
+++ b/.changeset/query-key-fn.md
@@ -1,0 +1,10 @@
+---
+"openapi-react-query": minor
+---
+
+Add `queryKeyFn` option to `createClient` for per-client query key namespacing.
+
+When two `openapi-fetch` clients target different servers but share endpoint paths,
+TanStack Query deduplicates their requests because the default `[method, path, init]`
+key is identical. Passing a `queryKeyFn` lets callers namespace keys per client
+instance so each client maintains an independent cache.
--- a/packages/openapi-react-query/README.md
+++ b/packages/openapi-react-query/README.md
@@ -60,6 +60,37 @@ const MyComponent = () => {
 
 > You can find more information about `createFetchClient` in the [openapi-fetch documentation](../openapi-fetch).
 
+## Options
+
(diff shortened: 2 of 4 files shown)


codr:
Friendly ping.

Is there anything I can do to help this move forward?

codr:
Friendly ping. @duncanbeevers is there anything I can do to help land this?

codr:
Closing this because of inactivity.
Happy to re-open this later, if we find it's useful.

PR #2813, opened 2026-05-17
fix(openapi-react-query): propagate undefined errors from openapi-fetch

### LINKED ISSUE #2070: openapi-react-query does not propagate `undefined` errors from openapi-fetch

Will-hxw:
## Summary
- Use `!response.ok` instead of `error` truthiness check in queryFn, useInfiniteQuery, and useMutation
- Properly handle cases where openapi-fetch returns `error: undefined` for non-OK responses with empty bodies

## Why
When openapi-fetch receives a non-OK response (e.g., 500) with an empty body, it returns `{ error: undefined, response }`. The previous `if (error)` check was falsy for `undefined`, causing the error to be silently swallowed. This fix checks `!response.ok` instead, ensuring errors are always propagated.

## Validation
- Ran `pnpm test --filter openapi-react-query` — all 38 tests pass (37 existing + 1 new)
- Added test case for 500 response with empty body

## Related
Closes #2070

### DIFF
--- a/.changeset/fix-propagate-undefined-errors.md
+++ b/.changeset/fix-propagate-undefined-errors.md
@@ -1,0 +1,7 @@
+---
+"openapi-react-query": patch
+---
+
+fix(openapi-react-query): propagate undefined errors from openapi-fetch
+
+Use `!response.ok` instead of `error` truthiness check to properly handle cases where openapi-fetch returns `error: undefined` for non-OK responses with empty bodies.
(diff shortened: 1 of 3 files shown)


Will-hxw:
Hi!@duncanbeevers,sorry to bother you. This PR has been open for a while, so I wanted to kindly check whether you might have time to review it or let me know if any changes are needed.I’m happy to make adjustments if required. Thanks!

PR #1098, opened 2023-05-08
Add openapi-fetch

drwpow:
## Changes

Pulls the [openapi-fetch](https://github.com/drwpow/openapi-fetch) package into this repo for easier management, and because they are spiritually part of the same project.

This doesn’t complete the swapover; there’s still some cleanup and migrating of issues. But it moves the bulk of the code over.

PR #2819, opened 2026-05-26
Support TypeScript 6.0 (completes #2711)

### LINKED ISSUE #2550: Not TS 6 compliant
https://github.com/openapi-ts/openapi-typescript/blob/e9d3c687972439650202f94f48cffc431d428080/packages/openapi-typescript/package.json#L61-L63

```shell
npm i -D openapi-typescript typescript@next
```

```
npm warn Could not resolve dependency:
npm warn peer typescript@"^5.x" from openapi-typescript@7.10.1
```

LMBernardo:
> [!NOTE]
> **This PR was prepared with AI assistance.** Flagging that up front for transparency. Every technical claim below is backed by the linked TypeScript 6.0 release notes or has notes on how to reproduce the checks (commands under "How to Review"), and the result is verified locally on the pinned toolchain.

## Changes

This makes `openapi-typescript`'s CI pass under TypeScript 6.0, completing the dependency bump in #2711.

#2711 (opened by Renovate) bumps the dev `typescript` to `6.0.3` and widens the `openapi-typescript` peer range to `^5.x || ^6.0.0`, but its `lint` check fails, so it can't merge. This branch contains that same bump plus the three minimal changes needed to make `lint` green. All other checks (`test-*`, `size-limit`, CodeQL) already pass on #2711.

The `lint` task runs `tsc`, and the first failure is a compiler *options* error that aborts type-checking before it starts — so each fix uncovers the next. Under `typescript@6.0.3`:

1. **`packages/openapi-fetch/tsconfig.json` — remove `downlevelIteration`.** `tsc` raises `TS5101`. Per the [TS 6.0 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#deprecated---downleveliteration): *"using `--downlevelIteration false` with `--target es2015` did not error in TypeScript 5.9 and earlier, even though it had no effect. In TypeScript 6.0, setting `--downlevelIteration` at all will lead to a deprecation error."* The option only affects ES5 emit; this package targets `ESNext`, so it was inert. It was introduced in the package's first commit (#1098) alongside `target: ESNext` and never had any effect.

2. **`packages/openapi-fetch/test/helpers.ts` — use `Headers.forEach` instead of `Headers.entries()`.** With (1) fixed, `tsc` raises `TS2578` (unused `@ts-expect-error`). That directive — commented `FIXME: missing "lib"` — suppressed a real error under TS 5.x, whose DOM lib doesn't type `Headers.entries()` (confirmed: removing it under `typescript@5.9.3` yields `TS2339`). TS 6.0's DOM lib types it, so the suppression is now unused. Rather than tie the test to one compiler version, this switches to `Headers.forEach`, which is typed in both DOM libs, so the helper type-checks under TS 5.x **and** 6.x with no suppression. (`CONTRIBUTING.md` rightly values intentional `@ts-expect-error` error-tests — this one wasn't testing an error, it was a lib-gap workaround.)

3. **`packages/openapi-fetch/test/no-strict-null-checks/tsconfig.json` — pin `rootDir`.** With (1)–(2) fixed, `tsc` raises `TS6059` (file outside `rootDir`). Per the [TS 6.0 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#rootdir-now-defaults-to-): *"Previously, if you did not specify a `rootDir`, it was inferred based on the common directory of all non-declaration input files. In TypeScript 6.0, the default `rootDir` will always be the directory containing the `tsconfig.json` file."* This config imports `../helpers.ts`, which fell outside the new default. Setting `"rootDir": "../.."` restores the 5.9-inferred root — the same adjustment the release notes' [`ts5to6`](https://github.com/andrewbranch/ts5to6) codemod performs.

A `patch` changeset is included so the widened peer range actually publishes. I chose `patch` to match how dependency/compatibility changes have been versioned here (e.g. `@redocly/openapi-core` bumps) — happy to make it `minor` if you'd prefer.

I'm opening this as a separate PR because #2711 is Renovate-managed and I can't push to its branch; glad to have these commits applied onto #2711 instead if that's easier for you.

Closes #2723
Closes #2550

## How to Review

On the pinned toolchain (`node` 24, `pnpm@10.30.3`):

- `pnpm run lint` → all packages pass (the failing task on #2711 was `openapi-fetch#lint`).
- `pnpm --filter openapi-fetch test` → 451 tests pass, no type errors.

To confirm the changes are TS6-specific and behavior-neutral:

- **TS6-specific:** on identical code, `tsc@5.9.3` is clean while `tsc@6.0.3` reports all three errors.
- **`downlevelIteration` is emit-neutral:** `tsc@6.0.3` produces byte-identical output at `target: ESNext` with the flag absent / `false` / `true`.
- **`Headers.forEach` is behavior-equivalent** (same key/value pairs) and type-checks under both `tsc@5.9.3` and `tsc@6.0.3`.

## Checklist

- [x] Unit tests updated — _no new tests needed; the change is type-level and the existing 451 tests pass unchanged_
- [x] `docs/` updated (if necessary) — _not necessary; no user-facing API/behavior change_
- [x] `pnpm run update:examples` run — _N/A; no `openapi-typescript` generator changes_

### DIFF
--- a/.changeset/support-typescript-6.md
+++ b/.changeset/support-typescript-6.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Allow TypeScript 6 as a peer dependency (`typescript: "^5.x || ^6.0.0"`).
--- a/packages/openapi-fetch/test/helpers.ts
+++ b/packages/openapi-fetch/test/helpers.ts
@@ -27,15 +27,15 @@ export function createObservedClient<T extends {}, M extends MediaType = MediaTy
  * Convert a Headers object to a plain object for easier comparison
  */
 export function headersToObj(headers: Headers | Record<string, string>): Record<string, string> {
-  const iter =
-    headers instanceof Headers
-      ? headers
-          // @ts-expect-error FIXME: this is a missing "lib" in tsconfig.json but dunno what
-          .entries()
-      : Object.entries(headers);
   const result: Record<string, string> = {};
-  for (const [k, v] of iter) {
-    result[k] = v;
+  if (headers instanceof Headers) {
+    headers.forEach((value, key) => {
+      result[key] = value;
+    });
+  } else {
+    for (const [key, value] of Object.entries(headers)) {
+      result[key] = value;
+    }
   }
   return result;
 }
--- a/packages/openapi-fetch/test/no-strict-null-checks/tsconfig.json
+++ b/packages/openapi-fetch/test/no-strict-null-checks/tsconfig.json
@@ -1,7 +1,8 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "strictNullChecks": false
+    "strictNullChecks": false,
+    "rootDir": "../.."
   },
   "include": ["."],
   "exclude": []
--- a/packages/openapi-fetch/tsconfig.json
+++ b/packages/openapi-fetch/tsconfig.json
@@ -2,7 +2,6 @@
   "compilerOptions": {
     "allowSyntheticDefaultImports": true,
     "declaration": true,
-    "downlevelIteration": false,
     "esModuleInterop": true,
     "lib": ["ESNext", "DOM"],
     "module": "NodeNext",
--- a/packages/openapi-typescript-helpers/package.json
+++ b/packages/openapi-typescript-helpers/package.json
@@ -34,6 +34,6 @@
     "lint:ts": "tsc --noEmit"
   },
   "devDependencies": {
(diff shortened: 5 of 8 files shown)


LMBernardo:
Just saw #2774 - moving this back to draft for now. Will close if that one lands.

LMBernardo:
After reviewing 2774 - it looks to be a strict superset of my changes here (except for the `forEach` change, 2774 uses a cast instead). Closing in preference of the pre-existing PR.

PR #2822, opened 2026-05-30
fix(openapi-react-query): propagate undefined errors from empty-body non-OK responses

### LINKED ISSUE #2070: openapi-react-query does not propagate `undefined` errors from openapi-fetch

EduardF1:
Closes #2070.

## What changed and why
- update `queryFn`, `useInfiniteQuery`'s `queryFn`, and `mutationFn` to throw when `error !== undefined || !response.ok`
- destructure `response` anywhere it was missing so the non-OK HTTP status is available
- add regression tests covering empty-body non-OK responses for query, mutation, and infinite query paths

## Scenario fixed
When `openapi-fetch` returns `{ data: undefined, error: undefined, response }` for a non-OK response with an empty body (for example, a 4xx/5xx response with `Content-Length: 0`), the previous `if (error)` check treated the result as successful and swallowed the failure. This change propagates those undefined errors correctly by also checking `response.ok`.


EduardF1:
Added the missing changeset for \\openapi-react-query\\ and pushed it in bdf4abf0.

PR #2829, opened 2026-06-16
fix(openapi-react-query): propagate undefined errors from empty-body non-OK responses

### LINKED ISSUE #2070: openapi-react-query does not propagate `undefined` errors from openapi-fetch

EduardF1:
Closes #2070.

## What changed and why

- Update `queryFn`, `useInfiniteQuery`'s `queryFn`, and `mutationFn` to throw when `error !== undefined || !response.ok`
- Destructure `response` in the two places it was missing so the non-OK HTTP status is available
- Add regression tests covering empty-body non-OK responses for query, mutation, and infinite query paths

## Scenario fixed

When `openapi-fetch` returns `{ data: undefined, error: undefined, response }` for a non-OK response with an empty body (e.g. a 4xx/5xx response with `Content-Length: 0`), the previous `if (error)` check treated the result as successful and silently swallowed the failure. This PR propagates those errors correctly by also checking `response.ok`.

## Test plan

- [x] `pnpm test` passes in `packages/openapi-react-query` (40 tests, all green)
- [x] Three new tests added: one each for `queryOptions`, `useMutation`, and `useInfiniteQuery`, asserting that empty-body 500 responses are correctly rejected/thrown

## Notes

This is a re-submission of #2822, which was closed because the branch was accidentally created as an orphan (not based on `main`) causing 1.8M spurious additions. This PR is cleanly branched from `main` with only the intended 91-line diff (10 source lines + 81 test lines).

### DIFF
--- a/.changeset/propagate-empty-error-response.md
+++ b/.changeset/propagate-empty-error-response.md
@@ -1,0 +1,5 @@
+---
+"openapi-react-query": patch
+---
+
+fix(openapi-react-query): propagate undefined errors from empty-body non-OK responses
--- a/packages/openapi-react-query/biome.json
+++ b/packages/openapi-react-query/biome.json
@@ -3,7 +3,7 @@
   "$schema": "https://biomejs.dev/schemas/2.3.14/schema.json",
   "extends": "//",
   "files": {
-    "includes": ["**", "!dist/**", "!test/fixtures/**"]
+    "includes": ["**", "!dist", "!test/fixtures"]
   },
   "linter": {
     "rules": {
--- a/packages/openapi-react-query/src/index.ts
+++ b/packages/openapi-react-query/src/index.ts
@@ -201,7 +201,7 @@ export default function createClient<Paths extends {}, Media extends MediaType =
     const fn = client[mth] as ClientMethod<Paths, typeof method, Media>;
     const { data, error, response } = await fn(path, { signal, ...(init as any) }); // TODO: find a way to avoid as any
     if (error !== undefined || !response.ok) {
-      throw error;
+      throw error ?? new Error(`Request failed with status ${response.status}`);
     }
     if (response.status === 204 || response.headers.get("Content-Length") === "0") {
       return data ?? null;
@@ -249,7 +249,7 @@ export default function createClient<Paths extends {}, Media extends MediaType =
 
             const { data, error, response } = await fn(path, mergedInit as any);
             if (error !== undefined || !response.ok) {
-              throw error;
+              throw error ?? new Error(`Request failed with status ${response.status}`);
             }
             return data;
           },
@@ -267,7 +267,7 @@ export default function createClient<Paths extends {}, Media extends MediaType =
             const fn = client[mth] as ClientMethod<Paths, typeof method, Media>;
             const { data, error, response } = await fn(path, init as InitWithUnknowns<typeof init>);
             if (error !== undefined || !response.ok) {
-              throw error;
+              throw error ?? new Error(`Request failed with status ${response.status}`);
             }
 
             return data as Exclude<typeof data, undefined>;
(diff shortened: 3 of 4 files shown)


EduardF1:
### Status update + root-cause of the failing `lint` check

**The bug fix is verified.** I built the workspace deps (`openapi-typescript-helpers`, `openapi-fetch`) and ran the package suite locally:

```
✓ test/index.test.tsx (40 tests) 1393ms
Test Files  1 passed (1)
     Tests  40 passed (40)
```

All 40 tests pass, including the three new regression tests (empty-body non-OK responses for `queryOptions`, `useMutation`, and `useInfiniteQuery`). The `Uncaught { code: 500 }` lines in the output are just jsdom echoing the intentionally-mocked 500 responses — not test failures.

**The failing `lint` check is _not_ caused by this PR.** The job fails at `openapi-react-query#lint` → `biome check .`. The flagged items are all in **pre-existing** code that this PR does not touch:

- `test/index.test.tsx:870` / `:873` — `noUnusedFunctionParameters` (`context` in the `onError`/`onSettled` mutation test)
- `test/index.test.tsx:1129` / `:1256` / `:1292` — `noUnusedVariables` (`firstRequestHandler` / `secondRequestHandler` in the pagination tests)
- a Biome `format` diff in pre-existing lines

Root cause is a Biome version drift: `@biomejs/biome` is pinned to `2.4.14` while `biome.json` still declares `"$schema": ".../2.3.14/schema.json"`. Biome `2.4.14` newly enforces these `correctness/*` rules. I reproduced the **identical** violations against `main` itself (`origin/main:packages/openapi-react-query/test/index.test.tsx`, lines 850/853/1085/1212/1248), so every open PR is currently blocked by this — `main`'s last green CI predates the bump.

For completeness, this PR's own added test blocks are Biome-clean (no new `lint`/`format` findings).

**Suggested resolution (maintainers):** land a small maintenance change on `main` — `biome migrate` to sync the schema, then `biome check --write --unsafe` for the pre-existing `noUnusedVariables`/`noUnusedFunctionParameters` + format — or pin Biome back to `2.3.x`. Once `main` is green, this PR's `lint` will pass (it's already `MERGEABLE`).

I'm happy to (a) fold the safe Biome auto-fixes into this PR to unblock CI immediately, or (b) rebase onto current `main`, whichever you prefer — just let me know.

EduardF1:
Pushed a small lint follow-up in 19a5ab9: applied the Biome 2.4-compatible ignore syntax for \\packages/openapi-react-query/biome.json\\ and the safe unused-parameter/format fixes in \\packages/openapi-react-query/test/index.test.tsx\\. Validation passes locally: \\pnpm --filter openapi-react-query lint\\ and \\pnpm --filter openapi-react-query test\\.

EduardF1:
@drwpow the `test-e2e` failure on this PR isn't caused by the change — it looks repo-wide.

The job hangs in `pnpm exec playwright install --with-deps` (`.github/workflows/ci.yml` line 50) and is cancelled at the 6-hour GitHub Actions job ceiling. Same step, same outcome on three different PRs and three different dates:

- #2829 — [run 28445979923](https://github.com/openapi-ts/openapi-typescript/actions/runs/28445979923/job/84295358804), 6h0m17s
- #2834 — [run 29234721736](https://github.com/openapi-ts/openapi-typescript/actions/runs/29234721736/job/86766648229), 6h0m19s
- #2835 — [run 28244142824](https://github.com/openapi-ts/openapi-typescript/actions/runs/28244142824/job/83678495759), 6h0m16s

In each one, `Run pnpm exec playwright install --with-deps` starts and never completes, and `pnpm run test-e2e` is then skipped, so the e2e tests never actually execute. `--with-deps` shells out to `apt-get` on the runner, which is the usual place this stalls waiting on input.

Flagging it here rather than on all three PRs. A `timeout-minutes` on that job would at least surface it in minutes instead of six hours.

PR #2842, opened 2026-07-14
Preserve tuples in readable and writable types

jessedijkstra:
Readable and writable OpenAPI client types now retain tuple length and element positions, so fixed-length response contracts remain compatible with generated schemas.

Previously, tuple responses were widened to generic arrays while filtering read-only and write-only fields.


lmcsu:
I tried the code in this PR and it makes this
image

Issue #1520, opened 2024-01-28
Empty object inside of allOf creates Record<string, never> in a type intersection


LostInStatic:
**Description**

I've decided to make it an issue mostly in case someone else is about to spend multiple hours figuring out what's wrong. That's why I'm going to be a little verbose - existing issues (#1474) are mostly tangentially related to the root cause, and I could have really used a description like this five hours ago. 

If there's an object with no properties in the schema, it will get converted to `Record<string, never>` - [this is the most sensible way to represent an empty object in TS](https://www.totaltypescript.com/the-empty-object-type-in-typescript). However, when such object is used in `allOf`, it is put along other object definitions in a TS type intersection. The way it is interpreted, it is combining other object types with an object that has all possible keys defined as invalid. 

Why would you put an object with no properties inside of an `allOf`? Beats me, but I've seen that pattern before - my best guess is that it is a common way of thinking about inheritance (baseRequest -> specificRequest even if the requests have nothing in common) or some common tooling works this way.

I ended up just running `openapi-typescript` with the `--empty-objects-unknown` CLI flag - this pretty much solved it for my use case, but you loose strict typing around empty objects. You could also add any property to the empty object if you control the schema.

If i had to fix it, I'd probably start with trying to omit any`Record<string, never>` in TS types intersection generation. Unfortunately I don't have space for taking a try at this and making a PR right now.

Hopefully this issue will help in documenting the problem. Cheers!

**Reproduction**

Run with the following input: 
```yaml
openapi: 3.0.0
components: 
  schemas:
    Empty:
      type: object
    Test: 
      allOf: 
        - $ref: "#/components/schemas/Empty"
        - type: object
          properties:
            foo: 
              type: string
```

**Expected result**

```typescript
[...]
export interface components {
    schemas: {
        Empty: Record<string, never>;
        Test: components["schemas"]["Empty"] & {
            foo?: string;
        }; // "{ }" is the only valid object for this type, even though the schema suggests otherwise.
    };
[...]
```

**Checklist**

- [x] My OpenAPI schema passes the [Redocly validator](https://redocly.com/docs/cli/commands/lint/) (`npx @redocly/cli@latest lint`)
- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/drwpow/openapi-typescript/blob/main/packages/openapi-typescript/CONTRIBUTING.md))

drwpow:
[...]

PR #2850, opened 2026-07-31
Handle required-only allOf constraints

### LINKED ISSUE #1474: allOff with required returning Record<string, never>
Similar question: https://github.com/OAI/OpenAPI-Specification/issues/1870

darkbasic:
## Summary

Handle `allOf` members that only make existing object properties required.

Fixes #1474  
Related to #1520 and #2570

## Problem

Given a schema such as:

```yaml
allOf:
  - $ref: "#/components/schemas/Base"
  - type: object
    required: [name]
```

`openapi-typescript` generated `Base & Record<string, never>`. The second member is a presence constraint, but it was transformed as an empty object, making the resulting type effectively unusable.

## Solution

Exact `{ required: [...] }` and `{ type: object, required: [...] }` members are translated through the existing parent `required` flow. For referenced schemas this produces `WithRequired<Base, "name">` instead of an empty-object intersection.

The implementation is intentionally conservative:

- Every required key must exist in the retained object schemas.
- Typed constraints require another retained, non-nullable object assertion.
- Schemas with extra keywords, callbacks, renamed or filtered properties, unions, nullability, or early-return types keep the previous behavior.
- Recursive `allOf` references are followed with cycle protection.
- Existing discriminator handling is preserved while allowing additional non-discriminator fields to become required.

The main logic is in `transform/schema-object.ts`. The composition tests include generated TypeScript compilation checks, and the DigitalOcean fixture was regenerated because it contains several real examples of this schema pattern.

## Validation

- 324 package tests passed
- TypeScript and generated-example checks passed
- Build and package export validation passed
- DigitalOcean generated fixture compiles

## Notes

While the original implementation has been written by me, after several rounds of validation by GPT-5.6 Sol it found so many corner cases that it basically rewrote it from scratch.

### DIFF
--- a/.changeset/clean-allof-required.md
+++ b/.changeset/clean-allof-required.md
@@ -1,0 +1,5 @@
+---
+"openapi-typescript": patch
+---
+
+Generate required composed properties from required-only `allOf` members instead of intersecting them with an empty object type.
--- a/.changeset/clean-allof-required.md
+++ b/.changeset/clean-allof-required.md
@@ -2,4 +2,4 @@
 "openapi-typescript": patch
 ---
 
-Generate required composed properties from required-only `allOf` members instead of intersecting them with an empty object type.
+Generate required composed properties from required-only `allOf` members, including typed object constraints when custom transforms are configured.
--- a/packages/openapi-typescript/examples/digital-ocean-api.ts
+++ b/packages/openapi-typescript/examples/digital-ocean-api.ts
@@ -7065,7 +7065,7 @@ export interface components {
             total?: number;
         };
         meta: {
-            meta: components["schemas"]["meta_properties"] & unknown;
+            meta: WithRequired<components["schemas"]["meta_properties"], "total">;
         };
         region: {
             /**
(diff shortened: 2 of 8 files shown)


darkbasic:
## Follow-up

After opening the PR, I tested the package with a generator that always supplies a custom `transform` callback. Even though the callback only converts binary schemas to `Blob`, its presence disabled the original required-only `allOf` translation for the entire document.

Commit `f82b194c` adds transform-aware handling. It preserves callback order and replacement behavior, then applies untouched typed object constraints to the completed `allOf` intersection through `WithRequiredObject<T, K>`. Referenced components still use their final transformed types; if a callback removes a required property, that property remains required as `unknown` instead of producing invalid TypeScript.

This keeps transformed output concise:

```ts
WithRequiredObject<CompanyResource, "organization" | "addresses" | "processingTypes">
```

The implementation deliberately keeps `WithRequired` unchanged because parent-level `required` and an explicit `type: object` constraint have different semantics. Callback replacements of the required-only member remain authoritative, and helper-name conflicts use an equivalent anonymous fallback.

The first transform-aware version was much larger than necessary. Commit `89e7c703` keeps the same behavior while:

- sharing one canonical AST between the named helper and collision fallback;
- consolidating repeated callback and compiler-test infrastructure;
- replacing one-off tests with focused behavior matrices;
- retaining coverage for recursion, unsafe schemas, callback lifecycle, replacements, discriminators, collisions, and generated TypeScript semantics.

That reduced the PR from 2,890 additions to 1,436 additions without dropping supported behavior.

The final suite passes with 306 tests. TypeScript checks, generated examples, build and package exports, the DigitalOcean fixture, live schema generation, and a freshly packed consumer all pass. The live callback-enabled schema produces three concise helper references, no `Record<string, never>` intersections, and no TypeScript diagnostics.

## Notes

The follow-up commit has been done entirely by GPT-5.6 Sol.

PR #2853, opened 2026-08-24
feat: add support for the HTTP QUERY method (RFC 10008)

SteveGT96:
## Changes

Adds support for the HTTP `QUERY` method ([RFC 10008](https://www.rfc-editor.org/rfc/rfc10008)) to `openapi-typescript` and `openapi-fetch`.

`QUERY` is a safe, idempotent method that carries a request body, filling the gap between `GET` (no body) and `POST` (neither safe nor idempotent). [OpenAPI 3.2](https://spec.openapis.org/oas/v3.2.0.html#path-item-object) recognises `query` as a Path Item verb, and it survives YAML→JSON conversion fine, but the generator treated it as an unknown property and silently dropped it — so no types were emitted and `openapi-fetch` had no way to call such an endpoint.

- **`openapi-typescript-helpers`** — adds `"query"` to `HttpMethod`.
- **`openapi-typescript`** — emits a `query` operation for the Path Items that declare one, and includes `query` in path-parameter extraction and the paths enum.
- **`openapi-fetch`** — adds `client.QUERY()` and `QUERY` on the path-based client.

This is a follow-up to #2844 (cc @jonty-comp, credited as co-author — this PR reuses their test schema and test cases), which has been open for a while. It takes the same approach with one deliberate difference, described below.

### Generated output is unchanged for documents that don't use `query`

The concern @jonty-comp raised for review in #2844 was:

> My principal concern in adding this feature is that all output types will be modified, even if they do not make use of `query` anywhere — which could confuse downstream consumers when all their fixtures suddenly change.

That is a real cost: adding `query` to the unconditional method list appends `query?: never;` to every Path Item of every document, which in #2844 meant ~4,400 changed lines across `examples/` and fixtures, and would mean a churned diff in every downstream repo that checks in generated types.

Since `query` was only introduced in OpenAPI 3.2, this PR emits it for the Path Items that declare it rather than for every Path Item:

```ts
const methods: readonly Method[] = "query" in pathItem ? METHODS : ALWAYS_EMITTED_METHODS;
```

This is behaviourally equivalent for consumers — `PathsWithMethod` treats an absent key and an optional `never` key identically — and `query?: never` is still emitted in the one case where it carries meaning (a `query` operation excluded via `excludeDeprecated`).

The result: `pnpm run update:examples` and regenerating every `openapi-fetch` test schema produce **zero** diff. The only generated file added is the new `query.d.ts` fixture.

I'm happy to switch to unconditional emission if you'd rather keep all nine verbs symmetrical — it's a one-line change.

### Idempotency

`QUERY` being safe and idempotent means sending a request twice must be equivalent to sending it once, so the client must not introduce per-request state of its own or consume the caller's input. `coreFetch` already satisfies this; the PR adds tests that pin the behaviour:

- repeated identical calls produce byte-identical requests (method, URL, headers, body);
- the caller's `init` object can be reused across calls and is neither consumed nor mutated;
- each call builds its own `Request`, so the body is not read once and replayed.

The request body is serialised with a `Content-Type` as RFC 10008 §2 requires.

No retry or response-caching layer is added — that's out of scope for this PR, and `openapi-fetch` doesn't have one today.

### Out of scope

OpenAPI 3.2's `additionalOperations` (arbitrary custom methods) is not addressed here. `query` is a first-class method under the RFC and is worth landing on its own, as @jonty-comp noted.

## How to Review

- `packages/openapi-typescript/src/transform/path-item-object.ts` holds the emission decision — that's the main thing worth a second opinion.
- To confirm the no-churn claim: `pnpm run update:examples` in `packages/openapi-typescript` and `pnpm run generate-types` in `packages/openapi-fetch` should both leave the tree clean.
- Runtime support: `QUERY` is not a forbidden method under the fetch spec, so `new Request(url, { method: "QUERY", body })` works on Node 18+ and in browsers. Servers and intermediaries that don't forward the verb will of course still reject it; the docs note this.

`pnpm run lint`, `pnpm run test` (780 tests), and `pnpm run size` all pass.

## Checklist

- [x] Unit tests updated


SteveGT96:
Rebased onto #2854.

`test-e2e` currently hangs on every PR in this repo until the 6h runner limit kills it, so this PR could not show a complete CI run. #2854 diagnoses and fixes that (a yauzl regression on Node 26 in Playwright < 1.60.0). I have rebased this branch on top of it so the e2e job can actually finish here.

That means this PR temporarily contains #2854's commit as well — the Playwright bump, the `timeout-minutes` guard and the lockfile change. Once #2854 is merged I will rebase onto `main` and that commit will drop out, leaving only the QUERY changes. If you would rather review this against `main` as-is, say so and I will drop the rebase.

PR #2867, opened 2026-09-07
Support TypeScript 7 applications with a package-owned compiler

### LINKED ISSUE #2841: openapi-typescript doesn't work with Typescript 7
### openapi-typescript version

7.13.0

### Node.js version

24.18.0

### OS + version

macOS 26.5.1

### Description

openapi-typescript doesn't work when I try to use typescript 7.0.2

```
const BOOLEAN = ts.factory.createKeywordTypeNode(ts.SyntaxKind.BooleanKeyword);
                           ^

TypeError: Cannot read properties of undefined (reading 'createKeywordTypeNode')
```

### Reproduction

N/A

### Expected result

It can generate types from openapi json schema

### Required

- [x] My OpenAPI schema is valid and passes the [Redocly validator](https://redocly.com/docs/cli/commands/lint/) (`npx @redocly/cli@latest lint`)

### Extra

shreyam1008:
## Changes

Fixes #2841. When an application installs TypeScript 7, the generator can resolve a compiler without the JavaScript `ts.factory` API and fail before generating anything. This PR gives `openapi-typescript` its own supported compiler dependency and exports it as `ts`. The application's compiler and `tsc` stay unchanged.

- Move the existing JavaScript compiler from a peer/dev dependency to a normal dependency.
- Export `ts`, update the Node API examples, and include a major changeset.
- Correct the CommonJS declarations to match the existing runtime's `.default` function and named exports. The installed-consumer test exposed unbuild's incorrect `export = openapiTS` rewrite.
- Test installed tarballs with pnpm and Bun's hoisted/isolated layouts in the existing Node 22/24 CI matrix. Cover fresh TS7 installs, TS7→TS5→TS7 transitions, frozen reinstalls, and preservation of the application's `tsc`.

**Major migration:** callers creating, inspecting, or printing AST nodes should change:

```diff
- import openapiTS from "openapi-typescript";
- import ts from "typescript";
+ import openapiTS, { ts } from "openapi-typescript";
```

This includes `transform`, `postTransform`, `transformProperty`, and AST type annotations. Different compiler versions can assign different numeric `SyntaxKind` values. CLI usage and generated output are unchanged. This supports TS7 applications through compiler isolation; it does not port generation to a native TS7 AST API.

The proposed API/release scope was posted [on the issue](https://github.com/openapi-ts/openapi-typescript/issues/2841#issuecomment-5574023607) before implementation. #2862 is the separate documentation workaround for the published release.

### Install-size tradeoff

TypeScript 5.9.3 is **23.6 MB unpacked**. A compatible TS5 application can share that compiler; TS7 applications need both versions. The Node guide now states this explicitly.

Clean npm 11.11.1 installations on Windows/Node 24.19.0, comparing the exact PR base `0cc7ee77` with the packed PR:

| Installation | Full node_modules size |
| --- | ---: |
| Base + TypeScript 5.9.3 | 39.39 MB |
| PR + TypeScript 5.9.3 | 39.39 MB |
| PR + TypeScript 7.0.2 | 70.26 MB |

These are logical file sizes in decimal MB, excluding external caches, not download sizes or browser bundles. The existing size-limit job measures `openapi-fetch`; it does not verify this dependency cost.

## How to Review

Using pnpm 10.30.3:

```sh
pnpm install --frozen-lockfile
pnpm --filter openapi-typescript lint
pnpm test
pnpm --filter openapi-typescript test:typescript-7
# With Bun 1.4.0 installed:
pnpm --filter openapi-typescript test:typescript-7 --bun
pnpm --filter openapi-typescript test:typescript-7 --bun --linker isolated
```

Verified locally:

- The three packed-consumer variants pass on **Node 22.12.0 and 24.19.0**, with TypeScript 5.9.3 and 7.0.2. CLI, ESM/CommonJS runtime, all three AST callbacks, compiler identity, and application `tsc` checks pass. Fixtures compile with `--strict --skipLibCheck false` and reject invalid generated-type data. Bun installations run under both Node and Bun.
- Full monorepo suite: **769 passed, two existing Windows-specific tests skipped**. Generator build/lint, example typechecks, and `attw` declaration/export checks pass; existing non-failing warnings remain.
- Separate npm strict-peer consumers pass with TS5 and TS7. Standalone generation works without a direct application compiler dependency. The small CLI fixture's output matches the published TS5 baseline byte for byte.
- Separate Bun upgrade checks reproduce the published package's TS7 crash, then pass after upgrading only the generator with the existing lockfile and `node_modules` retained, under both linkers.
- Browser E2E: **3 passed**, covering Chromium, Firefox, and WebKit. VitePress documentation build and `git diff --check` pass.

The strict fixture supplies `@types/js-yaml` and `json-schema-to-ts` for existing undeclared references in Redocly's public declarations. No declaration stubs, `any` casts, or `skipLibCheck` workaround are used.

## Checklist

- [x] Regression tests updated
- [x] Documentation updated
- [x] Major changeset includes migration impact
- [ ] `update:examples` — not run: no generation change; existing snapshots and example typechecks pass

Implementation and verification were performed with AI assistance. Maintainer agreement on compiler ownership, the size tradeoff, and the major release is still needed. The previous PR commit's upstream E2E job was cancelled during browser installation before tests ran; the results above are local verification, not a claim that the updated upstream CI has passed.


mrlubos:
@shreyam1008 if we're aiming for bundle size reduction and already accepted we'll need a major release, I'd be looking to remove the entire TypeScript dependency and switch to raw string templates. Can you make that change?

PR #2818, opened 2026-05-22
feat: Typescript 6 support

duncanbeevers:
This is based off of https://github.com/openapi-ts/openapi-typescript/pull/2774

I opened a PR against the original contributor's work, but it's taking a while to get the changes landed that way.
Please consider this PR as a replacement for #2774

---
# Summary

Adds TypeScript 6 support across the workspace, with a small fix to `Readable<T>` / `Writable<T>` so callable types survive the recursive mapped type under TS 6's stricter checks. Also tightens up the CI matrix and Playwright config so the E2E job actually exercises both major versions of TS.

### `openapi-typescript`

  - Widen typescript peer dependency: ^5.x → ^5.x || ^6.x.

### `openapi-typescript-helpers`

  - Fix `Readable<T>` / `Writable<T>` to pass callable types through unchanged (Date, RegExp, functions, class instance methods). Previously the recursive mapped type recursed into method signatures and collapsed them
  to `{}` under `--strict`, which broke patterns like `Readable<{ createdAt: Date }>.createdAt.toISOString()`. Reproduces on both TS 5 and TS 6.
  - Add type-level tests (`readable-writable.test-d.ts`) covering Date / RegExp / function passthrough plus negative controls that confirm $Read / $Write stripping still works on plain nested objects.
  - Move TypeScript `devDependency` onto the workspace catalog.

###  Workspace

  - Bump catalog typescript: ^5.9.3 → ^6.0.2.

### CI (`.github/workflows/ci.yml`)

  - Expand the test matrix to typescript-version: ["5", "6"]. The TS 5 leg rewrites the workspace catalog via sed and installs with `--no-frozen-lockfile` so we keep a backward-compat signal without committing two
  lockfiles.
  - Pin `test-e2e` to Node 22 (latest LTS) to avoid stream leaks seen on newer Node.
  - Add a 10-minute timeout, DEBUG=pw:webserver,pw:browser, and a playwright-report artifact upload (1-day retention) to the E2E job so hangs fail fast and are debuggable.

### `openapi-fetch`

  - Playwright config: fix the PORT parse expression (process.env.PORT || 4173 || "" → process.env.PORT || "4173"), bind to 127.0.0.1, use `pnpm --filter openapi-fetch` for the webserver command, enable
  reuseExistingServer locally, raise webserver timeout to 120s to cover TS 6 type-checking on cold starts, retain Playwright traces on failure, and pass `--no-sandbox` to Chrome.
[...]

PR #2868, opened 2026-09-11
feat!: generate TypeScript from string templates and drop the TypeScript dependency

### LINKED ISSUE #2841: openapi-typescript doesn't work with Typescript 7

im10furry:
## Changes

**Closes #2841. Supersedes #2867. Related: #2862, #2818.**

TypeScript 7 (the native compiler) is `latest` on npm and its root module ships **no classic
compiler API**. `openapi-typescript` built its output with `ts.factory` / `createPrinter` /
`createSourceFile`, so a TS7 application hits two hard blockers:

1. `npm i typescript@7 openapi-typescript` fails outright:
   ```
   npm error ERESOLVE unable to resolve dependency tree
   npm error peer typescript@"^5.x" from openapi-typescript@7.13.0
   ```
2. Forcing past it (`--legacy-peer-deps`) crashes on first import:
   ```
   TypeError: Cannot read properties of undefined (reading 'createKeywordTypeNode')
   ```

This PR does what @mrlubos asked for in #2841 — *"remove the entire TypeScript dependency and switch
to raw string templates"* — so **TypeScript is no longer a dependency at all**.

- **`src/lib/ts.ts` rewritten** as a string builder. The helper surface keeps its shape (`tsUnion`,
  `tsLiteral`, `oapiRef`, `tsEnum`, …) but now returns source text, plus new line-oriented builders
  (`typeLiteral`, `tupleType`, `propertySignature`, `indexSignature`, `typeAlias`, `interfaceDecl`,
  `enumDecl`) and `tsComment()` for JSDoc.
- **All 13 transforms, `lib/utils.ts` and `transform/index.ts` migrated.** Multi-line constructs thread
  an explicit indentation level, reproducing the printer's 4-space nesting exactly.
- **The printer's quirks are reproduced deliberately**, each verified against the real printer:
  trailing whitespace stripped on every comment line; nested unions/intersections parenthesised
  (including inside array types and after `readonly`); tuples always multi-line while `{}` is inline;
  string literals escaped the way `createStringLiteral` does (`"emoji🎉"` → `"emoji\uD83C\uDF89"`,
  while the `tsLiteral` UTF-8 workaround stays verbatim); and the `$defs` / template-literal /
  `operations` assembly keeps its exact shape.
- **`openapiTS()` resolves to a `string`** — the complete file body, ending in a newline.
  `astToString()` is retained as a normalizer for older call sites.
- **`peerDependencies.typescript` removed**; `typescript` stays a devDependency for the repo's own
  `tsc`. The built `dist/` contains **zero references to `typescript`**.

### Breaking changes (major)

| Before | After |
| --- | --- |
| `openapiTS(): Promise<ts.Node[]>` | `openapiTS(): Promise<string>` |
| `transform` → `ts.TypeNode \| TransformObject` | `transform` → `string \| { schema: string; questionToken: boolean }` |
| `postTransform(type: ts.TypeNode)` | `postTransform(type: string)` |
| `transformProperty(property: ts.PropertySignature)` | `transformProperty(property: { name; optional; type; comment?; indent })` |
| `GlobalContext.injectFooter: ts.Node[]` | `GlobalContext.injectFooter: FooterDeclaration[]` |
| `stringToAST()`, `tsModifiers()`, `QUESTION_TOKEN` | removed; see the new builders above |

Migration for AST callbacks is mechanical: `ts.factory.createTypeReferenceNode("Date")` → `"Date"`.
Attach JSDoc from `transformProperty` with `tsComment(tags, property.indent)`.

The `inject` option is now emitted verbatim rather than re-printed by the printer.

### Two bugs fixed along the way

Both produced **invalid TypeScript**; both are covered by new tests.

- A multi-line `x-enum-descriptions` entry leaked a bare token into the enum body
  (`// line1\nline2` → now `// line1 line2`).
- With `pathParamsAsTypes`, a URL containing a backtick broke out of the generated template literal
  type (`` `/a`b/${string}` `` → now `` `/a\`b/${string}` ``).

One deliberate, semantically-neutral output change: `tsUnion()`/`tsIntersection()` no longer emit a
redundant single-member union, so `(string)[][]` becomes `string[][]`.

## How to Review

- **The core is `packages/openapi-typescript/src/lib/ts.ts`.** Everything else is a mechanical port
  against it; `transform/schema-object.ts` is the largest consumer.
- **Ignore the diff noise in the transform files** — most of it is `ts.factory.createX(...)` becoming
  a string builder call plus an `indent` argument. The interesting parts are `$defs`, the array /
  `arrayLength` branches, and the `operations` assembly.
- `transform/index.ts` keeps a `typeLiteralMembers()` helper that mirrors the old `.members?.length`
  check on the generated AST node, which is how the root `paths`/`webhooks`/`components`/`$defs`
  interface-vs-fallback decision still works.
- The rewrite is one atomic commit on purpose: swapping the node type to `string` ripples through
  every transform, so the repo cannot be green in between. The byte-exact snapshot suite is what
  makes that safe.

### Verification

**Output parity.** Every committed example snapshot diffs clean, including the 123,077-line GitHub
API outputs:

```
github-api                          diff=0      octokit-ghes-3.6-diff-to-api   diff=0
github-api-immutable                diff=0      stripe-api                     diff=0
github-api-export-type-immutable    diff=0      digital-ocean-api              diff=0
github-api-root-types               diff=0      github-api-next                diff=0
                                                enum-root-types                diff=0
```

Beyond the committed fixtures, I ran the published 7.13.0 and this branch side by side over a
purpose-built schema (discriminators, allOf/oneOf/anyOf, patternProperties, `$defs`, prefixItems,
readOnly/writeOnly, `x-enum-*` metadata, non-ASCII in descriptions, enum values, property names and
`$ref` targets) across **27 option combinations** — including a kitchen-sink run — and every pair is
byte-identical. `pnpm run update:examples` produces no diff.

**Suite.** `pnpm test` → 26 files / 297 tests pass (byte-exact `toMatchFileSnapshot` fixtures, inline
`want` assertions, CLI snapshot tests, CJS bundle, Node API). `tsc --noEmit`,
`tsc -p tsconfig.examples.json --noEmit` and `biome check` are clean (the 3 remaining
`noTemplateCurlyInString` warnings are pre-existing on untouched lines).

**Compiler matrix** — same schema generated, then type-checked with each compiler:

| TypeScript | install | generation parity | `tsc --noEmit --strict` on output |
| --- | --- | --- | --- |
| 5.9.3 | clean | identical | PASS |
| 6.0.3 | clean | identical | PASS |
| 7.0.2 (native) | clean, **no peer conflict** | identical | PASS |

**End-to-end under TS7**, the exact scenario from #2841:

```
$ npm i typescript@7 openapi-typescript-<this PR>.tgz    # no flags, no ERESOLVE
$ npx openapi-typescript ./schema.yaml -o ./out.d.ts      # ✅
```

With **no TypeScript installed at all**, both the CLI and the Node API (including `transform`,
`postTransform` and `transformProperty`) work — `typescript` does not even resolve from the consumer
project.

## Checklist

- [x] Unit tests updated

### Notes

- The `ja`/`zh` translations of `docs/node.md` still show the old AST examples and need a follow-up
  translation pass.
- No changes to `openapi-typescript-helpers`, `openapi-fetch`, or any other package.
- With no compiler dependency at all, #2867's "compiler isolation" question becomes moot.


im10furry:
### Open question: the `transformProperty` hook shape

I posted a fuller sign-off request on #2841; this is the one API decision I'd like a maintainer's call on before this lands, because it is the only place where the string rewrite can't preserve the old shape exactly.

**What changed.** `transformProperty` used to receive a `ts.PropertySignature` node and return one. A plain string can't carry that structure — the hook needs to read and change the property name, whether it is optional, and its type — so it now receives and returns a small plain object:

```ts
transformProperty?: (
  property: { name: string; optional: boolean; type: string; comment?: string; indent: string },
  schemaObject: SchemaObject,
  options: TransformNodeOptions,
) => typeof property | undefined;
```

`comment` and `indent` exist because the old hook could attach JSDoc via `ts.addSyntheticLeadingComment`, and a correctly-indented block comment needs to know the property's indentation. The PR adds a public `tsComment(lines, indent)` helper so hooks don't have to hand-build one:

```ts
transformProperty(property, schemaObject) {
  if (schemaObject.minLength === undefined) return undefined;
  return { ...property, comment: tsComment([`@minLength ${schemaObject.minLength}`], property.indent) };
}
```

**The alternatives I considered:**

1. **This PR** — a structured `{ name, optional, type, comment?, indent }` object. Keeps the old ergonomics (patch one field, return the object) and keeps indentation out of the caller's hands.
2. **Raw property string** — the hook gets the already-rendered `name?: type;` line and returns a replacement line. Maximum flexibility and the most "string templates" spirit, but callers have to parse/re-emit the line and handle indentation themselves.
3. **Only change the return type** — keep passing an object but let the hook return a plain string for the rendered property. A middle ground, but the asymmetry reads oddly.

I went with (1) as the closest analogue to the old API, but (2) is defensible if you'd rather keep the public surface strictly string-based. Happy to rework it — it's contained to `transformProperty` in `src/transform/schema-object.ts`, the hook's type in `src/types.ts`, and two test cases.

Everything else in the breaking-change list should be a mechanical migration for callers (`ts.factory.createTypeReferenceNode("Date")` → `"Date"`), and CLI users plus consumers of the generated types need no migration at all.

im10furry:
### CI note: `test-e2e` is stuck on an infrastructure step

Every check passes except `test-e2e`, which has been sitting in `pnpm exec playwright install --with-deps` for ~20 minutes. That step never completed, so the suite itself (`pnpm run test-e2e`) is still `pending` and hasn't run — this is not a test failure.

That job isn't exercised by this change: it runs the `openapi-fetch` Playwright suite, and `packages/openapi-fetch/test/helpers.ts` only imports `openapi-typescript-helpers`. The same code passes `pnpm test` on the macOS and Windows jobs, which do cover the packages.

It also matches the flakiness this job already has on `main` — it runs on `node-version: latest`, which is exactly what #2818 pins to Node 22 to avoid. For context, `test-e2e` is `CANCELLED` on #2862, #2867 and #2815.

I don't have permission to cancel or re-run the workflow (`viewerPermission=none` on this repo), so a maintainer re-run would be needed to clear it. Everything else is green, and the failure mode here is a stuck dependency download rather than anything in the diff.

im10furry:
**Follow-up:** the stuck `test-e2e` job here is a repo-wide, pre-existing failure, and I've opened #2869 to fix it.

`test-e2e` has been dying in `pnpm exec playwright install --with-deps` at GitHub's 6-hour job limit on every run since 2026-08-24 (last green run 32722419757, 79s). #2869 moves that job into the official Playwright container so there is nothing to install, and it is green there in 57s.

Once #2869 lands, rebasing this PR on `main` will make `test-e2e` green here too — nothing in this PR's diff is involved.

PR #2870, opened 2026-09-11
fix(turbo): declare build outputs so cache hits restore dist/

im10furry:
## Changes

`turbo.json` never declared `outputs` for any task, so Turborepo had no idea which files a task produces. A cache hit therefore replays the task's logs and reports success **without writing `dist/`**:

```console
$ rm -rf packages/openapi-typescript-helpers/dist
$ npx turbo run build --filter=openapi-typescript-helpers
openapi-typescript-helpers:build: cache miss, executing a2386593374fa018

$ rm -rf packages/openapi-typescript-helpers/dist
$ npx turbo run build --filter=openapi-typescript-helpers
openapi-typescript-helpers:build: cache hit, replaying logs a2386593374fa018   # reports success
$ ls packages/openapi-typescript-helpers/dist
ls: No such file or directory                                                  # ...but produced nothing
```

That is a false green: `turbo run build` can succeed with no artifacts on disk, so anything that consumes a package fails or silently reads stale output. The clearest case is `openapi-react-query`, whose test script shells out to the `openapi-typescript` CLI:

```console
$ cd packages/openapi-react-query && pnpm run test
> openapi-typescript test/fixtures/api.yaml -o test/fixtures/api.d.ts
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
  '.../packages/openapi-typescript/dist/index.mjs' imported from '.../openapi-typescript/bin/cli.js'
```

### Reproduction

1. `npx turbo run build --filter=openapi-react-query` — populates the cache
2. delete every `packages/*/dist`
3. `npx turbo run build --filter=openapi-react-query` — cache hit, reports success
4. `cd packages/openapi-react-query && pnpm run test`

On `main`, step 4 exits 1 with the `ERR_MODULE_NOT_FOUND` above. With this PR it exits 0, with 37 tests passing. I checked both directions — removing the new line reproduces the failure again.

### Fix

Declare `"outputs": ["dist/**"]` on the `build` task so Turbo caches and restores the artifacts. Every package's `build` writes to `dist/`.

## How to Review

- One line plus a comment in `turbo.json`; nothing else changes.
- Declaring `outputs` changes a task's hash, so the first run after this lands rebuilds everything once. After that, cache hits restore `dist/` as they should.
- No effect on CI as it stands today, since the workflow doesn't persist `.turbo/cache`. It matters for local `turbo` usage, and it's a prerequisite for caching Turborepo output in CI.
- I found this while trying to add that CI caching in #2869. It looked like a large speedup (`>>> FULL TURBO`, 7/7 cached, 93ms), but that was only because nothing ran; the moment I forced the tests to execute, the missing artifacts surfaced on three platforms. I've kept #2869 to the e2e fix only and left the caching out.

For what it's worth, my first guess was wrong: I assumed the task graph was missing a cross-package `^build` edge. It isn't — the graph is correct (`openapi-react-query#test → openapi-react-query#build → openapi-typescript#build`), and the missing `outputs` is the whole cause.

## Checklist

- [ ] Unit tests updated — not applicable, build configuration only
- [ ] `docs/` updated (if necessary) — not applicable
- [ ] `pnpm run update:examples` run (only applicable for openapi-typescript) — not applicable

No changeset: `turbo.json` is build orchestration and is not published.


im10furry:
**Note on the pending `test-e2e`.** This branch is based on current `main`, so its e2e job hits the repo-wide hang in `playwright install --with-deps` (6-hour job limit on every run since 2026-08-24). That's unrelated to this change — two 1-line config edits can't affect it.

#2869 moves that job into the official Playwright container and is green in 57s. Like #2868, this PR doesn't touch `.github/workflows/ci.yml`, so once #2869 lands the fixed workflow applies here and `test-e2e` will go green without any change to this diff.

